PR #16, opened 2020-07-28
Add VCX Java demo

baegjae:
VCX Java Faber/Alice demo.
It references to the VCX node demo.

Signed-off-by: Ethan Sung <baegjae@gmail.com>


demo/java/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+repositories {
+    mavenCentral()
+    maven {
+        url "http://13.125.219.189/repository/maven-public/" // skt maven
+    }
+}
+
+dependencies {
+    // wrapper for vcx library
+    implementation 'com.evernym:vcx:0.8.2-sktston-SNAPSHOT' // use maven repository
```
Patrik-Stas, line 19: I'll merge it as is, but it would be cool if you'd add github action to publish java wrapper and then use that in this java demo

PR #18, opened 2020-07-29
Make a comment field as an optional

airoasis:
According to aries rfcs 
https://github.com/hyperledger/aries-rfcs/tree/master/features/0036-issue-credential
https://github.com/hyperledger/aries-rfcs/tree/master/features/0453-issue-credential-v2
the comment field is an optional field. 
However, libvcx expects this as a mandatary, so if  the comment is empty it causes Json parsing error 
```
Caused by: Cannot deserialize A2A message: missing field 'comment'
```
This PR fixes this issue.


airoasis:
FYI: I made PR in indy-sdk as well, and Daniel Hardman reviewed it already. 
https://github.com/hyperledger/indy-sdk/pull/2220

PR #26, opened 2020-08-11
Migrate docker base image to Alpine Linux

marekprague:


.github/workflows/main.yml
```diff
@@ -134,7 +134,7 @@ jobs:
 
   build-image-indypool:
     needs: workflow-setup
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
mirgee, line 137: Let's state version explicitly here as well.

ci/libvcx.dockerfile
```diff
@@ -1,81 +1,76 @@
-FROM base:1.0.0 AS BASE
+FROM alpine:3.12 AS builder
 
-# Install Rust toolchain
-ARG RUST_VER=1.40.0
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain ${RUST_VER}
-ENV PATH /home/indy/.cargo/bin:$PATH
+ARG UID=1000
+ARG GID=1000
```
mirgee, line 4: Let's make these ARGs global, since we reuse them (i.e. define them before the base image).

ci/libvcx.dockerfile
```diff
@@ -1,81 +1,76 @@
...
-      apt-transport-https \
-      ca-certificates \
-      curl \
-      build-essential
-
-RUN useradd -ms /bin/bash -u 1000 indy
+RUN apk update && apk upgrade && \
+    apk add --no-cache \
+        build-base \
+        cargo \
```
mirgee, line 17: Is it possible to install specific version of cargo / rust toolchain (passed as a build ARG)? If not, we should also remove the unnecessary `--build-arg` flags from CI.

ci/libvcx.dockerfile
```diff
@@ -1,81 +1,76 @@
...
+        git \
+        libsodium-dev \
+        libzmq \
+        openssl-dev \
+        rust \
+        sqlite-dev \
+        zeromq-dev
```
mirgee, line 24: What is the path for the libraries when installed this way?

ci/libvcx.dockerfile
```diff
@@ -1,81 +1,76 @@
...
+RUN apk update && apk upgrade && \
+    apk add --no-cache \
+        build-base \
+        cargo \
+        git \
+        libsodium-dev \
+        libzmq \
+        openssl-dev \
+        rust \
+        sqlite-dev \
```
mirgee, line 23: What do we need sqlite for?

ci/libvcx.dockerfile
```diff
@@ -1,81 +1,76 @@
...
+RUN apk update && apk upgrade
+RUN apk add --no-cache \
+        bash \
+        g++ \
+        gcc \
+        git \
+        libsodium-dev \
+        libzmq \
+        make \
+        nodejs@alpine38 \
```
mirgee, line 67: I think it would be useful to make the node version configurable via an ARG. If not possible, let's remove it from the `main.yml` as well.

ci/libvcx.dockerfile
```diff
@@ -1,81 +1,76 @@
...
+        npm@alpine38 \
+        openssl-dev \
+        python2 \
+        sqlite-dev \
+        zeromq-dev
+
+LABEL org.label-schema.schema-version="0.8.0"
```
mirgee, line 74: What schema should this version correspond to?

ci/libvcx.dockerfile
```diff
@@ -1,81 +1,76 @@
...
 LABEL org.label-schema.name="libvcx"
 LABEL org.label-schema.version="${INDYSDK_REVISION}"
```
mirgee, line 76: Isn't it confusing that the version label of libvcx image is a version of its dependency? We should change this to libvcx version, or remove it.

ci/libvcx.dockerfile
```diff
@@ -1,81 +1,76 @@
...
+RUN cargo build --release --manifest-path=$INDYSDK_PATH/libnullpay/Cargo.toml
+RUN cargo build --release --manifest-path=$INDYSDK_PATH/experimental/plugins/postgres_storage/Cargo.toml
 
-# TODO: Just copy the binary and add to path
-ARG RUST_VER=1.40.0
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain ${RUST_VER}
-ENV PATH /home/indy/.cargo/bin:$PATH
-
-LABEL org.label-schema.schema-version="1.0"
+USER root
```
mirgee, line 42: The last USER should probably be indy.

marekprague:
@mirgee requested changes: https://github.com/AbsaOSS/libvcx/pull/26/commits/10f5256a9be68e3e8a930440f9f1871b624c3b1c

mirgee:
The `libvcx.dockerfile` conflicts may be resolved in favor of your changes

PR #29, opened 2020-08-12
Simplify ubuntu libvcx image

Patrik-Stas:


ci/libvcx.dockerfile
```diff
@@ -1,81 +1,110 @@
...
+LABEL org.label-schema.name="indysdk"
+LABEL org.label-schema.version="${INDY_VERSION}"
 
 # TODO: Just copy the binary and add to path
-ARG RUST_VER=1.40.0
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain ${RUST_VER}
 ENV PATH /home/indy/.cargo/bin:$PATH
 
 LABEL org.label-schema.schema-version="1.0"
 LABEL org.label-schema.name="libvcx"
```
mirgee, line 109: The same label is defined multiple times

ci/libvcx.dockerfile
```diff
@@ -58,7 +107,7 @@ COPY --from=BASE --chown=indy /usr/lib/libindy.so /usr/lib/libindy.so
 COPY --from=BASE --chown=indy /usr/lib/libvcx.so /usr/lib/libvcx.so
 COPY --from=BASE --chown=indy /usr/lib/libnullpay.so /usr/lib/libnullpay.so
 COPY --from=BASE --chown=indy /usr/lib/libindystrgpostgres.so /usr/lib/libindystrgpostgres.so
-
+#
```
mirgee, line 110: Typo

ci/libvcx.dockerfile
```diff
@@ -67,7 +116,7 @@ ARG NODE_VER=8.x
 RUN curl -sL https://deb.nodesource.com/setup_${NODE_VER} | bash -
 RUN apt-get install -y nodejs
 
-RUN chown -R indy . 
+RUN chown -R indy .
 
 USER indy
```
mirgee, line 122: How about removing the LABELs?

ci/libvcx.dockerfile
```diff
@@ -58,7 +107,7 @@ COPY --from=BASE --chown=indy /usr/lib/libindy.so /usr/lib/libindy.so
 COPY --from=BASE --chown=indy /usr/lib/libvcx.so /usr/lib/libvcx.so
 COPY --from=BASE --chown=indy /usr/lib/libnullpay.so /usr/lib/libnullpay.so
 COPY --from=BASE --chown=indy /usr/lib/libindystrgpostgres.so /usr/lib/libindystrgpostgres.so
-
+#
 COPY --from=BASE --chown=indy /home/indy/libvcx ./libvcx
 COPY --from=BASE --chown=indy /home/indy/wrappers/node ./wrappers/node
```
mirgee, line 112: This is potentially unncessary and can further reduce the image size

ci/libvcx.dockerfile
```diff
@@ -1,4 +1,53 @@
-FROM base:1.0.0 AS BASE
+FROM ubuntu:16.04 as BASE
+
+ARG uid=1000
+
+# Install dependencies
+RUN apt-get update && \
```
mirgee, line 6: We can reduce the image size further by pruning those dependencies

ci/libvcx.dockerfile
Patrik-Stas: That wouldn't reduce size much. Avoiding installing rustup here would save 700mb, but lets leave that for later..(since Ubuntu libvcx might be thrown away soon anyway)

ci/libvcx.dockerfile
Patrik-Stas: But we need to test and publish node stuff. I was thinking of splitting the dockerfile into having
- libvcx without node, rust, without wrappers
- libvcx "testing image" built on top of the one above, used for testing in CI

But again, let's leave that for later. The small changes in this PR reduces image size since already by couple GB

PR #39, opened 2020-08-16
Test should produce logs only if LOG_LEVEL is specified

Patrik-Stas:


libvcx/src/settings.rs
```diff
@@ -73,7 +73,10 @@ pub static DEFAULT_WALLET_KEY: &str = "8dvfYSt5d1taSd6yJdpjq4emkwsPDDLYxkNFysFD2
 pub static DEFAULT_THREADPOOL_SIZE: usize = 8;
 pub static MASK_VALUE: &str = "********";
 pub static DEFAULT_WALLET_KEY_DERIVATION: &str = "RAW";
+#[cfg(not(target_os = "macos"))]
 pub static DEFAULT_PAYMENT_PLUGIN: &str = "libnullpay.so";
+#[cfg(target_os = "macos")]
+pub static DEFAULT_PAYMENT_PLUGIN: &str = "libnullpay.dylib";
```
mirgee, line 79: Good catch, is it related to the logging? Was it causing any issues?

libvcx/src/utils/logger.rs
```diff
@@ -23,6 +23,7 @@ use error::prelude::*;
 
 
 use utils::libindy;
+use std::env::VarError;
```
mirgee, line 26: Is this import used?

libvcx/src/utils/logger.rs
```diff
@@ -140,10 +141,14 @@ impl LibvcxDefaultLogger {
 
         // ensures that the test that is calling this wont fail simply because
         // the user did not set the RUST_LOG env var.
```
mirgee, line 143: I guess this comment is correct but confusing, I would remove it.

libvcx/src/settings.rs
Patrik-Stas: It's not related to logging actually - separate fix. Any tests trying to load libnullpay would fail on OSX because it was tried to load `.so`, while correctly it should be looking for `.dylib`

libvcx/src/utils/logger.rs
Patrik-Stas: good catch

libvcx/src/utils/logger.rs
```diff
@@ -138,12 +138,15 @@ impl LibvcxDefaultLogger {
...
-        let pattern = Some(env::var("RUST_LOG").unwrap_or("trace".to_string()));
-        match LibvcxDefaultLogger::init(pattern) {
-            Ok(()) => (),
-            Err(_) => (),
+        // if RUST_LOG is not set, logs will not be captured
+        match env::var("RUST_LOG") {
+            Ok(log_pattern) => {
+                match LibvcxDefaultLogger::init(Some(log_pattern)) {
+                    Ok(()) => (),
+                    Err(_) => (),
```
mirgee, line 146: I think we want to at least log the error message in case the user did set `RUST_LOG` but init failed, no?

mirgee (approved):

PR #53, opened 2020-08-21
Node wrapper aries testing

Patrik-Stas:
- Legacy tests were relying on passing `ENABLE_TEST_MODE` when calling `vcx_init`, which would pre-set libvcx into testing mode with mocked enabled.
- Instead of doing that, tests are passing their own desired configuration now
- The tests are now specifying they want to run with `3.0` protocol_type (let's move it to `4.0` in next step)
- Some tests are disabled. Especially those which are relying on some agency `1.0` mock responses. These responses are currently hardcoded in libvcx, so internals must be changed / or testing approach, before these tests are enabled again.
- Some tests were easy enough to migrate from legacy to Aries (for example test checking generating connection invitation)

- Additionally fixes issue with vcx node wrapper `suite3` logging tests (we were not running these tests and I think it's been probably failing since switch to different ffi packages). The fix is done in `wrappers/node/src/api/logging.ts`


wrappers/node/test/helpers/utils.ts
```diff
@@ -1,36 +1,71 @@
...
-  return stubInitVCX
-}
-export const patchInitVCX = () => {
-  if (!_patchInitVCXObj) {
-    _patchInitVCXObj = _patchInitVCX()
-  }
-  return _patchInitVCXObj
+// let _initVCXCalled = false
+// let _patchInitVCXObj: SinonStub | undefined
+// const _patchInitVCX = () => {
```
Patrik-Stas, line 9: Not quite sure what this was for. I'll keep it in for now.

wrappers/node/test/suite1/ariesvcx-credential.test.ts
```diff
@@ -47,10 +53,11 @@ describe('Credential:', () => {
     })
   })
 
-  describe('createWithMsgId:', () => {
-    it('success', async () => {
-      await credentialCreateWithMsgId()
-    })
+  describe('createWithMsgId', () => {
+    // TODO: to enable this test, credential offer must be mocked in aries code of get_credential_offer_msg
+    // it('createWithMsgIdsuccess', async () => {
```
mirgee, line 58: Why is this commented out, while other tests just skipped?

wrappers/node/test/suite1/ariesvcx-disclosed-proof.test.ts
```diff
@@ -47,9 +51,9 @@ describe('DisclosedProof', () => {
   })
 
   describe('createWithMsgId:', () => {
-    it('success', async () => {
-      await disclosedProofCreateWithMsgId()
-    })
+    // it('success', async () => {
```
mirgee, line 54: Please skip insted of commenting out or add a todo consistently, so it is easy to find which tests need to be restored

wrappers/node/test/suite1/ariesvcx-proof.test.ts
```diff
@@ -8,9 +8,10 @@ import {
 } from 'helpers/entities'
 import { initVcxTestMode, shouldThrow } from 'helpers/utils'
 import { Connection, DisclosedProof, Proof, ProofState, StateType, VCXCode, VCXMock, VCXMockMessage } from 'src'
+import {PROTOCOL_TYPE_ARIES} from '../helpers/test-constants'
```
mirgee, line 11: I am so sorry for nitpicking, but please keep the bracketing consistent :) (in all the imports)

wrappers/node/test/suite1/ariesvcx-proof.test.ts
```diff
@@ -69,9 +71,18 @@ describe('Proof:', () => {
...
+      // console.log(`Serialized proof1 = ${JSON.stringify(data1)}`)
       const proof2 = await Proof.deserialize(data1)
-      assert.equal(proof2.sourceId, proof1.sourceId)
+      // console.log(`Deserialized proof1data = ${JSON.stringify(proof2)}`)
+      // todo: Does not hold in aries, the TS/JS representation after serialize->deserialize in incorrect because
+      // IProofData structure is matching legacy structure
+      // perhaps we could make JS layer thinner and instead of trying to keeping attributes like _requestedAttributes
+      // in javascript representations, we could rather add method on libvcx proof vcx_proof_get_requested_attributes
+      // which could encompass the logic of how to retrieve this data from internal proof representation.
+      // The downside is some overhead associated with FFI.
```
mirgee, line 84: Honestly, I don't like the idea of adding `vcx_proof_get_requested_attributes`, requested attributes should be part of the serialized proof object in the first place. And since the wrapper is written in TS, this forces us to type the expected structure - and the benefits are arguable...

Anyway, no need to keep the this part of the comment here.

wrappers/node/test/suite1/ariesvcx-proof.test.ts
```diff
@@ -69,9 +71,18 @@ describe('Proof:', () => {
   describe('deserialize:', () => {
     it('success', async () => {
       const proof1 = await proofCreate()
+      console.log(`Original proof1 ${JSON.stringify(proof1)}`)
```
mirgee, line 74: Please, remove the logs

wrappers/node/test/suite1/ariesvcx-proof.test.ts
```diff
@@ -106,6 +117,7 @@ describe('Proof:', () => {
   })
 
   describe('requestProof:', () => {
+
```
mirgee, line 120: No need for a newline here?

wrappers/node/test/suite2/ffi.test.ts
```diff
@@ -32,7 +32,7 @@ describe('Using the vcx ffi directly', () => {
   const libDir = libPath[platform.toLowerCase()] || libPath['linux']
   const run = new VCXRuntime({ basepath: `${libDir}libvcx${postfix}` })
 
-  before(() => initVcxTestMode())
+  before(() => initVcxTestMode('3.0'))
```
mirgee, line 35: Should be `PROTOCOL_TYPE_ARIES`?

wrappers/node/test/helpers/entities.ts
```diff
@@ -149,7 +144,9 @@ export const credentialCreateWithMsgId = async (data?: ICredentialCreateWithMsgI
   if (!data) {
     data = await dataCredentialCreateWithMsgId()
   }
+  console.log(`Creating credential with data with msgId ${JSON.stringify(data)}`)
   const credential = await Credential.createWithMsgId(data)
+  console.log(`Created credential!`)
```
mirgee, line 149: Please remove the logs

wrappers/node/test/suite1/ariesvcx-connection.test.ts
```diff
@@ -1,13 +1,23 @@
...
+// @ts-ignore
+import {// @ts-ignore
+  INVITE_ACCEPTED_MESSAGE, // @ts-ignore
+  INVITE_DETAILS,// @ts-ignore
+  INVITE_REDIRECTED_MESSAGE,
+  PROTOCOL_TYPE_ARIES
+} from 'helpers/test-constants'
+// @ts-ignore
 import { initVcxTestMode, shouldThrow, sleep } from 'helpers/utils'
+// @ts-ignore
```
mirgee, line 16: Just out of curiosity, what was the issue with the compiler?

wrappers/node/test/suite1/ariesvcx-connection.test.ts
```diff
@@ -1,13 +1,23 @@
...
 import { connectionCreate, connectionCreateConnect, dataConnectionCreate } from 'helpers/entities'
-import { INVITE_ACCEPTED_MESSAGE, INVITE_REDIRECTED_MESSAGE, INVITE_DETAILS } from 'helpers/test-constants'
+// @ts-ignore
+import {// @ts-ignore
+  INVITE_ACCEPTED_MESSAGE, // @ts-ignore
+  INVITE_DETAILS,// @ts-ignore
+  INVITE_REDIRECTED_MESSAGE,
+  PROTOCOL_TYPE_ARIES
+} from 'helpers/test-constants'
```
mirgee, line 13: Maybe we should type the constants as well?

wrappers/node/test/suite1/ariesvcx-connection.test.ts
```diff
@@ -24,25 +34,20 @@ describe('Connection:', () => {
...
       assert.notEqual(inviteDetails, '')
     })
-
-    it('success: with phone', async () => {
-      const connection = await connectionCreate()
-      const inviteDetails = await connection.connect({ data: '{"connection_type":"SMS","phone":"7202200000"}' })
-      assert.notEqual(inviteDetails, '')
-    })
-
+  //
```
mirgee, line 40: Typo

wrappers/node/test/suite1/ariesvcx-connection.test.ts
```diff
@@ -24,25 +34,20 @@ describe('Connection:', () => {
...
     it('throws: not initialized', async () => {
       const connection = new (Connection as any)()
       const err = await shouldThrow(async () => connection.connect({ data: '{"connection_type":"QR"}' }))
       assert.equal(err.vcxCode, VCXCode.INVALID_CONNECTION_HANDLE)
     })
```
mirgee, line 45: What is QR connection type? Isn't the test description incorrect?

wrappers/node/test/helpers/utils.ts
```diff
@@ -1,36 +1,70 @@
...
+  agency_did: '2hoqvcwupRTUNkXn6ArYzs',
+  institution_did: '2hoqvcwupRTUNkXn6ArYzs',
+  sdk_to_remote_did: '2hoqvcwupRTUNkXn6ArYzs',
+  remote_to_sdk_did: '2hoqvcwupRTUNkXn6ArYzs',
+  sdk_to_remote_verkey: 'FuN98eH2eZybECWkofW6A9BKJxxnTatBCopfUiNxo6ZB',
+  wallet_key: '********',
+  sdk_to_remote_role: '0',
+  wallet_key_derivation: 'RAW',
+  agency_endpoint: 'http://127.0.0.1:8080',
+  // protocol_type: '1.0'
```
mirgee, line 57: Probably doesn't need to be here?

wrappers/node/test/helpers/utils.ts
```diff
@@ -1,36 +1,70 @@
 import '../module-resolver-helper'
 
-import { VCX_CONFIG_TEST_MODE } from 'helpers/test-constants'
-import { SinonStub, stub } from 'sinon'
+// import { VCX_CONFIG_TEST_MODE } from 'helpers/test-constants'
```
mirgee, line 3: Not used even in the code which was left commented out

wrappers/node/src/api/vcx-mock.ts
```diff
@@ -10,7 +10,8 @@ export enum VCXMockMessage {
   IssuerCredentialReq = 6, // get_message response with claim req
   Proof = 7, // get_message response with proof,
   CredentialResponse = 8, // reply to credential request with an actual credential
-  AcceptInvite = 9 // connection invite was accepted
+  AcceptInvite = 9, // connection invite was accepted
+  AcceptInviteAries = 10 // aries connection invite was accepted
```
mirgee, line 14: Also probably doesn't need to be in this PR

wrappers/node/package.json
```diff
@@ -1,96 +1,111 @@
...
+    "jslint:fix": "standard --fix",
+    "lint:demo": "standard demo/*",
+    "doc-gen": "./node_modules/.bin/typedoc --out doc --excludePrivate --excludeProtected --ignoreCompilerErrors src",
+    "test": "npm run test:connection && npm run test:credentialDef && npm run test:credential && npm run test:disclosedProof && npm run test:issuerCredential && npm run test:proof && npm run test:schema && npm run test:utils && npm run test:wallet && npm run test:ffi && npm run test:logging",
+    "test:aries": "npm run test:aries:connection && npm run test:aries:credentialDef && npm run test:aries:credential && npm run test:aries:disclosedProof && npm run test:aries:issuerCredential && npm run test:aries:proof && npm run test:aries:schema && npm run test:aries:utils && npm run test:aries:wallet && npm run test:ffi && npm run test:logging",
+    "test:aries:connection": "      DISALLOW_V1_CONNECTION=true LIBVCX_FFI_ARCHITECTURE=x64 TS_NODE_PROJECT=./test/tsconfig.json NODE_ENV=test RUST_BACKTRACE=full ./node_modules/.bin/mocha --timeout 10000 -gc --expose-gc --exit --recursive --use_strict --require ts-node/register ./test/suite1/ariesvcx-connection.test.ts",
```
mirgee, line 64: I guess this can't be done more concisely... Why not set `LIBVCX_FFI_ARCHITECTURE` `x64` by default? Can you please explain where is `DISALLOW_V1_CONNECTION` used?

mirgee (changes requested):

wrappers/node/src/api/logging.ts
```diff
@@ -1,16 +1,19 @@
...
+const Struct = buildStructType(ref)
+
 export const Logger = Struct({
   flushFn: ffi.Function('void', []),
   logFn: ffi.Function('void', ['int', 'string', 'string', 'string', 'string', 'int'])
 })
 
-export function loggerToVoidPtr (_logger: any | Struct): Buffer {
+// The _logger must in fact be instance of Struct type we generated above using buildStructType(ref)
+export function loggerToVoidPtr (_logger: any): Buffer {
```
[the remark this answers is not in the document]
Patrik-Stas: The type is constructed programatically as:
```
const Struct = buildStructType(ref)
```
so to be correct in signature I'd have to write 
```
export function loggerToVoidPtr (_logger: Struct): Buffer {
```
Typescript won't allow that because it doesn't think the `Struct` is actual type. It might be possible to get it done somehow but it doesn't seem straightforward. The code as was previously passed typescript checks but didn't work at all.

wrappers/node/test/suite1/ariesvcx-connection.test.ts
```diff
@@ -174,57 +190,32 @@ describe('Connection:', () => {
...
-
-  describe('getRedirectDetails:', () => {
-    it('success', async () => {
-      const connection = await connectionCreateConnect()
-      await connection.updateStateWithMessage(INVITE_REDIRECTED_MESSAGE)
-      const details = await connection.getRedirectDetails()
-      assert.include(details, '"DID":')
-    })
-  })
-
```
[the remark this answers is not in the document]
Patrik-Stas: Only supports `1.0`/`2.0 non-aries`, connection redirect will be deleted. It's not part of Aries, also it requires support of agency, which we don't support in Node agency.

wrappers/node/test/suite1/ariesvcx-connection.test.ts
Patrik-Stas: Some legacy stuff, the `data` option is not used when running with aries. So I'll remove the QR option here. The test name is unmodified and still passing. I think the test name is correct  - test is supposed to failed because I think using `new (Connection as any)()` instantiation it never calls into FFI, only create a dumb typescript object.

wrappers/node/test/helpers/utils.ts
Patrik-Stas: touche

wrappers/node/package.json
Patrik-Stas: - Good catch, `DISALLOW_V1_CONNECTION` was leaking here from another PR, removed now.
- Originally `LIBVCX_FFI_ARCHITECTURE` was made to default to x86 for backwards compatibility, but yeah I think it makes sense to switch default to x64. 👍

wrappers/node/test/helpers/utils.ts
```diff
@@ -1,36 +1,70 @@
...
 }
 
-export const initVcxTestMode = async () => {
-  patchInitVCX()
-  await vcx.initVcx(VCX_CONFIG_TEST_MODE)
+export async function initVcxTestMode (protocolType: string) {
+  const useTestConfig = { ... testConfig, protocol_type: protocolType}
+  console.debug('Patching vcx init')
+  // patchInitVCX()
+  console.debug('Patched vcx init')
```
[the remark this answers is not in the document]
Patrik-Stas: I deleted the whole patch thing completely now.

wrappers/node/test/suite1/ariesvcx-proof.test.ts
Patrik-Stas: They surely are part of serialized proof object. The question is where is the data stored once it's deserialized - rust layer or nodejs layer. 

I'd like to keep the comment because it's something to resolve in the future / next steps.

wrappers/node/test/suite1/ariesvcx-connection.test.ts
Patrik-Stas: good question - it was not necessary anymore. I've used these at the time I've had portions of code commented out and TS was complaining the imports are not being used.

wrappers/node/test/suite1/ariesvcx-connection.test.ts
Patrik-Stas: done

wrappers/node/test/suite1/ariesvcx-proof.test.ts
mirgee: Why do we need a new method? Can't we get the requested attributes from Rust layer by serialization?

Comments are not the place to keep things to resolve in the future, we have github issues for this.

PR #57, opened 2020-08-24
Separate android builds for device and emulators

mirgee:
Build and publish android artifacts separately for device and emulators.

Modularize into action.


.github/actions/publish-android/action.yml
```diff
@@ -0,0 +1,35 @@
...
+env:
+  DOCKER_BUILDKIT: 1
+
+runs:
+  using: "composite"
+  steps:
+    - name: Build, run android wrapper tests, and publish artifacts
+      run: |
+        set -x
+        # docker run --name test-android-wrapper -v $PWD:/home/indy/libvcx-absa:rw ${{ inputs.docker-img.name }} \
```
Patrik-Stas, line 26: I suppose this can be removed?

mirgee: Yes, we want to reduce the docker image size by mounting the source as volume instead of copying it in at buildtime, as described in this [issue](https://github.com/AbsaOSS/libvcx/issues/49).

PR #60, opened 2020-08-24
Simplify controlflow when mocking enabled

Patrik-Stas:


mirgee:
Can you please briefly explain why were these mocked responses superfluous? There doesn't seem to be more mocks further in in control flow...

Patrik-Stas:
@mirgee I don't understand why the code is there. It's trying to do this early return of 
```
Ok(Vec::new());
```
but I'd expect that to just happen naturally when empty vector (the `response`) is passed to the parse methods. 
Not tests are broken deleting this, so I think it's really unnecessary logic.

PR #65, opened 2020-08-26
Migrate tests to 3.0 and 4.0 protocol_type

Patrik-Stas:
- Reads environment variable `DISALLOW_V1 `. If set to `true`, code will panic on attempt to create v1 connection or to use onboarding v1
- Defaults to protocol V3
- Read environment variable `CONFIG_PROTOCOL_TYPE` - to override protocol type (useful when working on migration of 
a test from v1 to aries)
- When running in test mode, `connect_register_provision ` prints libvcx settings.
- Uses `agency_pool_tests` feature to separate tests which require both `pool` and `agency`.

Contains PR https://github.com/AbsaOSS/libvcx/pull/66
- removes support for `update_agent_info` - this was setting COM_METHOD on agency, we don't use this. We are using `update_webhook_url`
- removes support for `download_agent_messages` - is different from `download_messages`. The download_agent_messages was trying to download sent messages from agency - this was valid use case only on v1 agency, not with aries anymore.
also delete redirection test to reduce noise, we won't be supporting that, it's also legacy
- turned test test_update_agent_info_real into `test_update_agent_webhook_real` and activated it. I know we should maybe fetch the value from agency and compare it, but let's leave that for next phase. Let's just do necessary to get those tests passing now.
- remove `test test_download_agent_messages` - was testing `download_agent_messages` feature which only applies to V1 agency.
- the removed test was replaced by `test_send_and_download_messages` which tests downloading received messages.


libvcx/src/connection.rs
```diff
@@ -1353,37 +1358,38 @@ pub mod tests {
 
     pub fn create_connected_connections() -> (u32, u32) {
```
Patrik-Stas, line 1359: This is used only be tests `agency_pool_tests` tests. Those tests are still not passing due to some issues, however this is their shared step and changes in this PR fixes run of `create_connected_connections`. Previously this method assumed legacy connections

PR #70, opened 2020-08-27
Set comments on cred offer, credential as issuer

Patrik-Stas:
We'll probably be able to close this, instead of merging this we can migrate our client to 0.9.0 vcx. Just want to run CI with this.


mirgee:
Please, run CI on a branch next time.

PR #74, opened 2020-08-30
Migrate connections API tests to Aries

mirgee:


libvcx/src/api/connection.rs
```diff
@@ -276,21 +276,21 @@ pub extern fn vcx_connection_create_with_invite(command_handle: CommandHandle,
...
-/// # Examples connection_options ->
+/// Examples connection_options ->
 /// "{"connection_type":"SMS","phone":"123","use_public_did":true}"
 ///     OR:
 /// "{"connection_type":"QR","phone":"","use_public_did":false}"
 ///
 /// cb: Callback that provides error status of request
 ///
-/// #Returns
+/// eturns
```
Patrik-Stas, line 293: Seems like mass refactor accident :-) There's multiple of these

libvcx/src/api/connection.rs
```diff
@@ -120,8 +120,8 @@ use indy_sys::CommandHandle;
         Invitation - https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol#0-invitation-to-connect
         ConnectionRequest - https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol#1-connection-request
         ConnectionResponse - https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol#2-connection-response
-        ConnectionProblemReport - https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol#error-message-example
-        Ack - https://github.com/hyperledger/aries-rfcs/tree/master/features/0015-acks#explicit-acks
+        ConnectionProblemReport - https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocolrror-message-example
+        Ack - https://github.com/hyperledger/aries-rfcs/tree/master/features/0015-acksxplicit-acks
```
Patrik-Stas, line 124: I think the `#` was deleted by accident here

libvcx/src/api/connection.rs
```diff
@@ -5,15 +5,15 @@ use utils::threadpool::spawn;
...
 use indy_sys::CommandHandle;
+use v3::messages::a2a::A2AMessage;
 
 /*
     Tha API represents a pairwise connection with another identity owner.
     Once the connection, is established communication can happen securely and privately.
     Credentials and Presentations are exchanged using this object.
 
-    # States
+    States
```
Patrik-Stas, line 16: Just wondering, what's the reason behind removing all these `#` from docs?

libvcx/src/utils/constants.rs
```diff
@@ -55,6 +57,7 @@ pub const CREDENTIAL_RESPONSE: &'static [u8; 3686] = &[129, 167, 98, 117, 110, 1
...
...dq3iDRvUIbTrt7O0-22jbmsEJ1mOQ64pitfgNKQ0jkUpFoBiDuvmlA0RtDk8azkQpBNtgx1naykrHTIfGWxD4acYUaaN9mhUIWvkzaGOE0tPT-6v_oD-mGy4eU-qaP3DNxTJRn_FMWiGmwmsWee_GUL-8qTtLAeby_J9MHJli7Wt80M0MOElCkNkW6pQuMMC9IPYadvjZeMH8NFmOh5cGRZlBGWTzlFXGFwowvBr9DrjQ2sgZB4xGkG2owdi-VB676CboGO-GdPkvXnNeubZwOVtEW1W3u5_3nHnjIEyClN7lesizpy8PUI8Ud5sqe3mZUEyO9a_ywtOFkjF8xzTyoRK-fZj6XG6qqdjUo4BpcNv9Kt56GpigEKeQQ1mQfWIJROskbe_nJ5j_vc8HxsEnrB5-Nl2xEAZFwN5CSREN5FXaPIXE1U9Xp0HQgdOZ3_9D48twqAMvV2asTjlgRAxUh8uf0mwxg=","iv":"_AmDnIVEnS2-m14K","protected":"eyJlbmMiOiJ4Y2hhY2hhMjBwb2x5MTMwNV9pZXRmIiwidHlwIjoiSldNLzEuMCIsImFsZyI6IkF1dGhjcnlwdCIsInJlY2lwaWVudHMiOlt7ImVuY3J5cHRlZF9rZXkiOiJsbzRrM0R2RzJXcnZJNTJNNTVaM2tZdXJQX0pZd2gtcUtHak5uRkE4YjZpbmJNRVBpbkUxT1Nwb2dpY2FHSGNPIiwiaGVhZGVyIjp7ImtpZCI6IjRQWGFDU1BUQVpxRFRIaDRNNnYzWlExdVc1d2hQUW5EeDMyYkp0Q3lvSjhWIiwiaXYiOiJjdUhLcUZQMlFiN0s4djJxTEFya0VLOHpRZHFkS1dLMSIsInNlbmRlciI6IkY3S2FacmhVczc0bDg2OFAzX0swYk9QVzU2eFNNc0t5M1ZPV3FGTlpsMVprOGV0UUlfZUs4Q3JXUUktT1NZU3ZlVzRHNjRZOWFnRHlDS3ZHbWN1dWtieURxOV9KVUI2Tlk1SzNIV2pjalBUVk9qZnpEX2NpV21DNTZXTT0ifX1dfQ==","tag":"GLUAUyEK7oOfB1Qyk05Ubg=="},"senderDID":"","statusCode":"MS-103","type":"aries","uid":"0c6bd83f-1fd2-441d-a0b9-3293536afdb3"}]}"#;
```
Patrik-Stas, line 60: Could you document somewhere how to reproduce this data? There's e2e encrypted data inside this mock, it would be good to know what the e2e encryption keys were

libvcx/src/v3/utils/encryption_envelope.rs
```diff
@@ -62,20 +64,24 @@ impl EncryptionEnvelope {
     }
 
     pub fn open(payload: Vec<u8>) -> VcxResult<A2AMessage> {
-        let unpacked_msg = crypto::unpack_message(&payload)?;
+        trace!("EncryptionEnvelope::open >>> payload: {:?}", payload);
```
Patrik-Stas, line 67: Forgotten log message? Will probably print just a byte array, no?

libvcx/src/v3/utils/encryption_envelope.rs
```diff
@@ -62,20 +64,24 @@ impl EncryptionEnvelope {
...
-        let message: A2AMessage = ::serde_json::from_str(&message)
-            .map_err(|err| {
-                VcxError::from_msg(VcxErrorKind::InvalidJson, format!("Cannot deserialize A2A message: {}", err))
-            })?;
+            _message["message"].as_str()
+                .ok_or(VcxError::from_msg(VcxErrorKind::InvalidJson, "Cannot find `message` field"))?.to_string()
+        };
 
-        Ok(message)
+        Ok(::serde_json::from_str(&message)
```
Patrik-Stas, line 82: I think I prefer the explicitness of `let message: A2AMessage = :` in old code, but leaving it up to your judgement.

libvcx/src/v3/handlers/connection/connection.rs
```diff
@@ -169,6 +169,13 @@ impl Connection {
     pub fn send_message(&self, message: &A2AMessage) -> VcxResult<()> {
         trace!("Connection::send_message >>> message: {:?}", message);
 
+        match self.state_object() {
```
Patrik-Stas, line 172: Won't this be producing warn messages during within the connection protocol itself?

libvcx/src/connection.rs
```diff
@@ -908,14 +905,14 @@ impl Connection {
...
+
     CONNECTION_MAP.get_mut(handle, |connection| {
         match connection {
-            Connections::V1(ref mut connection) => {
-                connection.update_state(message.clone())
-            }
+            Connections::V1(_) => Err(VcxError::from(VcxErrorKind::ActionNotSupported)),
             Connections::V3(ref mut connection) => {
-                connection.update_state(message.as_ref().map(String::as_str))?;
+                connection.update_state(message.clone())?;
```
Patrik-Stas, line 915: It seems that the `message` could be just moved into `update_state`?

libvcx/src/v3/utils/encryption_envelope.rs
mirgee: Can you please be more specific?

libvcx/src/v3/utils/encryption_envelope.rs
mirgee: In the old code, the expliciteness is paid off by unnecessary duplication in my opinion.

libvcx/src/connection.rs
```diff
@@ -854,25 +856,20 @@ pub fn send_generic_message(connection_handle: u32, msg: &str, msg_options: &str
     })
 }
 
-pub fn update_state_with_message(handle: u32, message: Message) -> VcxResult<u32> {
+pub fn update_state_with_message(handle: u32, message: A2AMessage) -> VcxResult<u32> {
```
Patrik-Stas, line 859: Nice refactor, narrowing down necessary information from `Message` which contains agency information to only `A2AMessage`, hence only protocol messages, forgetting existence of Agency.

libvcx/src/v3/utils/encryption_envelope.rs
```diff
@@ -62,20 +64,24 @@ impl EncryptionEnvelope {
     }
 
     pub fn open(payload: Vec<u8>) -> VcxResult<A2AMessage> {
-        let unpacked_msg = crypto::unpack_message(&payload)?;
+        trace!("EncryptionEnvelope::open >>> payload: {:?}", payload);
 
-        let message: ::serde_json::Value = ::serde_json::from_slice(unpacked_msg.as_slice())
-            .map_err(|err| VcxError::from_msg(VcxErrorKind::InvalidJson, format!("Cannot deserialize message: {}", err)))?;
+        let message = if AgencyMockDecrypted::has_decrypted_mock_messages() {
```
Patrik-Stas, line 69: I think I'd rather be in favor of having separate "Mock container" for usage in this place. Currently it might be harder to understand  where the data from `AgencyMockDecrypted ` will be really used when I set it up in test. I'd suggest something like `ConnectionMessageMockDecrypted`. Leaving to your judgement, just consider this idea.

wrappers/node/demo/faber.js
```diff
@@ -138,6 +138,9 @@ async function runFaber (options) {
       logger.info(`proofState=${proofProtocolState}`)
     }
 
+    logger.info('Deleting connection...')
+    await connectionToAlice.delete()
```
Patrik-Stas, line 142: Good idea to include this call, but `connectionToAlice` is still being used below in ` await vcxProof.getProof(connectionToAlice)`

libvcx/src/v3/handlers/connection/connection.rs
```diff
@@ -102,22 +103,25 @@ impl Connection {
...
 
         if let Some((uid, message)) = self.connection_sm.find_message_to_handle(messages) {
+            trace!("Connection::update_state >>> handling message uid: {:?}", uid);
             self.handle_message(message.into())?;
-            agent_info.update_message_status(uid)?;
-        };
-
-        if let Some(prev_agent_info) = self.connection_sm.prev_agent_info().cloned() {
+            self.agent_info().clone().update_message_status(uid)?;
+        } else if let Some(prev_agent_info) = self.connection_sm.prev_agent_info().cloned() {
```
Patrik-Stas, line 111: Previously this `prev_agent_info` could execute even if previous executed, now can't. Why is it safe to change this logic? I have suspect this might had to do with redirection feature?

libvcx/src/v3/utils/encryption_envelope.rs
Patrik-Stas: Oh I see... this `get_next_decrypted_message` is separate method from `get_next_decrypted_response`. Then nevermind! This is good

libvcx/src/messages/mod.rs
```diff
@@ -264,6 +264,13 @@ impl<'de> Deserialize<'de> for A2AMessageV2 {
         let value = Value::deserialize(deserializer).map_err(de::Error::custom)?;
         let message_type: MessageTypeV2 = serde_json::from_value(value["@type"].clone()).map_err(de::Error::custom)?;
 
+        let message_json = serde_json::ser::to_string(&value);
+        let message_type_json = serde_json::ser::to_string(&value["@type"].clone());
+
+        debug!("Deserializing A2AMessageV2 json: {:?}", &message_json);
```
Patrik-Stas, line 270: maybe trace? There's similar output on line 447 and it's trace. Also, it'd be better to embed `serde_json::ser::to_string(&value);` into the `debug!` macro, so we could avoid serialization if the value won't ever be used (if we've enabled `info` logs only).
Another option would be something like this
```
/// Determines if a message logged at the specified level in that module will
/// be logged.
///
/// This can be used to avoid expensive computation of log message arguments if
/// the message would be ignored anyway.
///
/// # Examples
///
/// ```edition2018
/// use log::Level::Debug;
/// use log::{debug, log_enabled};
///
/// # fn foo() {
/// if log_enabled!(Debug) {
///     let data = expensive_call();
///     debug!("expensive debug data: {} {}", data.x, data.y);
/// }
/// if log_enabled!(target: "Global", Debug) {
///    let data = expensive_call();
///    debug!(target: "Global", "expensive debug data: {} {}", data.x, data.y);
/// }
/// # }
/// # struct Data { x: u32, y: u32 }
/// # fn expensive_call() -> Data { Data { x: 0, y: 0 } }
/// # fn main() {}
/// ```
```

libvcx/src/v3/handlers/connection/connection.rs
mirgee: Resolved via DMs.

libvcx/src/utils/constants.rs
mirgee: Resolved via DM.

libvcx/src/v3/handlers/connection/connection.rs
mirgee: Send message is not used during the connection protocol.

libvcx/src/connection.rs
mirgee: Used a reference instead since references should be more lightweight.

PR #79, opened 2020-09-02
Fix failing connection tests

mirgee:


libvcx/src/connection.rs
```diff
@@ -1374,7 +1129,7 @@ pub mod tests {
     #[test]
     #[cfg(feature = "general_test")]
     fn test_parse_acceptance_details() {
```
mirgee, line 1131: This needs to be refactored to use V2 protocol, along with `parse_acceptance_details`.

PR #87, opened 2020-09-11
Improve demo and agent logging, refactor

Patrik-Stas:


Patrik-Stas:
Created mostly to make mockdata generation easier, on the way found some agent-core methods would be better bit refactored.

mirgee (approved):

PR #114, opened 2020-09-25
Move java demo to wrapper dir

mirgee:


wrappers/java/demo/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -0,0 +1,13 @@
+arguments=
```
Patrik-Stas, line 1: seems like this was commited by a mistake

PR #119, opened 2020-09-28
Generate code coverage based on all tests

mirgee:


mirgee:
Waiting for unstable flags used in the job to be fixed on nightly.

Patrik-Stas:
Reflecting back on this, I feel like the best way to structure coverage would have it separate for unit tests and separate for integration tests. Having those heatmaps side by side for both would be pretty nice

mirgee:
@Patrik-Stas I absolutely agree and planned it that way as you can see by the commented code in GA. With this PR, I was just momentarily curious what is the TOTAL code coverage.

PR #120, opened 2020-09-28
Split connection state machines

Patrik-Stas:
Created this as separate PR into `v3/restructure-files` to make these changes easier to digest.

- Splits `DidExchangeStateSM` state machine into 2 separate state machines for inviter and invitee
- Updated mocks of serialized connections (aligned with recent changes to `agent info`, updated to `3.0` serialization format
- Added tests for serialization/deserialization
- File `v3/handlers/connection/connection.rs` serves as proxy interface to preserve shared inviter/invitee interface. 
- Removes behaviour where `inviter` was trying to find processable message in its bootstrap connection agent https://github.com/AbsaOSS/libvcx/pull/120/files#diff-8fdb72a4f52f111ea14ed40393e84bd8L102 
Though I am not sure 100% what original intention was, I suppose it was added to behave forgivingly towards Aries agents which would (incorrectly) ignore change in DDO (after `Connection Response` message is sent). This change should not cause any problems between VCX based agents.
- Modifies behaviour of `get_invite_details` https://github.com/AbsaOSS/libvcx/pull/120/files#diff-8fdb72a4f52f111ea14ed40393e84bd8R228 such that actual invitation object is only returned for Inviter SM in `InviterState::Invited` state. Otherwise empty object is returned. Previously the code was attempting to construct invitation from counterparty did_doc which doesn't seem make much sense. In case of invitee, it would actually produce invitation different than the one used to bootstrap the connection, as the DDO of its counterparty might change upon receiving `connection response`.


libvcx/src/connection.rs
```diff
@@ -418,6 +425,54 @@ pub mod tests {
...
 
+    #[test]
+    #[cfg(feature = "general_test")]
+    fn test_serialize_serialize_connection_inviter_requested() {
+        let _setup = SetupAriesMocks::init();
+
+        let handle = build_test_connection_inviter_invited();
+        let serialized = to_string(handle).unwrap();
+
+        info!("serialized={}", serialized);
```
mirgee, line 436: You probably forgot to remove this

libvcx/src/v3/handlers/connection/connection.rs
```diff
@@ -3,88 +3,268 @@ use std::collections::HashMap;
...
-        } else {
-            Ok(json!({}).to_string())
+        match &self.connection_sm {
+            SmConnection::Inviter(sm_inviter) => {
+                let invitation = sm_inviter.get_invitation()
+                    .ok_or(VcxError::from_msg(VcxErrorKind::ActionNotSupported, format!("Can't retrieve invitation")))?;
+                Ok(json!(invitation.to_a2a_message()).to_string())
+            }
+            SmConnection::Invitee(_sm_invitee) => {
+                Ok(json!({}).to_string())
```
mirgee, line 235: TODO: Throw an error instead

libvcx/src/v3/handlers/connection/connection.rs
```diff
@@ -3,88 +3,268 @@ use std::collections::HashMap;
...
+    Invitee operation
+     */
     pub fn process_invite(&mut self, invitation: Invitation) -> VcxResult<()> {
         trace!("Connection::process_invite >>> invitation: {:?}", invitation);
         self.step(DidExchangeMessages::InvitationReceived(invitation))
     }
 
+    /**
+    If called on Inviter, returns invitation to connect with him
+    If called on Invitee, returns invitee supplied at creation
```
mirgee, line 224: TODO: Fix docstring

libvcx/src/v3/handlers/connection/connection.rs
```diff
@@ -95,32 +275,35 @@ impl Connection {
...
-            trace!("Connection::update_state >>> handling obtained messages: {:?}", messages);
-            if let Some((uid, message)) = self.connection_sm.find_message_to_handle(messages) {
-                self.handle_message(message.into())?;
-                prev_agent_info.update_message_status(uid)?;
-            }
-        } else {
-            trace!("Connection::update_state >>> found no message to handle");
         }
 
+        // Todo: decide what to do with this. Previously inviter was also trying to fetch messages to handle
```
mirgee, line 284: TODO: Let's get this back so that inviter may receive ProblemReport

libvcx/src/v3/handlers/connection/connection.rs
```diff
@@ -306,6 +298,10 @@ impl Connection {
      */
     pub fn update_state_with_message(&mut self, message: &A2AMessage) -> VcxResult<()> {
         trace!("Connection: update_state_with_message: {:?}", message);
+        if self.is_in_null_state() {
+            warn!("Connection::update_state_with_message :: update state on connection in null state is ignored");
+            return Ok(());
+        }
```
mirgee, line 304: From the POV of the state machine, transitions can still occur even from the null state... Not critical, just a note that this seems like a hack

libvcx/src/v3/handlers/connection/connection.rs
```diff
@@ -258,45 +258,37 @@ impl Connection {
...
+            self.update_state_with_message(&message);
             self.agent_info().clone().update_message_status(uid)?;
+        } else if let SmConnection::Inviter(sm_inviter) = &self.connection_sm {
+            trace!("Connection::update_state >>> Inviter found no message to handel on main connection agent. Will check bootstrap agent.");
+            if let Some((messages, bootstrap_agent_info)) = sm_inviter.get_bootstrap_agent_messages()? {
+                if let Some((uid, message)) = self.find_message_to_handle(messages) {
+                    trace!("Connection::update_state >>> handling message found on bootstrap agent uid: {:?}", uid);
+                    self.update_state_with_message(&message);
+                    bootstrap_agent_info.update_message_status(uid)?;
+                }
```
mirgee, line 288: Maybe we should add a one line comment explaining why this is done?

libvcx/src/api/connection.rs
```diff
@@ -308,38 +308,30 @@ pub extern fn vcx_connection_connect(command_handle: CommandHandle,
         return VcxError::from(VcxErrorKind::InvalidConnectionHandle).into();
     }
 
-    let options = if !connection_options.is_null() {
+    let _options = if !connection_options.is_null() {
```
mirgee, line 311: Why is this defined but not used?

mirgee (approved):

libvcx/src/v3/handlers/connection/connection.rs
Patrik-Stas: preserved behaviour, this check was previously happening as well, but in difference place. It was taking place in `update_state`. Unlike now, `update_state_with_message `  was previously never called directly outside `connection.rs`

mirgee: I know :) Discussed via IM

PR #123, opened 2020-09-30
Add local revocation integration test

mirgee:


libvcx/src/lib.rs
```diff
@@ -287,22 +319,39 @@ mod tests {
...
-        info!("test_basic_revocation :: prover :: retrieved credential converted to selected: {}", &selected_credentials_str);
-        generate_and_send_proof(proof_handle_prover, faber, &selected_credentials_str);
+    #[cfg(feature = "agency_pool_tests")]
+    #[test]
+    fn test_local_revocation() {
+        let _setup = SetupLibraryAgencyV2ZeroFees::init();
+
+        let institution_did = settings::get_config_value(settings::CONFIG_INSTITUTION_DID).unwrap();
+        let (faber, alice) = ::connection::tests::create_connected_connections();
+        let (schema_id, cred_def_id, rev_reg_id, _cred_def_handle, credential_offer) = _issue_address_credential(faber, alice, &institution_did);
```
Patrik-Stas, line 337: The fact that the last element is credential handle but it is called `credential_offer` is bit confusing

Patrik-Stas: Also in the `_issue_address_credential ` function itself

libvcx/src/lib.rs
```diff
@@ -237,12 +237,26 @@ mod tests {
...
         // GET REV REG DELTA BEFORE REVOCATION
         let (_, delta, timestamp) = ::utils::libindy::anoncreds::get_rev_reg_delta_json(&rev_reg_id.clone().unwrap(), None, None).unwrap();
-        println!("revoking credential");
+        info!("revoking credential");
         ::issuer_credential::revoke_credential(issuer_handle).unwrap();
         let (_, delta_after_revoke, _) = ::utils::libindy::anoncreds::get_rev_reg_delta_json(&rev_reg_id.unwrap(), Some(timestamp + 1), None).unwrap();
         assert_ne!(delta, delta_after_revoke);
     }
 
+    fn revoke_credential_local(issuer_handle: u32, rev_reg_id: Option<String>) {
```
Patrik-Stas, line 246: I think it'd be better if this function only takes `issuer_handle` as parameter and gets `rev_reg_id`  from credential itself, even if it means adding new getter on issuer credential

PR #130, opened 2020-10-07
Add more integration tests, add proof request filtering by name

mirgee:


libvcx/src/connection.rs
```diff
@@ -295,38 +295,38 @@ pub mod tests {
...
         ::utils::devsetup::set_consumer(consumer_handle);
         debug!("Consumer is going to accept connection invitation.");
-        let alice_to_faber = create_connection_with_invite("faber", &details).unwrap();
-        connect(alice_to_faber).unwrap();
-        update_state(alice_to_faber).unwrap();
-        // assert_eq!(VcxStateType::VcxStateRequestReceived as u32, get_state(faber));
+        let consumer_to_institution = create_connection_with_invite("institution", &details).unwrap();
+        connect(consumer_to_institution).unwrap();
+        update_state(consumer_to_institution).unwrap();
+        // assert_eq!(VcxStateType::VcxStateRequestReceived as u32, get_state(institution));
```
Patrik-Stas, line 311: Either remove the comment, or else I think it should be moved after institution perform update on line 316

libvcx/src/connection.rs
```diff
@@ -295,38 +295,38 @@ pub mod tests {
...
         debug!("Institution is going to create connection.");
-        ::utils::devsetup::set_institution();
-        let faber_to_alice = create_connection("alice").unwrap();
+        ::utils::devsetup::set_institution(institution_handle);
+        let institution_to_consumer = create_connection("consumer").unwrap();
         let _my_public_did = settings::get_config_value(settings::CONFIG_INSTITUTION_DID).unwrap();
-        let details = connect(faber_to_alice).unwrap().unwrap();
-        update_state(faber_to_alice).unwrap();
+        let details = connect(institution_to_consumer).unwrap().unwrap();
+        update_state(institution_to_consumer).unwrap();
```
Patrik-Stas, line 304: I believe this `update_state` doesn't perform any action. After `connect` is called, `agent_connection` is created in agency and next `update_state` only checks whether the counterparty has sent connection request message.

libvcx/src/api/disclosed_proof.rs
```diff
@@ -401,6 +401,44 @@ pub extern fn vcx_disclosed_proof_get_requests(command_handle: CommandHandle,
     error::SUCCESS.code_num
 }
 
+#[no_mangle]
+pub extern fn vcx_v2_disclosed_proof_get_requests(command_handle: CommandHandle,
```
Patrik-Stas, line 405: As I mentioned, I am not much in favor of adding this unless we have usecase as extending api is adding up future maintenance cost. I'll leave the decision up to your judgement.

libvcx/src/aries/handlers/proof_presentation/prover/prover.rs
```diff
@@ -136,9 +133,12 @@ impl Prover {
                 .filter_map(|(_, message)| {
                     match message {
                         A2AMessage::PresentationRequest(presentation_request) => {
-                            Some(presentation_request)
+                            match (presentation_request.request_presentations_attach.content().ok(), match_name) {
+                                (Some(content), Some(match_name)) => _filter_by_name(&content, match_name, presentation_request),
+                                _ => Some(presentation_request)
```
Patrik-Stas, line 138: The behaviour is bit unexpected:
- if I specify match name and the attachmed is not valid json, the proof request is still considered as matched

Now previously `match_name` was never used here, content was not inspected and it was seemingly up to the caller to inspect the content and deal with it.

I understand this code is useful for the test you are writing. But there's no external usecase for `match_name` and the behaviour is surprising. 

Here's my proposal though -  how about we extract this filtering logic as sort of util function which you could do in your test (fetch all the presentation requests messages, then filter the one you need yourself) and perhaps even expose as C-API. 
That would also be more maintainable way going into future - adding different filtering strategies would be as simple as adding a new function (which again, each could be exposed on C-API layer too)

So we would now remove `vcx_v2_disclosed_proof_get_requests` and keep the old  `vcx_vdisclosed_proof_get_requests`
Instead we would add
 - `vcx_filter_proof_requests_by_name`
and in future we could add more, like 
- ``vcx_filter_proof_requests_by_issuer` etc.

mirgee: I like the idea, but it is probably the simplest and most easily extensible option we have, and those benefits are worth it. I will expose the first `vcx_filter_*` function as part of this PR and use it in the tests.

libvcx/src/api/disclosed_proof.rs
mirgee: Resolved in another comment.

libvcx/src/aries/handlers/proof_presentation/prover/prover.rs
Patrik-Stas: cool! 👍

mirgee: However, "if I specify match name and the attachmed is not valid json, the proof request is still considered as matched" is incorrect - `_filter_by_name` will return None in that case

libvcx/src/connection.rs
mirgee: You are right, good catch!

libvcx/src/lib.rs
```diff
@@ -341,27 +343,124 @@ mod tests {
     fn test_local_revocation() {
         let _setup = SetupLibraryAgencyV2ZeroFees::init();
 
+        set_institution(None);
         let institution_did = settings::get_config_value(settings::CONFIG_INSTITUTION_DID).unwrap();
-        let (faber, alice) = ::connection::tests::create_connected_connections(None);
+        let (faber, alice) = ::connection::tests::create_connected_connections(None, None);
```
Patrik-Stas, line 348: Let's try to follow `who_to_whom` connection handle naming convetion from now on

libvcx/src/lib.rs
```diff
@@ -341,27 +343,124 @@ mod tests {
...
+
+    #[cfg(feature = "agency_pool_tests")]
+    #[test]
+    fn test_double_issuance_separate_issuer() {
+        let _setup = SetupLibraryAgencyV2ZeroFees::init();
+        let institution_did = settings::get_config_value(settings::CONFIG_INSTITUTION_DID).unwrap(); // Issuer's did
+
+        let verifier = create_institution_config();
+        let consumer = create_consumer_config();
+        let (c2v, v2c) = ::connection::tests::create_connected_connections(Some(consumer), Some(verifier));
```
Patrik-Stas, line 415: I prefer the longer version like in previous test eg. `consumer_to_verifier `

libvcx/src/lib.rs
```diff
@@ -341,27 +343,124 @@ mod tests {
...
+    }
+
+    #[cfg(feature = "agency_pool_tests")]
+    #[test]
+    fn test_double_issuance_fail() {
+        let _setup = SetupLibraryAgencyV2ZeroFees::init();
+        let institution_did = settings::get_config_value(settings::CONFIG_INSTITUTION_DID).unwrap();
+
+        let consumer = create_consumer_config();
+        let (faber, alice) = ::connection::tests::create_connected_connections(Some(consumer), None);
```
Patrik-Stas, line 441: also naming here

libvcx/src/lib.rs
```diff
@@ -371,16 +470,18 @@ mod tests {
         let consumer_1 = create_consumer_config();
         let consumer_2 = create_consumer_config();
         let consumer_3 = create_consumer_config();
-        let (faber1, alice1) = ::connection::tests::create_connected_connections(Some(consumer_1));
-        let (faber2, alice2) = ::connection::tests::create_connected_connections(Some(consumer_2));
-        let (faber3, alice3) = ::connection::tests::create_connected_connections(Some(consumer_3));
+        let (faber1, alice1) = ::connection::tests::create_connected_connections(Some(consumer_1), None);
```
Patrik-Stas, line 473: and here ^_^

libvcx/src/lib.rs
```diff
@@ -341,27 +343,124 @@ mod tests {
...
+        let proof_handle_verifier = _verifier_create_proof_and_send_request(&institution_did, &schema_id, &cred_def_id, v2c, Some(verifier), request_name2);
+        _prover_select_credentials_and_send_proof(c2v, Some(consumer), request_name2);
+        set_institution(Some(verifier));
+        proof::update_state(proof_handle_verifier, None, None).unwrap();
+        assert_eq!(proof::get_proof_state(proof_handle_verifier).unwrap(), ProofStateType::ProofValidated as u32);
+    }
+
+    #[cfg(feature = "agency_pool_tests")]
+    #[test]
+    fn test_double_issuance_fail() {
```
Patrik-Stas, line 436: fix name test

libvcx/src/utils/devsetup.rs
```diff
@@ -626,12 +626,55 @@ pub fn create_consumer_config() -> u32 {
...
+            "path": constants::GENESIS_PATH.to_string(),
+            "protocol_type": "4.0"
+        });
+
+    debug!("create_institution_config >> Going to provision enterprise using config: {:?}", &config);
+    let enterprise_config = ::messages::agent_utils::connect_register_provision(&config.to_string()).unwrap();
+
+    let handle = CONFIG_STRING.add(config_with_wallet_handle(&enterprise_wallet_name, &enterprise_config.to_string())).unwrap();
+
+    set_institution(Some(handle));
```
Patrik-Stas, line 664: extract assigning role on ledger into function

Patrik-Stas (changes requested):

Patrik-Stas (approved):

PR #149, opened 2020-10-18
Refactor messages/agency_comm module, delete legacy proof

Patrik-Stas:
Bunch of cleanusp and refactors packed here:
- Deleted legacy proof code (code around `ProofRequestMessage`)
- Renamed `messages` to `agency_comm` to clearly signal the module is dealing with communication with agency
- Took `src/disclosed_proof_utils.rs` and `src/proof_utils` and along some still-used-datastructures which was located along with legacy proofs `src/messages/proofs/*.rs` restructured them into newly created module `libindy` (see `src/libindy/**`). 
- Also moved `utils/libindy` to `libindy/utils`. I think all libindy wrapping code should be in one place to be able to modularize credentials/ledgers and possibly swap implementations in the future. So this would be just a beginning, there still fragments of libindy wrapping code in `utils/` and other part of codebase
- Started breaking up global settings - I've created `src/agency_comm/agency_settings.rs` which looks a lot like the `src/settings.rs`, but only deals with `CONFIG_AGENCY_ENDPOINT, CONFIG_AGENCY_DID, CONFIG_AGENCY_VERKEY, CONFIG_REMOTE_TO_SDK_DID, CONFIG_REMOTE_TO_SDK_VERKEY, CONFIG_SDK_TO_REMOTE_DID, CONFIG_SDK_TO_REMOTE_VERKEY`. 

  Removing dependency on ::settings us closer toward making `agency_comm`  standalone module which could be used without using the rest of `aries-vcx`. 
  
  Naturally I've removed these from the `src/settings.rs`. To keep things working from user perspective, some calls into `src/settings.rs`are propagated further to `src/agency_comm/agency_settings.rs` such as `fn validate_config`, `fn set_testing_defaults, fn clear_config, fn process_config_string`.


mirgee (changes requested):
I think `agency_vcx` is way more confusing and inaccurate than `messages`. How about something like `comm`? Or, if you want to mention agency, than something like `agency_comm` or `agency_messages`... This is just throwing ideas out there.

libvcx/src/agency_comm/payload.rs
```diff
@@ -1,11 +1,10 @@
 use serde_json::Value;
 
+use agency_comm::get_message::MessagePayload;
+use agency_comm::message_type::*;
```
mirgee, line 4: I would prefer to avoid using `::*` in imports

libvcx/src/agency_comm/util.rs
```diff
@@ -0,0 +1,9 @@
+use agency_comm::agency_settings;
+use error::VcxResult;
+use utils;
+
+//Todo: change this RC to a u32
```
mirgee, line 5: There is no rc in here, also, I would name this file `utils.rs`, or maybe `utils/comm.rs` (and move `agent_utils` to `utils/agent.rs`)

libvcx/src/libindy/proofs/proof_request.rs
```diff
@@ -5,80 +5,10 @@ use serde_json;
...
-
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
-pub struct NonRevokedInterval {
-    pub from: Option<u64>,
-    pub to: Option<u64>,
-}
+use utils::validation;
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub struct ProofRequestData {
```
mirgee, line 14: I would say `ProofRequestData` belongs to `aries/messages/proof_presentation/presentation_request.rs`

libvcx/src/libindy/proofs/proof_request_internal.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub issuer_did: Option<String>,
+    pub cred_def_id: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+#[serde(untagged)]
+pub enum Restrictions {
+    V1(Vec<Filter>),
+    V2(::serde_json::Value),
+}
```
mirgee, line 16: Do we need to use both?

libvcx/src/libindy/proofs/proof_request_internal.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct PredicateInfo {
+    pub name: String,
+    //Todo: Update p_type to use Enum
+    pub p_type: String,
+    pub p_value: i32,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
mirgee, line 24: I wonder why is `skip_serializing_if` necessary?

libvcx/src/agency_comm/create_key.rs
```diff
@@ -57,55 +54,36 @@ impl CreateKeyBuilder {
...
-    pub fn version(&mut self, version: &Option<ProtocolTypes>) -> VcxResult<&mut Self> {
-        self.version = match version {
-            Some(version) => version.clone(),
-            None => settings::get_protocol_type()
-        };
-        Ok(self)
-    }
-
     pub fn send_secure(&self) -> VcxResult<(String, String)> {
         trace!("CreateKeyMsg::send >>>");
```
mirgee, line 58: This is not `CreateKeyMsg` anymore

libvcx/src/libindy/proofs/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod prover;
+pub mod verifier;
+pub mod proof_request;
+mod proof_request_internal;
```
mirgee, line 4: Good idea separating public / private data structures 👍

libvcx/src/settings.rs
```diff
@@ -11,21 +11,15 @@ use serde_json::Value;
 use strum::IntoEnumIterator;
 use url::Url;
 
+use agency_comm::agency_settings;
+use agency_comm::agency_settings::process_agency_config_string;
```
mirgee, line 15: This does not need to be two imports

libvcx/src/settings.rs
```diff
@@ -438,26 +383,16 @@ impl ::std::string::ToString for ProtocolTypes {
...
-            })
-        })
-        .into()
-}
-
 pub fn clear_config() {
     trace!("clear_config >>>");
     let mut config = SETTINGS.write().unwrap();
     config.clear();
+    agency_settings::clear_config_agency()
```
mirgee, line 390: One of the reasons all settings should share a module, IMO

mirgee (changes requested):

libvcx/src/agency_comm/agency_settings.rs
```diff
@@ -0,0 +1,173 @@
...
+    CONFIG_REMOTE_TO_SDK_VERKEY,
+    CONFIG_SDK_TO_REMOTE_DID,
+    CONFIG_SDK_TO_REMOTE_VERKEY,
+    CONFIG_ENABLE_TEST_MODE
+];
+
+
+pub static DEFAULT_DID: &str = "2hoqvcwupRTUNkXn6ArYzs";
+pub static DEFAULT_VERKEY: &str = "FuN98eH2eZybECWkofW6A9BKJxxnTatBCopfUiNxo6ZB";
+pub static DEFAULT_URL: &str = "http://127.0.0.1:8080";
```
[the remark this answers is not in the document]
Patrik-Stas: Though I disagree about placement to `settings.rs` as the idea is to make `agency_comm` standalone, thanks to your comment I've noticed that these things actually  do not need to be static at all. These are only used within `set_testing_defaults_agency` function in `agency_settings.rs`, so I am moving it right there.

libvcx/src/agency_comm/agency_settings.rs
```diff
@@ -0,0 +1,173 @@
...
+        warn!("Agency settings do not recognize setting key {}. Will be ignored.", key);
+    } else {
+        AGENCY_SETTINGS
+            .write().unwrap()
+            .insert(key.to_string(), value.to_string());
+    }
+}
+
+
+pub fn agency_mocks_enabled() -> bool {
```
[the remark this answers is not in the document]
Patrik-Stas: Touche, good catch.

libvcx/src/agency_comm/agency_settings.rs
```diff
@@ -0,0 +1,173 @@
...
+pub fn clear_config_agency() {
+    trace!("clear_config_agency >>>");
+    let mut config = AGENCY_SETTINGS.write().unwrap();
+    config.clear();
+}
+
+pub fn validate_agency_config(config: &HashMap<String, String>) -> VcxResult<u32> {
+    trace!("validate_agency_config >>> config: {:?}", config);
+
+    // todo: Since we scope these setting to agency module, these are not really optional anymore!
```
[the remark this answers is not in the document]
Patrik-Stas: They were optional previously so I didn't modify the behaviour. The reason why they were/are optional is because in theory, on `aries-vcx` level you were supposed to have option not to use agency at all. I am not sure if that was implemented only for legacy protocols, or if that still works in the code as we have it today.
I'd suggest to keep this as is to avoid mixing behaviour modifications with refactoring and we can review this behaviour separately. What do you think?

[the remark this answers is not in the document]
Patrik-Stas: Of course, if you think about whether they are optional on `agency_comm` as soon to become standalone module, of course no way. But as per comment above, on global `aries-vcx` level, these were optional.

libvcx/src/libindy/proofs/proof_request_internal.rs
Patrik-Stas: Well, I haven't done modifications here, just moved the datastructure, however I suppose it makes sense to skip serializing the field if there's no value set.

libvcx/src/libindy/proofs/mod.rs
Patrik-Stas: yup, there was a whole bunch of functions which were basically only consumed by either `validate_indy_proof` or `generate_indy_proof`

libvcx/src/agency_comm/payload.rs
Patrik-Stas: agree

libvcx/src/settings.rs
Patrik-Stas: Disagree, this "redirection" is to simply preserve `aries-vcx` functionality as a whole, but removing dependencies on `aries-vcx` globals is precondition to create independent modules.

libvcx/src/libindy/proofs/proof_request.rs
Patrik-Stas: I would say it's rather agnostic, it's data structure which verifier creates and prover consumes, so I don't think belongs to one of them exclusively.

libvcx/src/libindy/proofs/proof_request_internal.rs
Patrik-Stas: Honestly, I don't know. It doesn't seem that any of them is directly used in code, but I suppose you can get proof request in 2 different formats which needs to be deserialized.

libvcx/src/libindy/proofs/proof_request_internal.rs
mirgee: But why is it necessary? Extra fields are ignored during deserialization...

libvcx/src/agency_comm/agency_settings.rs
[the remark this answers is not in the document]
mirgee: I didn't say anything about putting to `settings.rs`, by one place I mean e.g. two submodules of a single module.

libvcx/src/agency_comm/agency_settings.rs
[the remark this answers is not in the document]
mirgee: I agree. Making smaller changes is less risky and this refactor is large enough change already.

libvcx/src/libindy/proofs/mod.rs
mirgee: I would like to see us separating these more often.

libvcx/src/settings.rs
mirgee: What do you mean by redirection? I think that on the contrary, `libvcx`, as a "user" of `agency_comm`, has the responsibility of configuring it as necessary. As I see it, `agency_settings`is a subset of `libvcx` settings, not separate from it.

libvcx/src/libindy/proofs/proof_request_internal.rs
Patrik-Stas:
```
use serde_json;

#[derive(Serialize)]
struct Data {
    a: Option<String>,
    c: Option<String>,
}

pub fn run()
{
    let data = Data { a:Some("sad".into()), c:None};
    let serialized = serde_json::to_string(&data).unwrap();
    println!("serialized={}", serialized);
}
```

```
serialized={"a":"sad","c":null}
```

libvcx/src/settings.rs
Patrik-Stas: By redirection I meant that call into the `src/settings.rs` further calls `agency_settings.rs`.

But yeah, it could be removed and code in all places would have to make calls into `agency_settings.rs` manually when needed. It was more convenient to implement this way, with "the redirection", but I also don't see this as ultimate state. I think we could eliminate `agency_settings.rs` completely in the next steps.

libvcx/src/agency_comm/agency_settings.rs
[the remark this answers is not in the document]
Patrik-Stas: Well, these particular here are default values used in agency tests, and as I've found, really just local to 1 single test.

libvcx/src/libindy/proofs/proof_request_internal.rs
mirgee: I don't see what you mean by this (I said "deserialization", not "serialization" :))...

libvcx/src/agency_comm/agency_settings.rs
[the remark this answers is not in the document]
mirgee: By "config-related constants" I meant the config keys as well. But if the plan is to get rid of `agency_settings.rs` anyways (which I agree with), let's keep it as it is for now.

libvcx/src/libindy/proofs/proof_request.rs
mirgee: Isn't that true both about `presentation_request.rs` and `ProofRequestData`? `ProofRequestData` is the content of proof request attachment, so I really think it belongs to messages. Also, it is renamed to `PresentationRequestData` there, which is unnecessary and confusing, and deserves to be refactored as well...

libvcx/src/libindy/proofs/proof_request_internal.rs
mirgee: What determines the proof request format? Let's try to use only V1 / V2 and see what breaks :)

mirgee (approved):

PR #137, opened 2020-10-14
Return / accept strictly A2AMessages

mirgee:
Signed-off-by: Miroslav Kovar <miroslavkovar@protonmail.com>

PR #168, opened 2020-10-23
Release 0.12.0

mirgee:

### DIFF
--- a/libvcx/Cargo.lock
+++ b/libvcx/Cargo.lock
@@ -652,7 +652,7 @@ dependencies = [
 
 [[package]]
 name = "libvcx"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "android_logger",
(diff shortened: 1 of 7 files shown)


Patrik-Stas:
@mirgee we should deprecate function we wanted to deprecate - old update states, message download and additionally I think we should also deprecate vcx_init_with_config and vcx_init_minimal. So there would be only 1 way of initializing - `init_core+open_pool+open_wallet`.
Then on next release we could delete these deprecated apis.

Patrik-Stas:
I would also suggest  bumping to `0.12.0`, we added couple new APIs and strictly speaking, this https://github.com/hyperledger/aries-vcx/pull/137 was actually breaking change - input which would previously be valid is not valid anymore (altough chance of someone running into that particular case are miniscule)

Patrik-Stas (approved):

PR #169, opened 2020-10-23
Remove dependencies on global settings

Patrik-Stas:
This PR is refactoring a few things. It started with attempt to extract open_pool, init_core, open_wallet logic from `api/` so that that these function would be located with other internal apis in `src/*.rs`. Having this done will enable start working on agent abstractions similar like what we have in `agents/node/vcxagent-core`
Many of wallet and pool tests started failing and it was difficult to figure out why as the logic in tests and `utils/libindy/wallet` was intertwined with reliance on global state in settings. Also when one test failed, it often caused domino effect as all wallet tests were using same wallet name. So I went down the rabbit hole and refactored the code a bit:
- renamed some functions
- removed some of the accesses to global state from `utils/libindy` and in tests. Instead, I made function interfaces more explicit. Where the function would previously read some data from global state (for example key derivation method), now it have to passed as argument
- all tests are now using `RAW` KDF ( i believe some previously were not), so testing might be somewhat faster
- Removed `AgencyModeSetup` for connection state machine tests - it was creating wallet but then mocking all calls into it - instead simple `SetupIndyMocks` works just fine for those tests.
- Similarly `SetupAriesMocks`, `SetupStrictAriesMocks` could now also be replaced by `SetupIndyMocks`
- `SetupLibraryAgencyV1, SetupLibraryAgencyV1ZeroFees, SetupLibraryAgencyZeroFees` were unused -> deleted
- `SetupWalletAndPool ` was deleted - test which needs to mock wallet and pool (`test_init_with_file`, `test_init_with_config`, `test_vcx_init_with_default_values`, `test_vcx_init_called_twice_fails`) use `SetupWallet`
`SetupPoolConfig` to achieve the same result.


Patrik-Stas:
Let's not try to merge this before release though.

libvcx/src/api/vcx.rs
```diff
@@ -885,17 +904,19 @@ mod tests {
 
     #[test]
     #[cfg(feature = "general_test")]
+    #[cfg(feature = "to_restore")] // fix this..., should be restored now
     fn test_init_after_importing_wallet_success() {
```
mirgee, line 908: Does this test work now?

libvcx/src/api/wallet.rs
```diff
@@ -891,6 +891,9 @@ pub extern fn vcx_wallet_export(command_handle: CommandHandle,
 /// according to fields provided in import_config
 /// Cannot be used if wallet is already opened (Especially if vcx_init has already been used).
 ///
+/// Note this only works for default storage type (file), as currently this function does not let
+/// you pass down information about wallet storage_type, storage_config, storage_credentials.
+///
```
mirgee, line 896: Is `vcx_wallet_import` somehow related to the rest of this PR?

libvcx/src/init.rs
```diff
@@ -0,0 +1,63 @@
...
+    open_pool_ledger(&pool_name, pool_config)
+        .map_err(|err| err.extend("Can not open Pool Ledger"))?;
+
+    info!("open_pool ::: Pool Opened Successfully");
+
+    Ok(())
+}
+
+pub fn open_as_main_wallet(wallet_name: &str, wallet_key: &str, key_derivation: &str, wallet_type: Option<&str>, storage_config: Option<&str>, storage_creds: Option<&str>) -> VcxResult<WalletHandle> {
+    trace!("open_wallet >>> wallet_name: {}", wallet_name);
```
mirgee, line 35: `open_as_main_wallet >>> ...`

libvcx/src/messages/agent_utils.rs
```diff
@@ -204,8 +206,10 @@ fn _create_issuer_keys(my_did: &str, my_vk: &str, my_config: &Config) -> VcxResu
 pub fn configure_wallet(my_config: &Config) -> VcxResult<(String, String, String)> {
     let wallet_name = get_or_default(&my_config.wallet_name, settings::DEFAULT_WALLET_NAME);
 
-    wallet::init_wallet(
+    wallet::create_and_open_as_main_wallet(
         &wallet_name,
+        &my_config.wallet_key,
+        &my_config.wallet_key_derivation.as_deref().unwrap_or("ARGON2I_INT".into()),
```
mirgee, line 212: The default KDF was extracted into a static `WALLET_KDF_DEFAULT`

libvcx/src/messages/agent_utils.rs
```diff
@@ -396,7 +400,10 @@ pub fn update_agent_webhook(webhook_url: &str) -> VcxResult<()> {
 
 fn update_agent_webhook_v2(to_did: &str, com_method: ComMethod) -> VcxResult<()> {
     info!("> update_agent_webhook_v2");
-    if settings::agency_mocks_enabled() { return Ok(()); }
+    if settings::indy_mocks_enabled() || agency_mocks_enabled() {
+        warn!("update_agent_webhook_v2 ::: Indy mocks enabled, skipping updating webhook url.");
```
mirgee, line 404: I think this function shouldn't need libindy? "Indy or agency mocks..."

libvcx/src/init.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    open_pool_ledger(&pool_name, pool_config)
+        .map_err(|err| err.extend("Can not open Pool Ledger"))?;
+
+    info!("open_pool ::: Pool Opened Successfully");
+
+    Ok(())
+}
+
+pub fn open_as_main_wallet(wallet_name: &str, wallet_key: &str, key_derivation: &str, wallet_type: Option<&str>, storage_config: Option<&str>, storage_creds: Option<&str>) -> VcxResult<WalletHandle> {
```
mirgee, line 34: These arguments occur together so often (e.g. devsetup structs) that maybe we can keep them together in a struct. On the other hand, maybe the upkeep would cost more than it's worth...

libvcx/src/api/wallet.rs
Patrik-Stas: Not really, it was just something I've found out in the process, so I've added this information to docs.

libvcx/src/messages/agent_utils.rs
Patrik-Stas: good catch

libvcx/src/api/vcx.rs
Patrik-Stas: restored

libvcx/src/init.rs
Patrik-Stas: Yeah it can appear so, however this signature is not that common (I found 4 methods with such signature). It's rather the fact that these methods are called on many places (mostly tests). I think having to build a struct to call those methods might make the calling code actually mode cumbersome.

mirgee (approved):

PR #176, opened 2020-10-29
Allow issuer to set custom tails location

mirgee:


libvcx/src/credential_def.rs
```diff
@@ -43,9 +43,43 @@ pub struct CredentialDef {
...
+    pub issuance_type: String,
+    pub max_cred_num: u32,
+    pub public_keys: serde_json::Value,
+    pub tails_hash: String,
+    pub tails_location: String
+}
+
+#[derive(Deserialize, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RevocationRegistryDefinition {
```
Patrik-Stas, line 62: Nice, it's cool great to have these structures mapped into datastructures 👍 

We could use this to type field `rev_reg_def ` inside existing structure `RevocationRegistry`. We'd just have to remove this `camelCase` annotations (also on `RevocationRegistryDefinitionValue `) and I think it should work out of the box? 

- In that case, `get_tails_hash` in `src/credential_def.rs` could be also simplified and we could rather fail fast during deserialization of credential definition.
- Also function `_replace_tails_location` could then easily be eliminated

libvcx/src/credential_def.rs
```diff
@@ -199,6 +252,8 @@ fn _create_credentialdef(issuer_did: &str,
                 anoncreds::generate_rev_reg(&issuer_did, &cred_def_id, &tails_file, max_creds, "tag1")
```
Patrik-Stas, line 252: Shouldn't we be passing `revocation_details.tails_url` as parameter instead of `tails_file`? It seems to me like current code will write on ledger the local path of issuer, instead of the public url.

edit: well I see you have tested this in `test_tails_url_written_to_ledger ` - what am I missing here?

Patrik-Stas (changes requested):

libvcx/src/aries/handlers/issuance/holder/states/finished.rs
```diff
@@ -37,4 +37,57 @@ impl FinishedHolderState {
         let credential = self.credential.as_ref().ok_or(VcxError::from_msg(VcxErrorKind::InvalidState, "No credential found"))?;
         credential.credentials_attach.content()
     }
+
+    // TODO: Avoid duplication
+    pub fn get_tails_location(&self) -> VcxResult<String> {
+        debug!("get_tails_location >>>");
+        let rev_reg_def_json = self.rev_reg_def_json.as_ref().ok_or(VcxError::from_msg(VcxErrorKind::InvalidState, "No revocation registry definition found - is this credential revokable?"))?;
```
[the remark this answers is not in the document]
mirgee: I think it is useful to be specific with the error message

agents/node/vcxagent-core/src/services/service-cred-issuer.js
```diff
@@ -60,6 +60,24 @@ module.exports.createServiceCredIssuer = function createServiceCredIssuer ({ log
     await issuerCred.revokeCredential()
   }
 
+  async function getRevRegId (issuerCredId) {
+    const issuerCred = await loadIssuerCredential(issuerCredId)
+    logger.info(`Getting rev reg id for credential ${issuerCredId}`)
+    return issuerCred.getRevRegId()
+  }
+
+  async function getTailsFile (credDefId) {
```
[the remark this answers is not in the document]
mirgee: I actually considered the latter, but didn't do it - will do

libvcx/src/aries/handlers/issuance/holder/states/finished.rs
```diff
@@ -37,4 +37,57 @@ impl FinishedHolderState {
...
+        let (_, ledger_rev_reg_def_json) = ::utils::libindy::anoncreds::get_rev_reg_def_json(cred_def_id).unwrap();
+        let ledger_rev_reg_def: serde_json::Value = serde_json::from_str(&ledger_rev_reg_def_json)
+            .map_err(|err| VcxError::from_msg(VcxErrorKind::SerializationError, format!("Cannot deserialize {:?} into Value, err: {:?}", rev_reg_def_json, err)))?;
+        let ledger_value = ledger_rev_reg_def["value"]
+            .as_object()
+            .ok_or(VcxError::from_msg(VcxErrorKind::InvalidJson, format!("The field 'value' not found on ledger rev_reg_def_json: {:?}", rev_reg_def_json)))?;
+        let ledger_tails_location = ledger_value["tailsLocation"]
+            .as_str()
+            .ok_or(VcxError::from_msg(VcxErrorKind::InvalidJson, format!("The field 'tailsLocation' not found on ledger rev_reg_def_json: {:?}", self.rev_reg_def_json)))?;
+        assert_eq!(ledger_tails_location, tails_location);
```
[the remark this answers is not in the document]
mirgee: Sorry, this is debugging leftover

libvcx/src/credential_def.rs
mirgee: This saves the local path to the wallet, but replaces with `tails_url` in `_maybe_set_url`

libvcx/src/credential_def.rs
mirgee: I don't think this is very practical, as anoncreds requires strings as inputs and also returns strings, which are not modified, except for replacing the tails location.

Patrik-Stas: Hmm I see, probably number of de/serializations would even increase. I am not sure how many changes that would be and if it's worth it, but I have a feeling it would be cleaner - I could look at `RevocationRegistry` and directly see it links to `RevocationRegistryDefinition`.
But I don't insist, we can keep as is.

libvcx/src/credential_def.rs
```diff
@@ -169,7 +208,21 @@ impl CredentialDef {
 
 fn _parse_revocation_details(revocation_details: &str) -> VcxResult<RevocationDetails> {
     serde_json::from_str::<RevocationDetails>(&revocation_details)
-        .to_vcx(VcxErrorKind::InvalidRevocationDetails, "Cannot deserialize RevocationDeltas")
+        .to_vcx(VcxErrorKind::InvalidRevocationDetails, "Cannot deserialize RevocationDetails")
+}
+
+fn _maybe_set_url(rev_reg_def_json: &str, revocation_details: &RevocationDetails, rev_reg_id: &str) -> VcxResult<String> {
+    // TODO: Deserialization into RevRegDef loses ver field
```
Patrik-Stas, line 215: If there's a `ver` field, why don't we put it on `RevocationRegistryDefinition` so we could use it?

Patrik-Stas: It would be especially nicer if we could return `RevocationRegistryDefinition ` instead of String

libvcx/src/credential_def.rs
```diff
@@ -169,7 +208,21 @@ impl CredentialDef {
...
+    let mut rev_reg_def: serde_json::Value = serde_json::from_str(&rev_reg_def_json)
+        .map_err(|err| VcxError::from_msg(VcxErrorKind::InvalidJson, format!("Invalid RevocationRegistryDefinition: {:?}, err: {:?}", rev_reg_def_json, err)))?;
+
+    if let Some(tails_url) = &revocation_details.tails_url {
+        let full_tails_location = vec![tails_url.to_string(), rev_reg_id.to_string()].join("/"); // TODO: Remove trailing slashes
```
Patrik-Stas, line 220: 1. What's the problem with slashes? Is this okay not deal with now? 
2. Why is full tails location vector?

libvcx/src/credential_def.rs
```diff
@@ -43,9 +43,43 @@ pub struct CredentialDef {
...
+pub struct RevocationRegistryDefinitionValue {
+    pub issuance_type: String,
+    pub max_cred_num: u32,
+    pub public_keys: serde_json::Value,
+    pub tails_hash: String,
+    pub tails_location: String
+}
+
+#[derive(Deserialize, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
```
Patrik-Stas, line 61: Remove `camelCase` here, if we actually happen to assign this serialized `RevocationRegistryDefinition ` as content of  `rev_reg_entry` in `RevocationRegistry`, we could interfere with how libindy anoncreds expects this data to look like.

libvcx/src/credential_def.rs
```diff
@@ -169,7 +208,21 @@ impl CredentialDef {
 
 fn _parse_revocation_details(revocation_details: &str) -> VcxResult<RevocationDetails> {
     serde_json::from_str::<RevocationDetails>(&revocation_details)
-        .to_vcx(VcxErrorKind::InvalidRevocationDetails, "Cannot deserialize RevocationDeltas")
+        .to_vcx(VcxErrorKind::InvalidRevocationDetails, "Cannot deserialize RevocationDetails")
+}
+
+fn _maybe_set_url(rev_reg_def_json: &str, revocation_details: &RevocationDetails, rev_reg_id: &str) -> VcxResult<String> {
```
Patrik-Stas, line 214: I'd be in favor of extracting the `if let Some(tails_url) = &revocation_details.tails_url {` condition out to the calling code, so this wouldn't have to be called `_maybe_*`.

libvcx/src/credential_def.rs
mirgee: 1. What do you mean by problem with slashes?
2. It isn't a vector

libvcx/src/credential_def.rs
mirgee: Personally, I wouldn't, I think this way the calling code is much simpler, and the name is still accurate.

libvcx/src/credential_def.rs
Patrik-Stas: I mean the TODO comment

libvcx/src/credential_def.rs
mirgee: You mean content of rev_reg_def? Why do you think so? Anoncreds expects camelCase

libvcx/src/credential_def.rs
Patrik-Stas: Okay then, it's not a big deal. Marking resolved

libvcx/src/credential_def.rs
Patrik-Stas: Oh I see. Nevermind then.

Patrik-Stas (approved):

Patrik-Stas:
Merging this, as we discussed we'll create another PR where behaviour of the tailsUrl should be slightly different. The `tailsUrl` provided as parameter should be exactly the location where the tails file shall be published.

PR #178, opened 2020-10-30
Ci/enhancements

Patrik-Stas:


Patrik-Stas:
Few things I suggest in this PR:
- Renames jobs like `publish-ios-wrapper` to `build-ios-wrapper` etc. I think it's more accurate, the wrapper is built and put into github cache, not published anywhere actually.
- Rename jobs like `release-ios-wrapper` to `publish-ios-wrapper`. Release is done only by the single release job we have. We also say we publish nodejs-wrapper when push it to npmjs, not release.
- Reverses dependency order for newly-called `publish-ios/andoroid-wrapper` on `make-release`. Instead, `make-release` depend on publishing of mobile artifacts. If some artifacts fails to publish, we would have incomplete release.
- Remove redundant dependencies, such as depending on `build-image-indypool` in some jobs
- Use different yaml array notation to avoid very long lines when enumerating job dependecies

.github/workflows/main.yml
```diff
@@ -1124,9 +1102,9 @@ jobs:
           asset_name: libvcx-ios-${{ needs.workflow-setup.outputs.PUBLISH_VERSION }}-device.zip
           asset_content_type: application/zip
 
-  release-ios-universal:
+  publish-ios-universal:
```
mirgee, line 1105: This name is confusing, as it uses the same word - "publish" - in a different sense than publishing docker image. I uploads *release assets*, which are used only in case a release is made.

.github/workflows/main.yml
```diff
@@ -1124,9 +1102,9 @@ jobs:
...
     runs-on: ubuntu-16.04
-    needs: [workflow-setup, build-image-indypool, build-image-libvcx, build-image-android, publish-ios-wrapper, publish-android-wrapper-device, publish-android-wrapper-emulator, make-release]
+    needs: [workflow-setup, build-ios-wrapper]
```
mirgee, line 1107: This should require `make_release` to run.

.github/workflows/main.yml
```diff
@@ -743,7 +743,7 @@ jobs:
                                 cd $HOME/agents/node/vcxagent-core && npm run test:integration)'
 
   # TODO: Add tests of iOS build
-  publish-ios-wrapper:
+  build-ios-wrapper:
```
mirgee, line 746: These jobs build and upload - publish - an an artifact. I would revert these renames.

mirgee (changes requested):

PR #182, opened 2020-11-03
Support config on rotation

mirgee:
This breaking change allows the issuer to pass revocation details on rotation, thus allowing to specify a custom tails url for each new credential definition. Moreover, the behavior introduced in #176, wherein revocation registry id is appended to the location, is reversed; thus allowing the issuer to specify the resulting tails url exactly.

When passed to `vcx_credentialdef_rotate_rev_reg_def`, the `tails_url` is a required field, whereas `max_creds` and `tails_file` are optionaly (if not specified, these are reused from when the credential definition was created).


libvcx/src/credential_def.rs
```diff
@@ -168,17 +166,22 @@ impl CredentialDef {
...
     fn get_state(&self) -> u32 { self.state as u32 }
 
-    fn rotate_rev_reg(&mut self) -> VcxResult<RevocationRegistry> {
+    fn rotate_rev_reg(&mut self, revocation_details: &str) -> VcxResult<RevocationRegistry> {
         debug!("rotate_rev_reg >>>");
-        let (tails_file, max_creds, issuer_did) = (self.get_tails_file(), self.get_max_creds(), self.issuer_did.as_ref());
+        let revocation_details = _parse_revocation_details(revocation_details)?;
+        let (tails_url, tails_file, max_creds, issuer_did) = (
+            revocation_details.tails_url.ok_or(VcxError::from_msg(VcxErrorKind::InvalidRevocationDetails, "tails_url not found in revocation details"))?,
+            revocation_details.tails_file.or(self.get_tails_file()),
```
Patrik-Stas, line 174: Not related to this PR, but `tails_file` is really confusing name when actually it's supposed to be a directory. I think we should make breaking change in some future PR and just rename it. What do you think?

agents/node/vcxagent-core/src/storage/storage-service.js
```diff
@@ -110,7 +110,7 @@ async function createStorageService (agentName) {
   async function loadDisclosedProof (name) {
     const serialized = await storageDisclosedProof.get(name)
     if (!serialized) {
-      throw Error(`Connection ${name} was not found.`)
+      throw Error(`DisclosedProof ${name} was not found.`)
```
Patrik-Stas, line 113: 👍

Patrik-Stas (approved):
This is great!

libvcx/src/credential_def.rs
mirgee: I agree with it being a confusing name, and it probably is worth making a breaking change

mirgee: I agree with it being a confusing name, and it probably is worth making a breaking change

PR #193, opened 2020-11-06
Move `agency_comm` to separate crate

mirgee:


Patrik-Stas:
- `cargo.toml` has extra dependencies (tokio-threadpool, libloading, ... probably more)
- tests should be adjust not to work with with aries. for example we have `test_update_agency_messages` which first establishes aries connection - it was convenient, but it is not necessary to test the integration agency_comm/agency. We can create agent_connection in agency using the agency_comm interface, and we know how a message should be sent to this agent_connection.
- I see you had to move bunch of stuff from `utils` aries-vcx module. I think little bit of duplication can be alright, sometimes its better than to overengineer and make module out of everything. The problem I see with this is that these functions are causing VcxError to be brough in, which seem unnecessary. I think we should return something like AgencyError from agency, and then let consumer of agency_comm module to deal with those errors (in case of vcx-aries as consumer of agency_comm, perhaps convert these errors)
- Ultimately it would be good to get rid of `agency_settings.rs` so that a consumer of could create multiple agency_comm client instances.
- How about renaming `agency_comm` to `agency_client`?

Patrik-Stas (approved):

Patrik-Stas:
@mirgee we are on a good course with this, great job!

PR #203, opened 2020-11-15
Use tokio runtime intead of tokio threadpool, update futures crate

Patrik-Stas:
Work associated with FFI calls is wrapped into closure and typically converted into `Future` and executed. 
- Previously was `tokio_threadpool` https://docs.rs/tokio-threadpool/0.1.18/tokio_threadpool/ used for execution of futures, which is an outdated crate. Its docs describe the status as:
  > This crate is deprecated in tokio 0.2.x and has been moved and refactored into various places in the tokio::runtime module of the tokio crate
- This PR migrates to `tokio` runtime instead. 
- function in `libindy::utils::*.rs` such as `ledger.rs` are using `wait`function, which was however defined only on old version `0.1.x` of `futures` crate. As I udpated `futures` in `Cargo.toml`, these files started to throw compile errors. So instead, I explicitly use `futures` crate brough in by indy FFI wrapper. Such as: `use indy::future::Future;`
- The only reason why we even have `futures` crate as dependency in `Cargo.toml` is to use `future::lazy` in `runtime.rs` (everywhere else `std::future` is enough). Perhaps once this https://doc.rust-lang.org/nightly/std/lazy/struct.Lazy.html stabilizes, we might even remove it as dependency.
- This is preparation to turn `aries-vcx` into crate with async api.


libvcx/src/init.rs
```diff
@@ -38,7 +34,7 @@ pub fn init_core(config: &str) -> VcxResult<()> {
     info!("init_core >>> config = {}", config);
     settings::process_config_string(&config, true)?;
     settings::log_settings();
-    utils::threadpool::init(None);
+    utils::runtime::init_runtime(ThreadpoolConfig { num_threads: Some(4) });
```
mirgee, line 37: I think we should add the number of threads to config (with a default value)

mirgee (changes requested):

libvcx/src/utils/runtime.rs
```diff
@@ -0,0 +1,79 @@
...
+pub static mut TP_HANDLE: u32 = 0;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ThreadpoolConfig {
+    pub num_threads: Option<usize>,
+}
+
+pub fn init_runtime(config: ThreadpoolConfig) {
+    if config.num_threads == Some(0) {
+        warn!("init_runtime >> threadpool_size was set to 0; every FFI call will executed on a new thread!");
```
mirgee, line 30: Mostly, we use `>>>`

libvcx/src/utils/runtime.rs
```diff
@@ -0,0 +1,79 @@
...
+
+            unsafe { TP_HANDLE = 1; }
+        });
+    }
+}
+
+pub fn execute<F>(closure: F)
+    where
+        F: FnOnce() -> Result<(), ()> + Send + 'static {
+    trace!("Closure is going to be executed.");
```
mirgee, line 57: This is a useless comment we can probably delete

libvcx/src/utils/runtime.rs
```diff
@@ -0,0 +1,79 @@
...
+    if TP_INIT.is_completed() {
+        execute_on_tokio(future::lazy(|_| closure()));
+    } else {
+        thread::spawn(closure);
+    }
```
mirgee, line 62: Don't we want to fail when threadpool was not inicialized? If not, we should cover both branches with tests.

mirgee (changes requested):

libvcx/src/utils/runtime.rs
Patrik-Stas: added test `test_call_c_callable_api_without_threadpool` which calls c-callable API which ten doesn't run task on threadpool, but on a new thread instead

PR #204, opened 2020-11-15
Update to Rust edition2018, fix compile warnings

Patrik-Stas:


mirgee (approved):
I like this 👍 , except for the conflicts I will have to go through - don't think it will be that bad, though

Issue #210, opened 2020-11-19
Next phase of initialization API


Patrik-Stas:
The way `aries-vcx` is currently supposed to be initialized typically looks as follows:
```
        logger.info('Initializing VCX settings')
        await initVcxCore(JSON.stringify(vcxAgentProvision))
        logger.info('Opening pool')
        await openVcxPool()
        logger.info('Opening wallet')
        await openVcxWallet()
        logger.info('LibVCX fully initialized')
```
1. Everthing starts with `initVcxCore` which does 2 things:
- initializes "settings" - global state of the library
- initializes logger
- initializes threadpool for async FFI calls

2. Then you have to open pool and wallet, whereas both of these functions relies on global state configured via `vcxInitCore`.

We previously had all these operations as part of `vcx_init_with_config` and we've split them into separate APIs. This enabled for more flexibility, but I see the current state as only a phase. So I propose next phase, where we deprecate `vcx_init_core`, `vcx_open_wallet`, `vcx_open_pool` and instead introduce:

`vcx_init_ffi` - creates runtime for executing work on async FFI APIs
`vcx_init_logger` - initializes logger
`vcx_init_issuer_configuration` - initializes settings.rs with information issuer needs to know to operate (see `vcx_configure_issuer_wallet ` at https://github.com/hyperledger/aries-vcx/issues/209)
`vcx_create_agency_client(agency_config)` - creates shared global instance of `agency_client` in `aries-vcx`. See detials of  (agency_config)[https://github.com/hyperledger/aries-vcx/issues/209]
`vcx_open_wallet_directly(wallet_config)` - opens wallet 
`vcx_open_pool_directly(pool_config)`- opens pool

Issue #209, opened 2020-11-19
Split vcx_provision_agent, add agency_client initialization API


Patrik-Stas:
# Split of `vcx_provision_agent`
I propose to split deprecate `vcx_provision_agent` and split it into 3 new functions `vcx_create_wallet`, `vcx_configure_issuer_wallet`, `vcx_provision_new_agent` due to following problems. To enable this, we will also need `vcx_open_wallet` which does not rely on `settings.rs`.

## Problem 1
Currently this is function which has was given 3 responsibilities:
- create new wallet
- configure wallet (create prover master secret in wallet, create issuer's DID from seed (using `enterprise_seed` attribute))
- using that wallet, onboard a new agency agent
Though this covers common scenarios, it's restrictive at the same time - what if I only want to create new wallet and onboard agency client later? (perhaps running on mobile app for the first while not being connected to the internet - we might not be able to onboard agency client, but we might at least want to create the app wallet). Or perhaps we have wallet(s) and want to onboard to agency using it.

## Problem 2
Additionally there's one more design issue with `vcx_provision_agent`. Currently, it requires you to pass in "config" looking as such:
```json
{
    "path": "/path/to/genesis/127.0.0.1",

    "name": "my_agent_name",
    "logo": "https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcTjkD2iJdaKMlH2sKgR_flz6sa3mgkB9PRQwve4GLoXqPI7r73RiQ",

    "agency_url": "http://127.0.0.1:8080",
    "agency_did": "VsKV7grR1BUE29mG2Fm2kX",
    "agency_verkey": "Hezce2UWMZ3wUhVkh2LfKSs8nDzWwzs2Win7EzNN3YaR",

    "payment_method": "null",
    
    "enterprise_seed": "000000000000000000000000Trustee1",

    "wallet_name": "my_wallet_name",
    "wallet_key": "123456",
    "wallet_key_derivation": "ARGON2I_MOD",
    "wallet_type": "postgres_storage",
    "storage_config": "{\"url\":\"localhost:5432\"}",
    "storage_credentials": "{\"account\":\"postgres\",\"password\":\"password_123\",\"admin_account\":\"postgres\",\"admin_password\":\"password_foo\"}"
}

```
Performs some sort of mappings to these fields, so that it produces similarly looking config, with additional agency_client related info (`remote_to_sdk_did`, .. etc.): 
```json
{
[...]

PR #219, opened 2020-11-30
Split provision into API-exposed stages

mirgee:
Adds new methods

* vcx_create_wallet
* vcx_configure_issuer_wallet
* vcx_provision_new_agent
* vcx_open_wallet_directly
* vcx_close_wallet_directly

as per issue [#209](https://github.com/hyperledger/aries-vcx/issues/209) in preparation for #210. 


libvcx/src/api/utils.rs
```diff
@@ -31,6 +31,7 @@ pub struct UpdateAgentInfo {
 /// #Returns
 /// Configuration (wallet also populated), on error returns NULL
 #[no_mangle]
+#[deprecated(since = "0.14.0", note = "")]
```
mirgee, line 34: Will add a reference to the new methods

libvcx/src/api/wallet.rs
```diff
@@ -13,6 +13,178 @@ use crate::utils::cstring::CStringUtils;
...
+
+/// Closes wallet chosen using provided wallet handle.
+///
+/// #Params
+/// command_handle: command handle to map callback to user context.
+///
+/// wallet_handle: wallet handle
+///
+/// #Returns
+/// Error code as a u32 and wallet handle as i32
```
mirgee, line 159: Will fix docstring

libvcx/src/api/wallet.rs
```diff
@@ -13,6 +13,178 @@ use crate::utils::cstring::CStringUtils;
...
+/// cb: Callback that provides configuration or error status
+///
+/// # Example wallet config ->
+/// {
+///   "wallet_name": "my_wallet_name",
+///   "wallet_key": "123456",
+///   "wallet_key_derivation": "ARGON2I_MOD",
+///   "wallet_type": "postgres_storage",
+///   "storage_config": "{\"url\":\"localhost:5432\"}",
+///   "storage_credentials": "{\"account\":\"postgres\",\"password\":\"password_123\",\"admin_account\":\"postgres\",\"admin_password\":\"password_foo\"}"
```
mirgee, line 32: Missing return value

libvcx/src/api/wallet.rs
```diff
@@ -13,6 +13,178 @@ use crate::utils::cstring::CStringUtils;
...
+
+/// Opens wallet chosen using provided config and returns its wallet handle.
+///
+/// #Params
+/// command_handle: command handle to map callback to user context.
+///
+/// wallet_config: wallet configuration
+///
+/// #Returns
+/// Error code as a u32 and wallet handle as i32
```
mirgee, line 122: Will fix docstring

libvcx/src/libindy/utils/wallet.rs
```diff
@@ -49,6 +60,43 @@ pub fn reset_wallet_handle() {
...
+
+    let wh = create_and_open_as_main_wallet(
+        &config.wallet_name,
+        &config.wallet_key,
+        &config.wallet_key_derivation,
+        config.wallet_type.as_ref().map(String::as_str),
+        config.storage_config.as_ref().map(String::as_str),
+        config.storage_credentials.as_ref().map(String::as_str),
+    )?;
+    set_wallet_handle(wh);
```
Patrik-Stas, line 75: Isn't wallet handle already set in `create_and_open_as_main_wallet `?

libvcx/src/libindy/utils/wallet.rs
```diff
@@ -49,6 +60,43 @@ pub fn reset_wallet_handle() {
...
+    trace!("initialized wallet");
```
Patrik-Stas, line 76: Maybe: `Wallet {} was created`?

libvcx/src/libindy/utils/wallet.rs
```diff
@@ -49,6 +60,43 @@ pub fn reset_wallet_handle() {
...
+
+    // If MS is already in wallet then just continue
+    anoncreds::libindy_prover_create_master_secret(settings::DEFAULT_LINK_SECRET_ALIAS).ok();
+    
+    close_main_wallet()?;
+    Ok(())
+}
+
+pub fn configure_issuer_wallet(enterprise_seed: &str, institution_name: &str) -> VcxResult<String> {
```
Patrik-Stas, line 85: I think this should only take the seed, as this function is now doing 2 things:
- setting up the wallet
- initializing memory

And actually, `institution name` is used by verifier, but mobile apps might commonly be (going bit into future) as well in position of verifier.`

Patrik-Stas (changes requested):

libvcx/src/api/wallet.rs
```diff
@@ -13,6 +13,178 @@ use crate::utils::cstring::CStringUtils;
 use crate::utils::error;
 use crate::utils::threadpool::spawn;
 
+/// Creates new wallet and master secret using provided config. Keeps wallet closed.
+///
+/// #Params
+/// command_handle: command handle to map callback to user context.
+///
+/// wallet_config: wallet configuration
```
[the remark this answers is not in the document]
mirgee: It is there already

libvcx/src/api/wallet.rs
```diff
@@ -13,6 +13,183 @@ use crate::utils::cstring::CStringUtils;
...
+///
+/// cb: Callback that provides wallet handle or error status
+///
+/// #Returns
+/// Error code as a u32
+#[no_mangle]
+pub extern fn vcx_open_wallet_directly(command_handle: CommandHandle,
```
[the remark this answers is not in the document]
mirgee: The `*_directly` suffix is there to distnguish from the current `vcx_open_wallet`, changing signature of which would break existing clients using it.

libvcx/src/libindy/utils/wallet.rs
mirgee: Good catch

libvcx/src/libindy/utils/wallet.rs
mirgee: I don't think the first one is an argument - it will set up the wallet and initialize settings anyway. As for the second one, it is true that this will probably be the case in the future, but making the initialization even more complex introducing another method just to set up the config for a future use case seems unnecessary for now. I would agree to to rename to `configure_institution_wallet`, however.

libvcx/src/api/utils.rs
```diff
@@ -101,6 +105,66 @@ pub extern fn vcx_agent_provision_async(command_handle: CommandHandle,
...
+///  "remote_to_sdk_did": "GkdUhwyWqNw3vGs6FQFFHb",
+///  "remote_to_sdk_verkey": "9axcTwXeJ1haJBw9LqexT8dRpiFCJwA6ZUevM5nfiDKg",
+///  "sdk_to_remote_did": "C5DiHD1n3MqNcv5h7PBK9J",
+///  "sdk_to_remote_verkey": "732pD7kDiBjSyS57aNXi52Xpg2DLCTb43aLpddo2X8CG"
+/// }
+///
+/// #Returns
+/// Error code as a u32
+#[no_mangle]
+pub extern fn vcx_provision_new_agent(command_handle: CommandHandle,
```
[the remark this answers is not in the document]
mirgee: Agreed, that's less ambiguous.

libvcx/src/api/wallet.rs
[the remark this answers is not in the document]
Patrik-Stas: yeah, since `vcx_open_wallet` is taken, I am suggesting `vcx_open_main_wallet`, I think you've missed the "_main_" in my suggestion

libvcx/src/libindy/utils/wallet.rs
Patrik-Stas: `institution_name ` could be just part of config being passed to `vcx_init_core`. if `configure_issuer_wallet` would be the only way to configure `institution_name` in library memory settings, that would be an agent would have to actually on every startup  call `configure_issuer_wallet `. But `configure_issuer_wallet ` should be only 1 time operation per wallet.

Patrik-Stas (approved):
lgtm

PR #222, opened 2020-12-03
Add prerelease support

mirgee:
Publish a prerelease on CI label or in a PR on a non-master branch on version change.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,20 +71,41 @@ jobs:
           echo "Highest released version was: $REPO_VERSION_MOST_RECENT, desired version (specified in Cargo.toml) is $REPO_VERSION_DESIRED"
 
           RELEASE="false"
+          PRERELEASE="false"
           if [[ "$IS_FORK" == "false" ]]
           then
-            if [[ "$BRANCH_NAME" == "master" && "$REPO_VERSION_DESIRED" != "$REPO_VERSION_MOST_RECENT" ]]
+            if [[ "$REPO_VERSION_DESIRED" != "$REPO_VERSION_MOST_RECENT" ]]
             then
-              PUBLISH_VERSION="$REPO_VERSION_DESIRED"
-              RELEASE="true"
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -71,22 +71,43 @@ jobs:
...
+                echo "This is push to master, and version was bumped from $REPO_VERSION_MOST_RECENT to $REPO_VERSION_DESIRED. Will publish a release of $REPO_VERSION_DESIRED."
+              else
+                PUBLISH_VERSION="$REPO_VERSION_DESIRED-${{ github.run_number }}"
+                PRERELEASE="true" 
+                echo "This is push to a non-master branch, and version was bumped from $REPO_VERSION_MOST_RECENT to $REPO_VERSION_DESIRED. Will publish a pre-release of $REPO_VERSION_DESIRED."
+              fi;
             else
               PUBLISH_VERSION="$REPO_VERSION_MOST_RECENT-$BRANCH_NAME-${{ github.run_number }}"
-              echo "This is not push to master. Will be publishing version: $PUBLISH_VERSION."
+              if [[ ${{ contains(github.event.pull_request.labels.*.name, 'ci') }} ]]
```
Patrik-Stas, line 91: So `ci` means it's pre-release? Perhaps we should use `pre-release` label instead?

Patrik-Stas, line 91: Will this 
```
              if [[ ${{ contains(github.event.pull_request.labels.*.name, 'ci') }} ]]
```
also work if I am running CI on a branch, not PR? I think if it's branch, field `pull_request` would not exist, so it might be broken in that case. 
Sometimes I run pipelines on branches

Patrik-Stas: I think the solution might be to just not allow making pre-releases from branches.

Patrik-Stas (commented):
Left a few comments about making pre-releases on PR branches

.github/workflows/main.yml
mirgee: No problem, but I thought we planned to do a pre-release on CI changes to test that the release process still works.

Patrik-Stas (approved):
Left a few comments about making pre-releases on PR branches

PR #223, opened 2020-12-03
Add new initialization methods

mirgee:
Added methods:
* vcx_init_threadpool - Initializes tokio threadpool
* vcx_create_wallet - creates a new wallet and master secret using provided config (wallet name, key, key der. meth.), but keeps wallet closed
* vcx_create_agency_client_for_main_wallet - creates a new agency client from config (agency and agent did, verkey, url, etc.)
* vcx_open_main_pool - opens the pool
* vcx_open_main_wallet - opens the created wallet
* vcx_close_main_wallet - closes the created wallet
* vcx_configure_issuer_wallet - creates and stores did and keypair for credential issuance
* vcx_init_issuer_config - stores provided config used by issuer in the memory (inst. did, name and verkey)

Implements #210 .

Deprecated methods:
* vcx_open_pool
* vcx_agent_update_info
* vcx_init_core
* vcx_open_wallet


libvcx/src/api/vcx.rs
```diff
@@ -15,17 +15,117 @@ use crate::utils::error;
 use crate::utils::threadpool::spawn;
 use crate::utils::version_constants;
 
-/// Initializes VCX with config settings
+#[no_mangle]
+pub extern fn vcx_init_threadpool() -> u32 {
```
Patrik-Stas, line 19: Let's add json config parameter and let enable use to specify number of threads

Patrik-Stas: Perhaps later we can add other options to the config file, if we'd want to give user more options over threadpool configuration

libvcx/src/api/vcx.rs
```diff
@@ -15,17 +15,117 @@ use crate::utils::error;
...
+///    my_vk - verkey of this client used to communicate with it's agent in the agency
+/// }
 ///
-/// cb: Callback that provides error status of initialization
+/// cb: Callback that provides error status
+///
+/// #Returns
+/// Error code as a u32
+#[no_mangle]
+pub extern fn vcx_create_agency_client(command_handle: CommandHandle, config: *const c_char, cb: Option<extern fn(xcommand_handle: CommandHandle, err: u32)>) -> u32 {
```
Patrik-Stas, line 50: Let's "save" this good name `vcx_create_agency_client` for later :-) Let's use  `vcx_create_agency_client_for_main_wallet` - I know it's ugly, but the behaviour of this function is still linked to "singleton" nature of the library

libvcx/src/api/vcx.rs
```diff
@@ -15,17 +15,117 @@ use crate::utils::error;
...
+    error::SUCCESS.code_num
+}
+
+
+/// Stores institution did, verkey and name in memory.
+///
+/// #Params
+/// command_handle: command handle to map callback to user context.
+///
+/// issuer_config: Pool configuration
```
Patrik-Stas, line 81: "Pool configuration" - probably forgot to change afte copypaste?

libvcx/src/api/vcx.rs
```diff
@@ -15,17 +15,117 @@ use crate::utils::error;
...
+/// {
+///     "institution_name" - Name of the issueing / verifying institution
+///     "institution_did" (optional) - Institution did obtained on vcx_configure_issuer_wallet
```
Patrik-Stas, line 84: According to implementation this doesn't seem to be optional. If it's not provided, `CONFIG_INSTITUTION_DID` /   `CONFIG_INSTITUTION_VERKEY`  is not set, which would cause runtime error when trying to perform operation relying on these configurations (such as `vcx_issuer_create_credential`)

libvcx/src/api/vcx.rs
```diff
@@ -15,17 +15,117 @@ use crate::utils::error;
...
+
+/// Initializes VCX with config settings
+///
+/// #Params
+///
+/// config: The agent provision configuration. You can produce this by provisioning agent using function vcx_provision_agent
 ///
 /// #Returns
 /// Error code as a u32
+#[deprecated(since = "0.14.0", note = "Use vcx_init_threadpool, vcx_init_logger, vcx_init_issuer_config instead.")]
```
Patrik-Stas, line 128: Damn nice! Happy to abolish `vcx_init_core`!

Patrik-Stas (changes requested):
Could you also add this to NodeJS wrapper and write a test using new APIs for initialization? Perhaps migrate alice/faber demo?

libvcx/src/api/vcx.rs
mirgee: Yes, but it is optional according to implementation of the documented function, and the point of making it optional is that if `vcx_configure_issuer_wallet` was called in the same runtime session before `vcx_init_issuer_config`, `institution_did` and `institution_verkey` does not have to be saved in memory, as this was done in `vcx_configure_issuer_wallet`.

mirgee: I am not necessarily arguing strongly for making them optional, just explaining that the point of making them optional is due to this use case.

agents/node/vcxagent-core/src/agent.js
```diff
@@ -32,12 +34,16 @@ async function createVcxAgent ({ agentName, genesisPath, agencyUrl, seed, usePos
   async function agentInitVcx () {
     logger.info(`Initializing ${agentName} vcx session.`)
     logger.silly(`Using following agent provision to initialize VCX settings ${JSON.stringify(agentProvision, null, 2)}`)
-    await initVcxCore(JSON.stringify(agentProvision))
+    logger.silly('Initializing threadpool')
+    await initThreadpool({})
```
Patrik-Stas, line 38: This looks greaat now!

agents/node/vcxagent-core/src/utils/vcx-workflows.js
```diff
@@ -49,32 +54,43 @@ async function provisionAgentInAgency (agentName, genesisPath, agencyUrl, seed,
...
+  await createWallet(walletConfig)
+  logger.debug(`Opening wallet with config: ${JSON.stringify(walletConfig, null, 2)}`)
+  const wh = await openMainWallet(walletConfig)
+  logger.debug(`Configuring issuer's wallet with seed: ${seed}`)
+  const issuerConfig = JSON.parse(await configureIssuerWallet(seed))
+  issuerConfig.institution_name = agentName
+  logger.debug(`Provisioning agent with config: ${JSON.stringify(agencyConfig, null, 2)}`)
+  agencyConfig = JSON.parse(await provisionCloudAgent(agencyConfig))
+  logger.debug(`Provisined agent with config: ${JSON.stringify(agencyConfig, null, 2)}`)
+  await closeMainWallet(wh) // TODO: Get rid of wh
```
Patrik-Stas, line 91: Yup let's do that before :-)

agents/node/vcxagent-core/src/utils/vcx-workflows.js
```diff
@@ -49,32 +54,43 @@ async function provisionAgentInAgency (agentName, genesisPath, agencyUrl, seed,
...
-    agency_url: agencyUrl,
-    agency_did: 'VsKV7grR1BUE29mG2Fm2kX',
-    agency_verkey: 'Hezce2UWMZ3wUhVkh2LfKSs8nDzWwzs2Win7EzNN3YaR',
+
+  const walletConfig = {
     wallet_name: agentName,
     wallet_key: '123',
-    payment_method: 'null',
-    enterprise_seed: seed
+    wallet_key_derivation: 'ARGON2I_INT'
```
Patrik-Stas, line 61: Let's use `RAW` with key `8dvfYSt5d1taSd6yJdpjq4emkwsPDDLYxkNFysFD2cZY` (these can be generated with function provided by libindy). That should make tests a little bit faster.

agents/node/vcxagent-core/src/agent.js
```diff
@@ -32,12 +34,16 @@ async function createVcxAgent ({ agentName, genesisPath, agencyUrl, seed, usePos
...
+    logger.silly('Initializing issuer config')
+    await initIssuerConfig(agentProvision.issuerConfig)
```
Patrik-Stas, line 40: Can you write a small migration function for NodeJS? Should take old agent provision config and turn it into new one. Should be idempotent, so passing in config in new format should do no change.

libvcx/src/api/vcx.rs
```diff
@@ -37,6 +155,62 @@ pub extern fn vcx_init_core(config: *const c_char) -> u32 {
...
+///         By default Libindy sends a read requests to 2 nodes in the pool.
+///         If response isn't received or `state proof` is invalid Libindy sends the request again but to 2 (`number_read_nodes`) * 2 = 4 nodes and so far until completion.
+/// }
+///
+/// cb: Callback that provides error status
+///
+/// #Returns
+/// Error code as a u32
+#[no_mangle]
+pub extern fn vcx_open_pool_directly(command_handle: CommandHandle, pool_config: *const c_char,  cb: extern fn(xcommand_handle: CommandHandle, err: u32)) -> u32 {
```
Patrik-Stas, line 190: Maybe `vcx_open_main_pool` for symmetry? It's the same case, right now we have 1 main pool opened, but perhaps in the future 1 agent can be connected to multiple ledgers.

Patrik-Stas:
Please list also deprecated functions in PR description

libvcx/src/api/vcx.rs
```diff
@@ -89,6 +262,7 @@ pub extern fn vcx_open_pool(command_handle: CommandHandle, cb: extern fn(xcomman
 ///
 /// #Returns
 /// Error code as a u32
+#[deprecated(since = "0.14.0", note = "Use vcx_open_wallet_directly instead.")]
```
Patrik-Stas, line 265: Use vcx_open_main_wallet instead

agents/node/vcxagent-core/src/agent.js
```diff
@@ -32,12 +34,16 @@ async function createVcxAgent ({ agentName, genesisPath, agencyUrl, seed, usePos
   async function agentInitVcx () {
```
Patrik-Stas, line 34: Since. this is now used by all integrations tests (I think all of them rely on vcxagent-core), could you also add a test which would just verify that the old workflow also still works? (initVcxCore+openVcxPool+openVcxWallet)

Then in some next PR, let's also add java+ios wrappers for the functions added in this PR.

wrappers/node/src/rustlib.ts
```diff
@@ -58,6 +58,13 @@ export type rust_listener_handle = rust_object_handle;
 export type rust_connection_handle = rust_object_handle;
 
 export interface IFFIEntryPoint {
+  vcx_open_main_pool: (commandId: number, config: string, cb: any) => number,
+
+  vcx_create_agency_client_for_main_wallet: (commandId: number, config: string, cb: any) => number,
+  vcx_provision_cloud_agent: (commandId: number, config: string, cb: any) => number,
```
Patrik-Stas, line 64: Seems like there no C function like this defined in rust?

Patrik-Stas (changes requested):
I left few comments and updated list of added functions in PR description. Could you please also add there short 1-liner description for each new function

wrappers/node/src/rustlib.ts
mirgee: It is, see libvcx/src/api/utils.rs:139

Patrik-Stas (dismissed):

Patrik-Stas (approved):

PR #228, opened 2020-12-08
Init rust api once

OoDeLally:
This PR prevents to call `initRustAPI()` several time and needlessly load the Rust library several times.
User functions are no longer responsible for loading it. Instead the top caller should load it once for everyone.


wrappers/node/test/helpers/utils.ts
```diff
@@ -55,3 +59,17 @@ export const shouldThrow = (fn: () => any): Promise<vcx.VCXInternalError> => new
 })
 
 export const sleep = (timeout: number) => new Promise((resolve, reject) => setTimeout(resolve, timeout))
+
+
+let garbageCollectionBeforeExitIsScheduled = false
+
+// For some reason, The Rust library segfaults if global.gc() is not called explicitly.
```
Patrik-Stas, line 66: Perhaps just note that this only happens on exit and the trick only works for Mocha

Patrik-Stas (changes requested):
Please also strip `initRustAPI(options.libVCXPath)` from `initVcxCore`

wrappers/node/test/helpers/utils.ts
OoDeLally: Changed to a better comment.

Patrik-Stas (dismissed):

Patrik-Stas (approved):

PR #237, opened 2020-12-14
Node wrapper eslint fix anys

OoDeLally:
These changes require a bit of attention, since they are not merely cosmetics

### DIFF
--- a/wrappers/node/src/api/wallet.ts
+++ b/wrappers/node/src/api/wallet.ts
@@ -6,6 +6,7 @@ import { rustAPI } from '../rustlib';
 import { createFFICallbackPromise } from '../utils/ffi-helpers';
 import { IUTXO } from './common';
 import { voidPtrToUint8Array } from './connection';
+import {PtrBuffer} from "./utils";
 
 export type PaymentAddress = string;
 export type PaymentAmount = number;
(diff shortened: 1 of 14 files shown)


wrappers/node/src/api/wallet.ts
```diff
@@ -215,16 +215,16 @@ export class Wallet {
           Callback(
             'void',
             ['uint32', 'uint32', 'pointer', 'uint32'],
-            (xHandle: number, err: number, details: any, length: number) => {
+            (xHandle: number, err: number, detailsPtr: ref.Type, length: number) => {
```
Patrik-Stas, line 218: Ref type is complex structure according to its definition
```
export interface Type {
    /** The size in bytes required to hold this datatype. */
    size: number;
    /** The current level of indirection of the buffer. */
    indirection: number;
    /** To invoke when `ref.get` is invoked on a buffer of this type. */
    get(buffer: Buffer, offset: number): any;
    /** To invoke when `ref.set` is invoked on a buffer of this type. */
    set(buffer: Buffer, offset: number, value: any): void;
    /** The name to use during debugging for this datatype. */
    name?: string;
    /** The alignment of this datatype when placed inside a struct. */
    alignment?: number;
}
```
whereas rust function looks like this
```
pub extern fn vcx_wallet_sign_with_address(command_handle: CommandHandle,
                                           payment_address: *const c_char,
                                           message_raw: *const u8,
                                           message_len: u32,
                                           cb: Option<extern fn(xcommand_handle: CommandHandle, err: u32,
                                                                signature: *const u8, signature_len: u32)>) -> u32 {

```
where `signature: *const u8` in callback maps to `detailsPtr ` and it's just u8 pointer.


We don't use `signWithAddress` now, probably not even testing it so I don't know if your change actually works or not, but doesn't seem correct to me.

Patrik-Stas (changes requested):

Patrik-Stas (approved):

PR #248, opened 2020-12-24
Use RwLock instead of Mutex in ObjectCache

Patrik-Stas:
- `ObjectCache` struct/implementation is used for library-wide global storage of vcx objects when using `aries-vcx` via FFI. The objects are identified by numeric "handle". These handles are keys to `map`, which is protected by `mutex` as there might be simultaneous calls into `aries-vcx` using this shared global data. Also each individual object inside the map is protected by `mutex`.
- Mutex assures there's only 1 write OR read at any point of time. Having `mutex` wrapping the map hence prevents for example creating new connection, if currently existing connection is being read to access some of its data. This can probably drive performance down rapidly in concurrent applications.

### DIFF
--- a/libvcx/src/utils/object_cache/mod.rs
+++ b/libvcx/src/utils/object_cache/mod.rs
@@ -1,8 +1,7 @@
 use std::collections::HashMap;
 use std::ops::Deref;
 use std::ops::DerefMut;
-use std::sync::Mutex;
-use std::sync::MutexGuard;
+use std::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 use rand::Rng;
 
@@ -10,7 +9,7 @@ use crate::error::prelude::*;
 
 pub struct ObjectCache<T> {
     pub cache_name: String,
-    pub store: Mutex<HashMap<u32, Mutex<T>>>,
+    pub store: RwLock<HashMap<u32, RwLock<T>>>,
 }
 
 impl<T> ObjectCache<T> {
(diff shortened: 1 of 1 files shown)


mirgee (commented):
I suppose the reason for this change is safety and performance (to not block multiple readers) - however, it is not self-evident that `RwLock` is would be faster (although probable, since we expect more reads, I think)? It seems that it would depend on the target platform's implementation of `RwLock` algorithm...

Patrik-Stas:
@mirgee good to think about unintended consequences, but I think even if RwLock is significantly slower than Mutex, the fact that read access to VCX Object of type X would block read accesses to all other objects of type X outweights any potential slowness of RwLocks.
I you look at implementation of any `update` method, for example disclosed proof:
```
HANDLE_MAP.get_mut(handle, |proof| {
   .. do stuff ..
  .. send something over network, synchronously ...
 }
```
This operation can take long time because it contains synchronous network and filesystem operations - and during this time, with Mutex any access to read information about other disclosed proofs are blocked

mirgee:
> @mirgee good to think about unintended consequences, but I think even if RwLock is significantly slower than Mutex, the fact that read access to VCX Object of type X would block read accesses to all other objects of type X outweights any potential slowness of RwLocks.
> If you look at implementation of any `update` method, for example disclosed proof:
> 
> ```
> HANDLE_MAP.get_mut(handle, |proof| {
>    .. do stuff ..
>   .. send something over network, synchronously ...
>  }
> ```
> 
> This operation can take long time because it contains synchronous network and filesystem operations - and during this time, with Mutex any access to read information about other disclosed proofs are blocked

Agree

mirgee (dismissed):

mirgee:
> @mirgee good to think about unintended consequences, but I think even if RwLock is significantly slower than Mutex, the fact that read access to VCX Object of type X would block read accesses to all other objects of type X outweights any potential slowness of RwLocks.
> If you look at implementation of any `update` method, for example disclosed proof:
> 
> ```
> HANDLE_MAP.get_mut(handle, |proof| {
>    .. do stuff ..
>   .. send something over network, synchronously ...
>  }
> ```
> 
> This operation can take long time because it contains synchronous network and filesystem operations - and during this time, with Mutex any access to read information about other disclosed proofs are blocked

In this case, so would RwLock, but you are right,
```
HANDLE_MAP.get(handle, |proof| {
   .. do stuff ..
 }
```
blocks any other concurrent read access.

PR #249, opened 2020-12-25
Remove ~service decorator from PresentationRequest message

Patrik-Stas:
This removes `~service` field from Proof Presentation message - it's not part of Aries proof presentation 1.0 rfc https://github.com/hyperledger/aries-rfcs/tree/master/features/0037-present-proof

it was also not being set by verifier role within our aries-vcx.


mirgee (approved):
It might make sense to include service endpoint in presentation request in case it was not exchanged, e.g. as part of [out-of-band protocol](https://github.com/hyperledger/aries-rfcs/tree/master/features/0434-outofband); but I agree it is superfluous until that is implemented.

PR #272, opened 2021-05-03
Verifier SM sends Aries messages through a closure

mirgee:

### DIFF
--- a/libvcx/src/aries/handlers/proof_presentation/verifier/state_machine.rs
+++ b/libvcx/src/aries/handlers/proof_presentation/verifier/state_machine.rs
@@ -86,9 +86,14 @@ impl VerifierSM {
         let state = match state {
             VerifierState::Initiated(state) => {
                 match message {
-                    VerifierMessages::SendPresentationRequest => {
-                        let presentation_request = PresentationRequest::create().set_request_presentations_attach(&state.presentation_request_data)?;
-                        send_message.ok_or(VcxError::from_msg(VcxErrorKind::InvalidState, "Attempted to call undefined send_message callback"))?(&presentation_request.to_a2a_message())?;
+                    VerifierMessages::SendPresentationRequest(comment) => {
+                        let presentation_request =
+                            PresentationRequest::create()
+                                .set_comment(comment)
+                                .set_request_presentations_attach(&state.presentation_request_data)?;
+                        send_message.ok_or(
+                            VcxError::from_msg(VcxErrorKind::InvalidState, "Attempted to call undefined send_message callback")
+                        )?(&presentation_request.to_a2a_message())?;
                         VerifierState::PresentationRequestSent((state, presentation_request).into())
                     }
                     _ => {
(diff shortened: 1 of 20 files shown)


libvcx/src/aries/handlers/proof_presentation/verifier/state_machine.rs
```diff
@@ -87,22 +87,8 @@ impl VerifierSM {
             VerifierState::Initiated(state) => {
                 match message {
                     VerifierMessages::SendPresentationRequest => {
-                        let my_did = get_pw_did(connection_handle)?;
-                        let remote_did = get_their_pw_verkey(connection_handle)?;
-
-                        let presentation_request: PresentationRequestData =
-                            state.presentation_request_data.clone()
-                                .set_format_version_for_did(&my_did, &remote_did)?;
```
Patrik-Stas, line 95: So I've checked this and I see that this method, based on format of DIDs sets up "format" of the proof request - either V1 or V2. If we remove this, what `format` will be used? Do you know what's difference between V1/V2?

libvcx/src/aries/handlers/proof_presentation/verifier/state_machine.rs
```diff
@@ -87,22 +87,8 @@ impl VerifierSM {
...
-
-                        let title = format!("{} wants you to share {}",
-                                            settings::get_config_value(settings::CONFIG_INSTITUTION_NAME)?, presentation_request.name);
-
-                        let presentation_request =
-                            PresentationRequest::create()
-                                .set_comment(title)
```
Patrik-Stas, line 102: We could keep this feature if parametrize `SendPresentationRequest` of `VerifierMessages` enum. Something like:
```
pub enum VerifierMessages {
    SendPresentationRequest(comment),
    ...
    }
    ```
 And so we could pass the comment variable from upper layers instead of having the logic hardcoded here (as it was until now)

File: libvcx/src/aries/handlers/proof_presentation/verifier/state_machine.rs [hunk 3]
```diff
@@ -122,8 +108,7 @@ impl VerifierSM {
                                     ProblemReport::create()
                                         .set_comment(err.to_string())
                                         .set_thread_id(&state.presentation_request.id.0);
-
-                                connection::send_message(connection_handle, problem_report.to_a2a_message())?;
+                                send_message.ok_or(VcxError::from_msg(VcxErrorKind::InvalidState, "Attempted to call undefined send_message callback"))?(&problem_report.to_a2a_message())?;
```
Patrik-Stas, line 111: Comes down to personal preference but I find this hard to read. I'd prefer
```
let send_message = send_message.ok_or(VcxError::from_msg(VcxErrorKind::InvalidState, "Attempted to call undefined send_message callback"))?;
send_message(&problem_report.to_a2a_message())?;
```
I'll leave it up to you, anyway as I mentioned in private discussion, I think eventually the side effect of sending response should IMO be pulled out from Aries FSMs.

libvcx/src/aries/handlers/proof_presentation/verifier/state_machine.rs
mirgee: It's format version for DID, not for the proof request and as far as I am aware, it is internal to aries-vcx and never read (only set by verifier), so `set_format_version_for_did` can be safely deleted as well. But perhaps it refers to some DID-spec / aries standard.

libvcx/src/aries/handlers/proof_presentation/verifier/state_machine.rs
mirgee: OK, it can be actually quite useful when not hardcoded

Patrik-Stas (changes requested):
generally LGTM, let's just few suggested changes

libvcx/src/aries/handlers/proof_presentation/verifier/state_machine.rs
mirgee: You are right in that the line is way too long. Even 
```
let send_message = send_message.ok_or(VcxError::from_msg(VcxErrorKind::InvalidState, "Attempted to call undefined send_message callback"))?;
```
seems too long. How about
```
send_message.ok_or(
    VcxError::from_msg(VcxErrorKind::InvalidState, "Attempted to call undefined send_message callback")
)?(&problem_report.to_a2a_message())?;
```

Patrik-Stas (approved):

PR #279, opened 2021-05-24
Refactor connection state machine to expose transition methods

Patrik-Stas:
- Previously in order to invoke transition/action on message, the caller of Connection FSM API had to call step function with `DidExchangeMessages` variant, representing the desired FSM transition. 

- This PR modifies the API such that each transition is represented by a method defined on Connection FSM.

- This has benefits as more complex arguments (for example trait objects, impl) can be passed down into transition, as opposed what could been previously injected into structure arguments within the `DidExchangeMessages` variants.


libvcx/src/aries/handlers/connection/connection.rs
```diff
@@ -262,7 +269,15 @@ impl Connection {
...
         trace!("Connection::connect >>> source_id: {}", self.source_id());
-        self.step(DidExchangeMessages::Connect())
+        self.connection_sm = match &self.connection_sm {
+            SmConnection::Inviter(sm_inviter) => {
+                SmConnection::Inviter(sm_inviter.clone().transition_connect()?)
+            }
+            SmConnection::Invitee(sm_invitee) => {
+                SmConnection::Invitee(sm_invitee.clone().transition_connect()?)
+            }
+        };
```
mirgee, line 279: Maybe it would be nice if inviter / invitee SMs implemented a trait and `Connection` would reference an implementor of the trait, so that we may avoid these branches

libvcx/src/aries/handlers/connection/connection.rs
```diff
@@ -378,29 +386,115 @@ Get messages received from connection counterparty.
...
+            }
+            A2AMessage::ConnectionProblemReport(problem_report) => {
+                sm_inviter.transition_receive_problem_report(problem_report)
+            }
+            A2AMessage::PingResponse(ping_response) => {
+                sm_inviter.transition_ping_response_received(ping_response)
+            }
+            // A2AMessage::Disclose((query_, comment)) => {
+            //     sm_inviter.transition_discover_features_received(query_, comment) // todo
+            // }
```
mirgee, line 425: What is the problem with this message?

libvcx/src/aries/handlers/connection/connection.rs
```diff
@@ -378,29 +386,115 @@ Get messages received from connection counterparty.
...
+            }
+            A2AMessage::ConnectionProblemReport(problem_report) => {
+                sm_invitee.transition_receive_problem_report(problem_report)
+            }
+            A2AMessage::PingResponse(ping_response) => {
+                sm_invitee.transition_ping_response_received(ping_response)
+            }
+            // DidExchangeMessages::DiscoverFeatures((query_, comment)) => { // todo
+            //     sm_invitee.transition_discover_features_received(query_, comment)
+            // }
```
mirgee, line 461: dtto

libvcx/src/aries/handlers/connection/invitee/state_machine.rs
```diff
@@ -193,74 +187,162 @@ impl SmConnectionInvitee {
...
-                    }
-                    DidExchangeMessages::ProblemReportReceived(problem_report) => {
-                        InviteeState::Null((state, problem_report).into())
-                    }
-                    _ => {
-                        InviteeState::Invited(state)
-                    }
-                }
+            _ => {
+                state.clone()
```
mirgee, line 222: According to the [connection RFC](https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol), receiving an invitation in requested state should start a new relationship, and probably return error in responded and complete. Of course, there are more differences like this (some addressed in my backchannel PR, some not covered by the harness tests ATM) we should probably leave to another PR, though.

mirgee (approved):

PR #280, opened 2021-05-25
Changes related to aries test harness backchannel

mirgee:
Previously, invitee's states were 
```
Null, Invited, Requested, Completed
```
whereas inviter's states were
```
Null, Invited, Responded, Completed
```
This in order to achieve compatibility with [RFC0160](https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol), this PR adds a new state for both inviter and invitee, making the state set
```
Null, Invited, Requested, Responded, Completed
```
for both.

Moreover, mapping of the states to `VcxStateType` was changed such that the `Null` state (the initial and "errored" state) maps the the zeroth state (`VcxStateNone`) instead of the first state (`VcxStateInitialized`) and the transition progress the mapped `VcxStateType` in unit increments. This is a breaking change.

Finally, to achieve backwards compatibility, an `autohop` flag was added to the state machine, which skips the `Responded` and `Requested` states for invitee and inviter respectively.

Related: https://github.com/hyperledger/aries-agent-test-harness/pull/243

----

Additionally this includes following refactoring changes
https://github.com/hyperledger/aries-vcx/pull/279


libvcx/src/aries/handlers/connection/inviter/state_machine.rs
```diff
@@ -48,6 +54,7 @@ impl SmConnectionInviter {
             source_id: source_id.to_string(),
             state: InviterState::Null(NullState {}),
             agent_info: AgentInfo::default(),
+            autohop: false
```
mirgee, line 57: This should not be hardcoded, obviously

libvcx/src/lib.rs
```diff
@@ -48,7 +48,7 @@ pub mod disclosed_proof;
 
 pub mod aries;
 mod filters;
-mod libindy;
+pub mod libindy;
```
mirgee, line 51: This is because backchannel needs access to wallet utils

libvcx/src/aries/handlers/connection/inviter/state_machine.rs
```diff
@@ -217,8 +232,87 @@ impl SmConnectionInviter {
...
+                                    .set_thread_id(&state.request.id.0);
+         
+                                state.request.connection.did_doc.send_message(&problem_report.to_a2a_message(), &agent_info.pw_vk).ok();
+                                (InviterState::Null((state, problem_report).into()), agent_info)
+                            }
+                        }
+                    }
+                    _ => (state.clone(), agent_info.clone())
+                };
+                Ok(SmConnectionInviter { source_id, agent_info, state, autohop })
```
mirgee, line 309: Thinking that we should use `Self` wherever possible

PR #294, opened 2021-06-14
Use enums instead of u32 to distinguish connection FSM states on Rust layer

Patrik-Stas:
Replaces `u32` representation of AriesFMS states by enums (which do not contain complete copy of state details).  u32 representation of FSM state exists only so it could be used on C ABI layer and therefore it should also be expressed in that layer.

### DIFF
--- a/libvcx/src/abi_utils/mod.rs
+++ b/libvcx/src/abi_utils/mod.rs
@@ -1,0 +1,28 @@
+use crate::aries::handlers::connection::connection::ConnectionState;
+use crate::aries::handlers::connection::inviter::state_machine::InviterState;
+use crate::aries::handlers::connection::invitee::state_machine::InviteeState;
+
+impl From<ConnectionState> for u32 {
+    fn from(state: ConnectionState) -> u32 {
+        match state {
+            ConnectionState::Inviter(inviter_state) => {
+                match inviter_state {
+                    InviterState::Null => 0,
+                    InviterState::Invited => 1,
+                    InviterState::Requested => 2,
+                    InviterState::Responded => 3,
+                    InviterState::Completed => 4,
+                }
+            }
+            ConnectionState::Invitee(invitee_state) => {
+                match invitee_state {
+                    InviteeState::Null => 0,
+                    InviteeState::Invited => 1,
+                    InviteeState::Requested => 2,
+                    InviteeState::Responded => 3,
+                    InviteeState::Completed => 4,
+                }
+            }
+        }
+    }
+}
--- a/libvcx/src/aries/handlers/connection/connection.rs
+++ b/libvcx/src/aries/handlers/connection/connection.rs
@@ -124,13 +124,13 @@ impl Connection {
         }.into()
     }
 
-    pub fn state(&self) -> ConnectionState {
+    pub fn get_state(&self) -> ConnectionState {
         match &self.connection_sm {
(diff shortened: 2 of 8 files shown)


libvcx/src/aries/handlers/connection/connection.rs
```diff
@@ -41,6 +41,13 @@ struct ConnectionInfo {
     their: Option<SideConnectionInfo>,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum ConnectionState {
```
Patrik-Stas, line 45: We'll be able to remove this enum once we split general `Connection` into `Inviter` and `Invitee` VCX objects, analogically to `issuer` and `holder`

Patrik-Stas:
I am sitting on fence whether it wouldn't be better just to return the existing structures (what this PR renamed to `InviteeFullState` / `InviterFullState` ) - I was thinking that way the Connection object API would be leaking too much internal details (by calling `getState() -> &InviteeFullState`, gaining access to call methods directly on the AriesFSM. But giving it a second thought, if it's just a reference to FSM state object and we properly setup field/method visibility, shouldn't do much harm?
What do you think @mirgee

mirgee:
@Patrik-Stas Awesome, I love this change. As for returning the full state structs instead of just enums, I dislike the idea. I think that the contract between `get_state` and the caller should be as tight as possible, so that it would less prone to change, and the state machine objects themselves should stay private and internal to the logic of the layers above (Inviter, Invitee, Issuer, Holder, ...).

libvcx/src/aries/handlers/connection/connection.rs
```diff
@@ -117,13 +124,13 @@ impl Connection {
         }.into()
     }
 
-    pub fn state(&self) -> u32 {
+    pub fn state(&self) -> ConnectionState {
```
mirgee, line 127: Let's rename this to `get_state`. Function names make more sense as verbs rather than nouns (as opposed to fields).

libvcx/src/abi_utils/mod.rs
```diff
@@ -0,0 +1,28 @@
+use crate::aries::handlers::connection::connection::ConnectionState;
```
mirgee, line 1: How about moving this code to `src/api/mod.rs`, or moving some of the enum conversion stuff to this module, along with `src/api/return_types_u32.rs`?

mirgee (changes requested):

libvcx/src/abi_utils/mod.rs
Patrik-Stas: happy to do that change

libvcx/src/aries/handlers/connection/connection.rs
Patrik-Stas: agreed

libvcx/src/abi_utils/mod.rs
Patrik-Stas: Well I played a bit and took this suggestion few more steps ahead - I turned it into separate PR https://github.com/hyperledger/aries-vcx/pull/295/files so the original purpose of this one does not get too polluted

mirgee: That's perfect, thank you

mirgee (approved):

PR #297, opened 2021-06-15
Revert rust version to 1.45.2, release aries-vcx 0.18.3

Patrik-Stas:
###
- It appears like after upgrading to Rust 1.52.1, we run into a strange issue on iOS where `create_wallet` operating just kind of hangs forever, as if it never executed the callback.
- Downgrade to previously 1.45.2 seems to fix this issue
- Will try to upgrade Rust in next releases to see if (or which) version of rustc introduces this issue

### 
- Release 0.18.3


mirgee:
We should try updating rustc when new version is released to see if the issue is fixed (that's probably what you are suggesting as well). Ideally, we would build multiple pre-releases to find out which version was the bug introduced, but I agree it does not seem worth the time.

PR #300, opened 2021-06-17
Move CloudAgentInfo creation to Connecton constructor

mirgee:
Move `CloudAgentInfo` from `connect()` to `Connection` constructor in order to be able to download messages immediately after the struct creation instead after calling `connect()`.


libvcx/src/aries/handlers/connection/connection.rs
```diff
@@ -826,33 +828,33 @@ pub mod tests {
         thread::sleep(Duration::from_millis(1000));
 
         let all_messages = download_messages_noauth(None, None, None).unwrap();
-        assert_eq!(all_messages.len(), 1);
-        assert_eq!(all_messages[0].msgs.len(), 3);
-        assert!(all_messages[0].msgs[0].decrypted_msg.is_some());
-        assert!(all_messages[0].msgs[1].decrypted_msg.is_some());
+        assert_eq!(all_messages.len(), 2);
```
Patrik-Stas, line 831: Took me a while to understand why this is happening - Alice's cloud agent has 2 `agent connection`s created, but one has no messages.
What happens is that the way `Alice` testagent is constructed in `pub fn setup() -> Alice {` looks like this:
```
let alice = Alice {
                is_active: false,
                config_wallet,
                config_agency,
                connection: Connection::create("tmp_empoty", true).unwrap(),
                credential_handle: 0,
                presentation_handle: 0,
            };
```
The connection it creates is dummy and never used, just to avoid having to handle `Option` everywhere in testing code. But now creating instance of `Connection` also creates `agent connection` in agency.

I am not strongly opinionated, so just an idea: how about instead, in constructor of `Connection`, instead of creating `agent connection`, we would instead execute `connect` transition conditionally based on `autohop`?  (and `connection` action would be the one creating `agent connection` in agency just as before)

Patrik-Stas: Well, but afaik your needs in harness code is to basically have this `connection` autohop enabled, whilst having the original `autohop` usecase having `disabled`. And my suggestion doesn't enable this, it would have to be more granular - perhaps connection could take some Config object with both options
```
struct ConnectionConfig {
   autohop_connection_response: bool,
   autohop_connect: bool
}
```
But having extra options like this  I think comes at the cost of some additional complexity.

Patrik-Stas: hmmm nevermind, I think let's just go with it as is

Patrik-Stas (dismissed):

Patrik-Stas (approved):

PR #301, opened 2021-06-18
Try updating rustc to 1.53.0

mirgee:
Since we discovered an [iOS hangup bug ](https://github.com/hyperledger/aries-vcx/pull/297) when creating a wallet which was fixed by downgrading rustc, a new version of rustc was released. This is to test if the bug is still present in the new release.


mirgee:
The bug still present.

PR #306, opened 2021-06-23
Use libindy with rusql fix for ios build

Patrik-Stas:
Includes new version of libindy with [this](https://github.com/hyperledger/indy-sdk/pull/2402) changes which fixes a bug present on iOS where thread panics on wallet opening.

In order to successfully build for iOS however, we have to use the `vendored` [feature](https://docs.rs/openssl/0.10.35/openssl/#vendored) for openssl dependency. This, in turn, breaks compilation for the musl C environment on Alpine. Hence, `X86_64_ALPINE_LINUX_MUSL_OPENSSL_NO_VENDOR` environment variable was set as a way to not use the `vendored` flag on Alpine. This is a temporary fix, as the plan is to remove the dependency on openssl completely.


Patrik-Stas:
We would have to release indy-sdk to get into android, as for Android we rely on prebuilt libindy binaries

wrappers/ios/ci/build.sh
```diff
@@ -15,7 +15,8 @@ for i in $(ls -t $OPENSSL_PATH); do
     break
 done
 
-INDY_VERSION="v1.16.0"
+#INDY_VERSION="efb7215" # this is "v1.16.0" + rusql update fix + (number of other commits on master branch)
+INDY_VERSION="b4b330ef3" # this is "v1.16.0" tag
```
mirgee, line 19: What is the purpose of changing this from `v1.16.0` to the tag `v1.16.0` commit?

mirgee:
Moreover, why do you link the binary against a different version of the indy crate?
```
...
indy = "1.16.0"
indy-sys = "1.16.0"
...
```

Patrik-Stas:
> Moreover, why do you link the binary against a different version of the indy crate?
> 
> ```
> ...
> indy = "1.16.0"
> indy-sys = "1.16.0"
> ...
> ```

Good catch 👍  thanks

wrappers/ios/ci/build.sh
Patrik-Stas: Experiment ^_^

mirgee:
Successfully verified that the error on wallet opening is not present on iOS device with this build anymore.

.github/workflows/main.yml
```diff
@@ -453,7 +453,7 @@ jobs:
         run: |
           set -x
           sudo rm -rf "/usr/local/share/boost" "$AGENT_TOOLSDIRECTORY" "/usr/local/lib/android" "/usr/share/dotnet"
-          docker run --rm -i --name libvcx --network host $DOCKER_IMG_NAME_LIBVCX \
+          docker run --rm -i --name libvcx --network host -e X86_64_ALPINE_LINUX_MUSL_OPENSSL_NO_VENDOR=true $DOCKER_IMG_NAME_LIBVCX \
```
Patrik-Stas, line 456: Could you please provide some explanation what is this doing / why is this needed?

libvcx/Cargo.toml
```diff
@@ -45,10 +45,10 @@ regex = "1.1.0"
 rust-base58 = "0.0.4"
 rmp-serde = "0.13.7"
 base64 = "0.8.0"
-openssl = { version = "0.10.29" }
+openssl = { version = "0.10.35", features = ["vendored"] }
```
Patrik-Stas, line 48: I was looking at this and got concerned about this

> The vendored copy will not be configured to automatically find the system’s root certificates, but the openssl-probe crate can be used to do that instead.


So I went through actual usage of openssl crate and it turn out it *might* be an overkill. What's being used is:
```
use openssl::sha::sha256
use openssl::bn::BigNum;
```
I somewhat presumed `openssl` is there because `reqwest` needs it, but `reqwest` comes has its own `tls` dependency https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml

So out openssl usage is not actually ever dealing with certificates, so the quote on top of this comment doesn't matter. I think ultimate solution will be to get rid of openssl dependency. 
Perhaps:
-  https://github.com/RustCrypto/hashes for Sha256 
- https://github.com/rust-num/num-bigint for big integer

Patrik-Stas:
@mirgee I am the original submitter of this PR, so I can't click `approve` as you mostly did the heavy lifting.

So please feel free to `approve` on my behalf, just note the comments I've left.

.github/workflows/main.yml
mirgee: Done

mirgee (approved):
Approved on behalf of @Patrik-Stas

PR #313, opened 2021-06-29
Remove unncesessary deps from agency client

mirgee:


Patrik-Stas (changes requested):
lgtm, except for the failing coverage tool. I think we shouldn't trade the libc update for coverage, what do you think? There must be some way to fix the coverage tool

mirgee:
@Patrik-Stas I agree we shouldn't dispose of codecov, but we also have to update libc so that aries-vcx can be used along with popular packages like `reqwest`.

Patrik-Stas:
> @Patrik-Stas I agree we shouldn't dispose of codecov, but we also have to update libc so that aries-vcx can be used along with popular packages like `reqwest`.

Do you suggest to temporarily disable codecov in favor of libc update, or hold this till we figure out how to get codecov working with this change?

mirgee:
@Patrik-Stas If I preferred disabling codecov in favor updating libc, I would do so in the PR. As I said, I think we need both new libc version and codecov in the long run, I am not suggesting trading one for the other. This means figuring why they are not compatible and fixing it (if possible).

mirgee:
@Patrik-Stas Turned out to be a problem with `core-foundation`a transitive dependency of reqwest library used in `agency_client`. Updating reqwest seems to have fixed the issue.

mirgee:
It turned out that updating `reqwest` requires updating `futures`, which in turn would mean making `post_to_agency` async and thus making all upstream functions async, or explicitly entering tokio threadpool context before every call and then blocking the thread when waiting for agency response (which I assume we are doing now, albeit implicitly).

Between updating just libc and skipping codecov, updating libc, reqwest and futures and making all upstream functions async, and leaving the updates for a later date, I would skip the updates for now.

Issue #196, opened 2020-11-08
Follow hexagonal architecture pattern


Patrik-Stas:
# The problem
Currently, the only way to consume `aries-vcx` is by making FFI calls from provided wrappers. As ironic as it is, even though the library is written in Rust, we don't have Rust wrapper. Esentially the only way to consume `aries-vcx` from external crate (such as Rust agent implementation) is by working directly with internals. 

## Solution 1.  - Rust wrapper + FFI
Safer option would be to actually create Rust FFI wrapper, much like has been done for IndySDK https://github.com/hyperledger/indy-sdk/tree/master/wrappers/rust

Pros:
+ API is guaranteed to be the same in every language.

Cons
- Extra wrapper to maintain and test
- The only guarantee that wrapper is matching the FFI is by actually testing every function from the wrapper
- As the wrapper is completely standalone, `aries-vcx` binary would be its runtime dependency. An user would need to make sure his version of wrapper is matching version of binary on host's the system.

# Solution 2 - Turning Aries-vcx into consumable crate
Better solution however, would be to turn `aries-vcx` into consumable crate directly, whilst preserving consistent API with FFI API. Currently we only have `api` module which however expose C-Callable API, so we would new API layer. Here's steps I suggest to take to get desirable results:

1. First, take files `src/*.rs` and move them to new `service` layer, so we could give this layer some actual name.
2. Rename `api` to `api_c`. Then `api_c` module to be more "pure", as currently it does more than handling FFI calls, more specifically it:
     - Some function are reading data from `settings.rs`, passing it down to further called functions (see 
  `vcx_schema_prepare_for_endorser` for example).
     - Spawning threads.
3. Create new module `api`. These 2 extra responsibilities (spawning, settings access) should be extracted out from `api_c` into new `api`. Every call to `api_c` would be delegated to similarly named function in `api`. Every call into `api` would spawn thread, perhaps access `settings,rs` and call appropriate function in service layer.

By doing this `api` would become layer which could be public on a crate layer. Whether someone would like to use `aries-vcx` via FFI, or as crate dependency, he would get the same API. We would add extra layer to call, but that's arguably easier than maintaining wrapper. It would also enhance testability, as the tests which are currently in `api_c` module could be moved to the new `api` module and wouldn't have to deal with expressing C function signatures and managing C callbacks.

Pros:
+ Low maintenance efforts, complexity of mapping FFI Wrapper onto C-Callable interface is gone.
[...]

PR #320, opened 2021-07-19
Separate aries module into a crate

mirgee:
Another step as part of [this](https://github.com/hyperledger/aries-vcx/issues/196) epic.


Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 
 members = [
+    "aries",
```
Patrik-Stas, line 4: I'd suggest calling the crate `aries-vcx`, there's already `aries` in crates.io anyway

aries/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+version = "0.1.0"
+authors = ["Absa Group Limited", "Hyperledger Indy Contributors <hyperledger-indy@lists.hyperledger.org>"]
+publish = false
+license = "Apache-2.0"
+edition = "2018"
+
+[lib]
+name = "aries"
+path = "src/lib.rs"
+crate-type = ["staticlib","rlib", "cdylib"]
```
Patrik-Stas, line 13: I suppose these could be removed, as it would no longer be compiled as static/dynamic library

aries/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+
+name = "aries"
+version = "0.1.0"
```
Patrik-Stas, line 4: I suppose it's good idea to version `libvcx` and `aries-vcx` separately, but maybe it'd be better to start with everything on the same version, and then decouple them as a next steps? It will def need  CI rethinking and changes

aries/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+path = "src/lib.rs"
+crate-type = ["staticlib","rlib", "cdylib"]
+
+[dependencies]
+env_logger = "0.5.10"
+log = "0.4"
+chrono = "0.4"
+time = "0.1.36"
+lazy_static = "1.3"
+libc = "=0.2.66"
```
Patrik-Stas, line 21: I think this could be removed?

aries/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+regex = "1.1.0"
+rust-base58 = "0.0.4"
+rmp-serde = "0.13.7"
+base64 = "0.8.0"
+openssl = { version = "0.10.35", features = ["vendored"] }
+num-traits = "0.2.0"
+indy = "1.16.0-post-59"
+indy-sys = "1.16.0-post-59"
+futures = "0.3.15"
+tokio = { version = "1.7.1", features = ["rt-multi-thread"] }
```
Patrik-Stas, line 36: also this

aries/Cargo.toml
mirgee: I think this is still a draft? 😜

mirgee: Libc is still used.

aries/Cargo.toml
mirgee: Yes

aries/Cargo.toml
mirgee: Correct

Cargo.toml
mirgee: `aries_vcx`, so it's easily addressable

aries_vcx/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+
+name = "aries"
```
Patrik-Stas, line 3: `aries_vcx`

aries_vcx/src/error.rs
```diff
@@ -0,0 +1,550 @@
+use std::cell::RefCell;
```
Patrik-Stas, line 1: I think this file can be reuse from `libvcx`? (it's currently duplicated)

Patrik-Stas:
I think these files:
-  `aries_vcx/src/libindy/utils/callback.rs`:
-  `aries_vcx/src/libindy/utils/callback_u32.rs`:
can be moved to `libvcx`


Bonus points for this ^_^:
-  `aries_vcx/src/libindy/utils/cstring.rs`, however that is dependency of `aries_vcx/src/libindy/utils/logger.rs` and that's dependency of `devsetup.rs`. But I think it should be possible to adjust devsetup not to use `aries_vcx/src/libindy/utils/logger.rs`, instead configure rust logger directly in `devsetup.rs` without all the C-stuff.

aries_vcx/src/error.rs
mirgee: Not sure that's a good pattern. I think each library should have its own error types specific to its use-cases, purposes, concerns, APIs, and flows. Else, adding a new error type would require modification of a different library.

Patrik-Stas:
👍  By the way, an unconsidered pleasant effect of splitting into `libvcx` and `aries_vcx` is that as user of libvcx, I have now better control over logging granularity - I can do for example:
```
VCX_LOG_LEVEL=aries_vcx=info,vcx=error npm run test:integration:update-state
```

Patrik-Stas:
I suggest changing:
```

[lib]
name = "vcx"
```
to
```

[lib]
name = "libvcx"
```

in `libvcx/Cargo.toml`

aries_vcx/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+
+name = "aries-vcx"
+version = "0.1.0"
+authors = ["Absa Group Limited", "Hyperledger Indy Contributors <hyperledger-indy@lists.hyperledger.org>"]
+publish = false
+license = "Apache-2.0"
+edition = "2018"
+
+[lib]
```
Patrik-Stas, line 10: This section can be removed

Patrik-Stas:
```
libloading = "0.5.0"
```
can be removed from `libvcx`

Patrik-Stas:
`strum` can be removed from `libvcx`

aries_vcx/Cargo.toml
mirgee: I woul remove the edition from all packages in the workspace since 2018 is the default. I am not sure about removing the licence, though.

Patrik-Stas:
`indy`, `indy-sys` can be removed from `libvcx`

mirgee:
@Patrik-Stas  `indy`, `indy_sys`, `serde`, `serde_json` are still used

mirgee:
> I suggest changing:

That would make the resulting library called `liblibvcx` :)

Patrik-Stas (approved):

PR #328, opened 2021-08-03
Ci/fix test execution

Patrik-Stas:
- Tests in `aries_vcx` were not being executed in CI 
- Some tests in `aries_vcx` did not compile,  needed small tweak
- Modification in `aries_vcx` did not trigger `libvcx` image rebuild

### DIFF
--- a/aries_vcx/src/handlers/proof_presentation/prover/prover.rs
+++ b/aries_vcx/src/handlers/proof_presentation/prover/prover.rs
@@ -146,7 +146,6 @@ mod tests {
     use crate::utils::get_temp_dir_path;
 
     use super::*;
-    use crate::settings::{get_config_value, CONFIG_INSTITUTION_DID};
 
     #[test]
     #[cfg(feature = "pool_tests")]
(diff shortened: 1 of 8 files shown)


aries_vcx/src/handlers/proof_presentation/prover/prover.rs
```diff
@@ -198,7 +199,7 @@ mod tests {
         let setup = SetupLibraryWalletPoolZeroFees::init();
 
         libindy::utils::anoncreds::tests::create_and_store_credential(utils::constants::DEFAULT_SCHEMA_ATTRS, false);
-        let did = setup.faber.config_issuer.institution_did.clone();
+        let did = get_config_value(CONFIG_INSTITUTION_DID).unwrap();
```
mirgee, line 202: Just and idea - how about hardcoding the institution did into the tests, or moving it `mockdata/` or `constants.rs`? The reason is I am not enjoying adding another dependency on `settings.rs`

mirgee (changes requested):

aries_vcx/src/handlers/proof_presentation/prover/prover.rs
Patrik-Stas: I don't object, but alternatively I'd suggest to let `SetupLibraryWalletPoolZeroFees` return the value of institution did

mirgee: OK (I would prefer an object with the value of institution did)

mirgee (approved):

Issue #330, opened 2021-08-09
Supersede vcx_v2_messages_download


Patrik-Stas:
We found that `vcx_v2_messages_download` is currently exposing API which doesn't play nice with with language wrapper, which don't have `u32` / `uint` representation of handles. 

In particular, the Rust API returns `u32` handles as references to various objects stored in memory on `libvcx` layer. Java wrapper types these values as `int`, which is 32bits long number, but includes negative range, eg. equivalent of `i32` in Rust. Because of this, on Java layer, these values appear negative.

This is generally not problem, as far as the C-APIs deal with these handles as numbers - it doesn't matter the value "appears" negative in Java, the sequence of bytes is always the same, hence the memory reference is valid.

The problem is that 
```
pub extern fn vcx_v2_messages_download(command_handle: CommandHandle,
                                       conn_handles: *const c_char,
                                       message_statuses: *const c_char,
                                       uids: *const c_char,
                                       cb: Option<extern fn(xcommand_handle: CommandHandle, err: u32, messages: *const c_char)>) -> u32 {
```
is taking `conn_handles`, which is supposed to be JSON array of handles. Because these in Java appear negative, in result what we pass down to Rust is something like:
```
conn_handles = "[-15161235, -2161243]"
```
And we then try to parse these negative numbers into unsigned 32 bit int, which is not possible without dirty tricks (doing unsafe conversion based on whether received number is negative or positive)

Possible solutions could are following.
1. Change `conn_handles` to be array of `u32`s, but then we probably also need additional parameter to define length of the array. 
2. Implement new function to download messages for particular connection only, which removes the problem with passing down array of numeric u32 handles. This would also just make the API more user friendly, as one could then request download of messages on particular connection object like:
```
connection.downloadMessages(STATUS_PENDING, ["uid123", "uid456"])
```
instead of using utility function as currently:
```
downloadMessages([connection1.handle], STATUS_PENDING, ["uid123", "uid456"])
```
With this solution, if you want to download messages for more connections, you'll have to call the function on each of them. This is not less efficient than previous implementation with `vcx_v2_messages_download`, as this itself was looping the connections internally on Rust layer.

Patrik-Stas:
Implemented

PR #331, opened 2021-08-09
Download messages via a connection handle as u32

mirgee:
The Java wrapper currently stores handles as `int` datatypes, which are converted back to `u32` when calling Rust FFI. However, the signature of `vcx_v2_messages_download` takes a list of handles as string. When passed the handles represented as `int`s through the Java API, the function fails. To enable safe message downloads from Java wrappers, this PR exposes a function which takes a single connection handle instead of a string. See #330 .


Patrik-Stas (changes requested):
Looks good, just please add:
- NodeJS integration test,
- deprecation for former download functions.

Patrik-Stas (approved):

PR #340, opened 2021-08-23
Support invitation messages with public DID

mirgee:
Moved from [here](https://github.com/hyperledger/aries-vcx/pull/7).

Adds libvcx API:

### Pubic agents
- `vcx_public_agent_create` - create new public agent 
- `vcx_public_agent_generate_public_invite` - generate public DID based invitation with custom label
- `vcx_public_agent_serialize` - serialize public agent info
- `vcx_public_agent_deserialize` - deserialize public agent info into an Public Agent object
- `vcx_public_agent_release`- release Public Agent object from libvcx memory state
- `vcx_public_agent_download_connection_requests` - downloads all received connection requests (or a specific connection request based on uid), marks them as Reviewed in agency, and marks all non-request messages as rejected

### Connections
- `vcx_connection_create_with_connection_request` - create new Connection from Aries "Connection Request" message

# Workflow
The idea is
- Create "public agent" in Agency, write its info on ledger under your public DID
- Generate Connection Invitation containing your DID
- Publish Connection Invitation somewhere publicly (for example as QR code). Anyone can see, use it, scan it.
- Check your Public Agent for newly received Aries messages of "Connection Request" type
- For each new "Connection Request", create new Connection object and progress the aries connection protocol as usual


agents/node/vcxagent-core/test/utils/faber.js
```diff
@@ -137,6 +151,31 @@ module.exports.createFaber = async function createFaber () {
     return agencyMessages
   }
 
+  async function _downloadConnectionRequests () {
+    logger.info('Faber is going to download connection requests')
+    let agencyMessages = await vcxAgent.serviceConnections.getAllMessages()
```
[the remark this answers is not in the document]
Patrik-Stas: Good question....maybe we could make it even a bit different
```
const request = await publicAgent.downloadConnectionRequest()
const connection = await createWithConnectionRequest(request)
connection.updateState()
```
have
```
const connection = await publicAgent.tryCreateConnectionFromConnectionRequest()
connection.updateState()
```
so the public agent would try to find connection request, and if found, it would update message in agency and return the connection. This would be basically equivalent to how `.updateState()` operates.

Otherwise
```
const request = await publicAgent.downloadConnectionRequest()
```
would have to return request message, but in "agency format", eg. including `uid`, so you could manually update the message state. But I'd rather go for the first option for now.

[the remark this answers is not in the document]
Patrik-Stas: About further filters - one thing which comes to my mind is how to delete non-ConnectionRequest messages. It *should not* happen that we have other messages on public agent, but if someone for whatever reason sends there something else, we should have way to clean up these things. It could be either convenient Rust API like `public_agent_clean_up` which update status of all non-ConnectionRequests as processed. 
Or it could be a generic API, like `public_agent_download_messages` which would return messages in the agency format (eg. including msgUid, status etc.) and it would be up to caller to call this with a proper filter and update messages in agency. I tend to lean toward the more convenient "cleanUp" API here again.

aries_vcx/src/handlers/connection/invitee/states/null.rs
```diff
@@ -1,12 +1,19 @@
 use crate::handlers::connection::invitee::states::invited::InvitedState;
-use crate::messages::connection::invite::Invitation;
+use crate::messages::connection::invite::{Invitation, PairwiseInvitation, PublicInvitation};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct NullState {}
 
-impl From<(NullState, Invitation)> for InvitedState {
-    fn from((_state, invitation): (NullState, Invitation)) -> InvitedState {
+impl From<(NullState, PairwiseInvitation)> for InvitedState {
```
Patrik-Stas, line 7: can't we use just the wrapping `Invitation` instead of having separate transition functions for PairwiseInvitation and PublicInvitation? The code for both seems the same

aries_vcx/src/handlers/connection/inviter/states/invited.rs
```diff
@@ -10,16 +10,17 @@ pub struct InvitedState {
     pub invitation: Invitation,
 }
 
-impl From<(InvitedState, ProblemReport)> for NullState {
-    fn from((_state, _error): (InvitedState, ProblemReport)) -> NullState {
-        trace!("ConnectionInviter: transit state from InvitedState to NullState");
+// TODO: These have no justification for being here anymore
```
Patrik-Stas, line 13: Why is that? I see we still use this, for example  here
```
aries_vcx/src/handlers/connection/inviter/state_machine.rs:304:48
```

aries_vcx/src/handlers/connection/invitee/states/null.rs
Patrik-Stas: This pattern can be found also in `aries_vcx/src/handlers/connection/inviter/states/null.rs`

aries_vcx/src/handlers/connection/inviter/states/null.rs
```diff
@@ -1,12 +1,19 @@
 use crate::handlers::connection::inviter::states::invited::InvitedState;
-use crate::messages::connection::invite::Invitation;
+use crate::messages::connection::invite::{Invitation, PairwiseInvitation, PublicInvitation};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct NullState {}
 
-impl From<(NullState, Invitation)> for InvitedState {
-    fn from((_state, invitation): (NullState, Invitation)) -> InvitedState {
+impl From<(NullState, PairwiseInvitation)> for InvitedState {
```
Patrik-Stas, line 7: Some comment as for `aries_vcx/src/handlers/connection/invitee/states/null.rs`

aries_vcx/src/libindy/utils/cache.rs
```diff
@@ -151,3 +151,24 @@ pub fn clear_rev_reg_delta_cache(rev_reg_id: &str) -> VcxResult<String> {
         Err(VcxError::from(VcxErrorKind::IOError))
     }
 }
+
+pub fn get_from_cache<'a, T: serde::de::DeserializeOwned>(prefix: &str, id: &str) -> VcxResult<T> {
```
Patrik-Stas, line 155: seems not to be used

libvcx/src/api_lib/utils/error.rs
```diff
@@ -158,8 +159,7 @@ impl From<u32> for VcxErrorKind {
             _ if { error::INVALID_REVOCATION_TIMESTAMP.code_num == code } => VcxErrorKind::InvalidRevocationTimestamp,
             _ if { error::INVALID_SCHEMA_CREATION.code_num == code } => VcxErrorKind::CreateSchema,
             _ if { error::INVALID_SCHEMA_HANDLE.code_num == code } => VcxErrorKind::InvalidSchemaHandle,
-            _ if { error::INVALID_SCHEMA_SEQ_NO.code_num == code } => VcxErrorKind::InvalidSchemaSeqNo,
-            _ if { error::DUPLICATE_SCHEMA.code_num == code } => VcxErrorKind::DuplicationSchema,
+            _ if { error::INVALID_SCHEMA_SEQ_NO.code_num == code } => VcxErrorKind::InvalidSchemaSeqNo, _ if { error::DUPLICATE_SCHEMA.code_num == code } => VcxErrorKind::DuplicationSchema,
```
Patrik-Stas, line 162: Why is it? This will actually break ou code built on top of aries-vcx.

Patrik-Stas (changes requested):
a few changes requested

aries_vcx/src/libindy/utils/cache.rs
mirgee: I told you I know about this

libvcx/src/api_lib/utils/error.rs
mirgee: What are you referring to? This change just accidentally removed a newline.

aries_vcx/src/handlers/connection/inviter/states/invited.rs
mirgee: The comment does not say it is not used, but that it has no justification for being in the folder dedicated to inviter state.

agents/node/vcxagent-core/test/utils/faber.js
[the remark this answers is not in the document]
mirgee: As discussed, rejecting non-request messages when downloading now.

Patrik-Stas (approved):

Issue #188, opened 2020-11-05
Use correct @type values in aries protocols


Patrik-Stas:
Currently `@type` values use incorrect prefix:
```
did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/invitation
did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/1.0/request-presentation
```

According to RFCs (https://github.com/hyperledger/aries-rfcs/blob/master/features/0160-connection-protocol/README.md, https://github.com/hyperledger/aries-rfcs/tree/master/features/0037-present-proof .. ), prefix `https://didcomm.org/` should be used
```
https://didcomm.org/connections/1.0/invitation
https://didcomm.org/present-proof/1.0/request-presentation
```

This goes for all `@type` annotations used accross the codebase. It's apparently some legacy format which were brough in for backward compatibility (by look at acapy codebase https://github.com/hyperledger/aries-cloudagent-python/search?q=https%3A%2F%2Fdidcomm.org )

I am not sure if there's some agents using the legacy `did:sov:BzCbsNYhMrjHiqZDTUASHg;spec`, perhaps we can support both for now and drop the old one later. Acapy also seems to support both.

PR #341, opened 2021-08-30
Use correct prefix on created @type on aries messages.

Patrik-Stas:
Addressing issue https://github.com/hyperledger/aries-vcx/issues/188

Previously Aries messages created by `aries-vcx` used `did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/` prefix for `@type` attribute. 
According to https://github.com/hyperledger/aries-rfcs/blob/08653f21a489bf4717b54e4d7fd2d0bdfe6b4d1a/concepts/0020-message-types/README.md#aries-core-message-namespace the prefix for core aries protocols should be `https://didcomm.org/`
> https://didcomm.org/ is used to namespace protocols defined by the community as "core protocols" or protocols that agents should minimally support.

This PR updates aries-vcx to create messages with the correct prefix. 

The support for legacy prefix `did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/` is preserved so this PR is not breaking change, however it can be expected the support for this prefix will be removed in future releases. 

This PR also adds tests for serialization/deserialization of aries messages in general.


aries_vcx/src/messages/a2a/message_type.rs
```diff
@@ -52,6 +52,73 @@ impl Serialize for MessageType {
 
 impl std::string::ToString for MessageType {
     fn to_string(&self) -> String {
-        format!("{};spec/{}/{}/{}", self.did, self.family.to_string(), self.version, self.type_)
+        format!("{}/{}/{}/{}", self.prefix, self.family.to_string(), self.version, self.msg_type)
+    }
+}
+
+
+pub fn parse_message_type_legacy(message_type: &str) -> Option<MessageType> {
```
Patrik-Stas, line 60: FYI for review:

The code in `parse_message_type_legacy` and `parse_message_type` was copied here from `use agency_client::message_type::parse_message_type` to make it possible to have different behaviour for aries messages and agency messages (`@type` encoding changes in client2agency protoco are not desirable at the moment).

Patrik-Stas: Also it feels cleaner that this module does not depend on `agency_client`

Patrik-Stas: When in future we rewrite client2agency protocol to be aries comptabile, we can extract `message_type`, `message_family` etc.  and reuse it from both `aries-vcx` and `agency-client`

mirgee: Why are they not desirable? I think it is desirable that forward messages are treated the same as client-to-client messages, but the problem is that agency does not support both formats ATM, correct? If that's so, we should implement the new format in the agency ASAP and get rid of this duplication by extracting the types into a separate module used both by `agency-client` and by `aries-vcx` - do you agree?

aries_vcx/src/utils/mockdata/mockdata_connection.rs
```diff
@@ -2,7 +2,7 @@
 pub const ARIES_CONNECTION_INVITATION: &str = r#"
 {
     "@id": "28b39b79-f5db-4478-879a-15bb12632d00",
-    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/invitation",
+    "@type": "https://didcomm.org/connections/1.0/invitation",
```
mirgee, line 5: Do we test deserialization of the legacy format somewhere?

mirgee (changes requested):

aries_vcx/src/utils/mockdata/mockdata_connection.rs
Patrik-Stas: yup I've created tests `test_deserialize_connection_ack_legacy, test_serialize_forward_message_to_legacy_format` to assure this PR is not introducing breaking change

mirgee: The test does not seem to test deserialization of the legacy format, however?

Patrik-Stas: fixed

Patrik-Stas: good catch! Must have find+replace by mistake. Fixed now

aries_vcx/src/messages/a2a/message_type.rs
Patrik-Stas: I agree, it should be done. To clarify, I didn't want to make the scope of the PR too broad, aiming primarily fixing the issue at hand (e2e aries interoperability issue).

mirgee (approved):

PR #343, opened 2021-09-03
Remove unused service module, split service into enum

mirgee:
Removes unused `service` module, replaces used `Service` struct with an enum allowing for easy interchangeability between a did and a service (becomes slightly useful in Out-of-Band implementation).

### DIFF
--- a/aries_vcx/src/messages/connection/invite.rs
+++ b/aries_vcx/src/messages/connection/invite.rs
@@ -1,4 +1,5 @@
 use crate::messages::a2a::{A2AMessage, MessageId};
+use crate::messages::connection::did_doc::Did;
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(untagged)]
@@ -26,7 +27,7 @@ pub struct PublicInvitation {
     #[serde(rename = "@id")]
     pub id: MessageId,
     pub label: String,
-    pub did: String,
+    pub did: Did,
 }
 
 impl PairwiseInvitation {
--- a/aries_vcx/src/messages/connection/service.rs
+++ b/aries_vcx/src/messages/connection/service.rs
@@ -1,75 +1,53 @@
-use crate::messages::connection::did_doc::DidDoc;
+use crate::messages::connection::did_doc::Did;
+use crate::libindy::utils::ledger;
+use crate::error::prelude::*;
 
-#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Default)]
-#[serde(rename_all = "camelCase")]
-pub struct Service {
-    #[serde(default)]
-    pub recipient_keys: Vec<String>,
-    pub routing_keys: Option<Vec<String>>,
-    pub service_endpoint: String,
+pub const SERVICE_SUFFIX: &str = "indy";
+pub const SERVICE_TYPE: &str = "IndyAgent";
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub enum Service {
+    FullService(FullService),
+    Did(Did)
 }
 
 impl Service {
-    pub fn create() -> Self {
-        Service::default()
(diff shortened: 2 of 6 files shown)


aries_vcx/src/messages/connection/service.rs
```diff
@@ -1,17 +1,46 @@
...
-    pub routing_keys: Option<Vec<String>>,
+    #[serde(default)]
+    #[serde(rename = "routingKeys")]
+    pub routing_keys: Vec<String>,
+    #[serde(rename = "serviceEndpoint")]
     pub service_endpoint: String,
 }
 
-impl Service {
+impl FullService {
```
Patrik-Stas, line 41: Since we renamed `Service` to `ServiceResolvable` (in the original version of the PR, I think we can keep here the original `Service` name

aries_vcx/src/messages/connection/service.rs
```diff
@@ -20,7 +49,7 @@ impl Service {
     }
 
     pub fn set_routing_keys(mut self, routing_keys: Vec<String>) -> Self {
-        self.routing_keys = Some(routing_keys);
+        self.routing_keys = routing_keys;
```
Patrik-Stas, line 52: generally speaking routing keys can be undefined, but if this makes things easier for now, I don't mind the change

aries_vcx/src/messages/connection/invite.rs
```diff
@@ -26,7 +27,7 @@ pub struct PublicInvitation {
     #[serde(rename = "@id")]
     pub id: MessageId,
     pub label: String,
-    pub did: String,
+    pub did: Did,
```
Patrik-Stas, line 30: 👍

Patrik-Stas (changes requested):
left a comment

aries_vcx/src/messages/connection/service.rs
mirgee: Personally, I am torn on this one. I think that `Service` sounds way too general in a situation where there is a general enum representing a service, but possibly only a resolvable reference to it. There is an argument for calling the enum `Service` (since it is more general), and the struct `FullService` / `Service` as well. Hence, I would like to disambiguate between them very clearly by naming them very descriptively and differently.

aries_vcx/src/messages/connection/service.rs
mirgee: It is not strictly necessary, but it does simplify the code slightly and we don't have a use-case for undefined routing keys ATM AFAIK

aries_vcx/src/messages/connection/service.rs
Patrik-Stas: Idk I just don't see where is the `Full` coming from. The naming makes me think "okay this is `FullService`, so do we have some sort of 'incomplete'/'unfull' variant of service too?". It seems to me more accurate to just call the struct `Service`, since that's how tie field it reflects is called in DDO. 

Now that "wrapping enum" is called `ServiceResolvable `, whether you call this struct `FullService`, or `Service`, it's equally distinct from `ServiceResolvable` and I don't see an ambiguity in either case.

One more idea to throw in - what about calling it `DdoService`?

These are my thought's, I'll leave the final decision up to you, in the end this is mostly internal naming, so not too big of a deal.

mirgee: > Idk I just don't see where is the Full coming from. The naming makes me think "okay this is FullService, so do we have some sort of 'incomplete'/'unfull' variant of service too?". It seems to me more accurate to just call the struct Service, since that's how tie field it reflects is called in DDO.

I agree with that thinking, and to me, that "unfull" variant of the service is just the did.

> Now that "wrapping enum" is called ServiceResolvable , whether you call this struct FullService, or Service, it's equally distinct from ServiceResolvable and I don't see an ambiguity in either case.

My point was different - that the enum could be reasonably called `Service`.

PR #344, opened 2021-09-03
Out of Band protocol

mirgee:


aries_vcx/src/handlers/out_of_band/sender/sender.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    pub fn append_service(mut self, service: ServiceResolvable) -> Self {
+        self.services.push(service);
+        self
+    }
+
+    pub fn append_handshake_protocol(mut self, protocol: HandshakeProtocol) -> VcxResult<Self> {
+        let new_protocol = match protocol {
+            HandshakeProtocol::ConnectionV1 => MessageType::build(MessageFamilies::Connections, ""),
+            HandshakeProtocol::DidExchangeV1 => { return Err(VcxError::from(VcxErrorKind::ActionNotSupported)) }
```
Patrik-Stas, line 37: Good reminder of protocol work ahead of us :-) Went ahead and created an issue to track this
https://github.com/hyperledger/aries-vcx/issues/346

aries_vcx/tests/test_integration.rs
```diff
@@ -1001,38 +1014,104 @@ mod tests {
...
+            .set_goal("To exchange message")
+            .append_service(ServiceResolvable::FullService(service));
+        let oob_msg = oob_sender.to_a2a_message();
 
         consumer.activate().unwrap();
-        let consumer_msgs = consumer_to_institution.download_messages(Some(vec![MessageStatusCode::Received]), None).unwrap();
-        assert_eq!(consumer_msgs.len(), 1);
+        let oob_receiver = OutOfBand::create_from_a2a_msg(&oob_msg).unwrap();
+        let conn = oob_receiver.connection_exists(vec![&consumer_to_institution]).unwrap();
+        assert!(conn.is_some());
```
Patrik-Stas, line 1114: ideally let's have Alice create more than 1 connection and make sure that `connection_exists` picks the right one. 
We should also test it returns None, if no matching one is supplied

Patrik-Stas: We could also test the connection will not match if the original connection was created "privately", instead of  public did invite.

aries_vcx/tests/test_integration.rs
```diff
@@ -1001,38 +1014,104 @@ mod tests {
...
+            .append_a2a_message(request_sender.to_a2a_message()).unwrap();
+        let oob_msg = oob_sender.to_a2a_message();
 
         consumer.activate().unwrap();
-        let mut consumer_to_institution = Connection::create_with_invite("institution", public_invite, true).unwrap();
-        consumer_to_institution.connect().unwrap();
-        consumer_to_institution.update_state().unwrap();
+        let oob_receiver = OutOfBand::create_from_a2a_msg(&oob_msg).unwrap();
+        let conn = oob_receiver.connection_exists(vec![]).unwrap();
+        assert!(!conn.is_some());
```
Patrik-Stas, line 1056: `is_none()`

aries_vcx/src/handlers/out_of_band/receiver/receiver.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl OutOfBand {
+    pub fn create_from_a2a_msg(msg: &A2AMessage) -> VcxResult<Self> {
+        match msg {
+            A2AMessage::OutOfBand(oob) => Ok(oob.clone()),
+            _ => Err(VcxError::from(VcxErrorKind::InvalidMessageFormat))
+        }
+    }
+
+    pub fn connection_exists<'a>(&self, connections: Vec<&'a Connection>) -> VcxResult<Option<&'a Connection>> {
```
Patrik-Stas, line 22: I think before resolving `full_service` of the oob, we should just try to match by DID itself - if available on connection ofcourse.
With current code, if:
- i establish connection via public invite, using their did `sov:12345`
-  institutions updates their `service` on ledger for their did `sov:12345`
- I find scan their OOB message which contains `services: ["sov:12345"]`

the code would determine these connection not to match. Which is not desired behaviour I think. Also, this scenario should be covered by a test

Patrik-Stas: Also I've realized I am not quite sure how well the reference return will actually work once we try to expose these features on handle/ffi layer. You'll have to find out the handle of the returned reference, so we could return that handle via FFI. But I guess we'll find out once we start implementing that.

aries_vcx/tests/test_integration.rs
```diff
@@ -1001,38 +1014,104 @@ mod tests {
...
+            .append_handshake_protocol(HandshakeProtocol::ConnectionV1).unwrap()
+            .append_a2a_message(request_sender.to_a2a_message()).unwrap();
+        let oob_msg = oob_sender.to_a2a_message();
 
         consumer.activate().unwrap();
-        let mut consumer_to_institution = Connection::create_with_invite("institution", public_invite, true).unwrap();
-        consumer_to_institution.connect().unwrap();
-        consumer_to_institution.update_state().unwrap();
+        let oob_receiver = OutOfBand::create_from_a2a_msg(&oob_msg).unwrap();
+        let conn = oob_receiver.connection_exists(vec![]).unwrap();
```
Patrik-Stas, line 1055: there's not much point calling this if you only pass in empty vector though

aries_vcx/tests/test_integration.rs
```diff
@@ -1001,38 +1014,104 @@ mod tests {
...
+        conn_sender.update_state().unwrap();
+        assert_eq!(ConnectionState::Inviter(InviterState::Completed), conn_sender.get_state());
 
-        institution_to_consumer.send_generic_message("Hello Alice, Faber here").unwrap();
+        let conn = oob_receiver.connection_exists(vec![&conn_receiver]).unwrap();
+        assert!(conn.is_some());
+
+        let a2a_msg = oob_receiver.extract_a2a_message().unwrap().unwrap();
+        assert!(matches!(a2a_msg, A2AMessage::PresentationRequest(..)));
+        if let A2AMessage::PresentationRequest(request_receiver) = a2a_msg {
```
Patrik-Stas, line 1087: I suggest only using BasicMessage for this test, not a big deal but:
- no need to create schema and cred def
- it's clearer this test is not testing oob+proofs, as it might appear at quick glance

aries_vcx/tests/test_integration.rs
```diff
@@ -1001,38 +1014,104 @@ mod tests {
...
+            .set_goal_code(GoalCode::P2PMessaging)
+            .set_goal("To exchange message")
+            .append_service(ServiceResolvable::FullService(service));
+        let oob_msg = oob_sender.to_a2a_message();
 
         consumer.activate().unwrap();
-        let consumer_msgs = consumer_to_institution.download_messages(Some(vec![MessageStatusCode::Received]), None).unwrap();
-        assert_eq!(consumer_msgs.len(), 1);
+        let oob_receiver = OutOfBand::create_from_a2a_msg(&oob_msg).unwrap();
+        let conn = oob_receiver.connection_exists(vec![&consumer_to_institution]).unwrap();
```
Patrik-Stas, line 1113: I'd like to see BasicMessage exchanged to verify the connection is truly reused and message arrives where expected

aries_vcx/tests/test_integration.rs
```diff
@@ -1001,38 +1014,104 @@ mod tests {
...
+        assert!(matches!(a2a_msg, A2AMessage::PresentationRequest(..)));
+        if let A2AMessage::PresentationRequest(request_receiver) = a2a_msg {
+            assert_eq!(request_receiver.thread.unwrap().pthid.unwrap(), oob_receiver.id.0);
+            assert_eq!(request_receiver.request_presentations_attach, request_sender.request_presentations_attach);
+        }
+    }
+
+    #[test]
+    #[cfg(feature = "agency_pool_tests")]
+    fn test_connection_reuse() {
```
Patrik-Stas, line 1095: I suggest rename: `test_oob_connection_reuse`

aries_vcx/tests/test_integration.rs
```diff
@@ -1001,38 +1014,104 @@ mod tests {
...
+        institution_to_consumer.send_generic_message("Hello Alice, Faber here").unwrap();
+
+        consumer.activate().unwrap();
+        let consumer_msgs = consumer_to_institution.download_messages(Some(vec![MessageStatusCode::Received]), None).unwrap();
+        assert_eq!(consumer_msgs.len(), 1);
+    }
+
+    #[test]
+    #[cfg(feature = "agency_pool_tests")]
+    fn test_establish_connection_via_oob_message_request_included() {
```
Patrik-Stas, line 1028: I suggest rename: `test_oob_connection_bootstrap`

aries_vcx/tests/test_integration.rs
```diff
@@ -1001,38 +1014,104 @@ mod tests {
...
         thread::sleep(Duration::from_millis(500));
-        let mut conn_requests = institution.agent.download_connection_requests(None).unwrap();
-        assert_eq!(conn_requests.len(), 1);
-        let mut institution_to_consumer = Connection::create_with_connection_request(conn_requests.pop().unwrap(), &institution.agent).unwrap();
-        assert_eq!(ConnectionState::Inviter(InviterState::Requested), institution_to_consumer.get_state());
-        institution_to_consumer.update_state().unwrap();
-        assert_eq!(ConnectionState::Inviter(InviterState::Responded), institution_to_consumer.get_state());
+        let mut requests = institution.agent.download_connection_requests(None).unwrap();
+        assert_eq!(requests.len(), 1);
+        let mut conn_sender = Connection::create_with_connection_request(requests.pop().unwrap(), &institution.agent).unwrap();
```
Patrik-Stas, line 1067: Seems to be decent fragment of code duplicated in following lines and in `create_connected_connections_via_public_invite` - I think if `create_connected_connections_via_public_invite` is perhaps split in 2 functions / generalized, we could get rid of the duplicity. 

Following lines seems to be duplicate:
```
 let mut requests = institution.agent.download_connection_requests(None).unwrap();
        assert_eq!(requests.len(), 1);
        let mut conn_sender = Connection::create_with_connection_request(requests.pop().unwrap(), &institution.agent).unwrap();
        assert_eq!(ConnectionState::Inviter(InviterState::Requested), conn_sender.get_state());
        conn_sender.update_state().unwrap();
        assert_eq!(ConnectionState::Inviter(InviterState::Responded), conn_sender.get_state());

        consumer.activate().unwrap();
        thread::sleep(Duration::from_millis(500));
        conn_receiver.update_state().unwrap();
        assert_eq!(ConnectionState::Invitee(InviteeState::Completed), conn_receiver.get_state());

        institution.activate().unwrap();
        thread::sleep(Duration::from_millis(500));
        conn_sender.update_state().unwrap();
        assert_eq!(ConnectionState::Inviter(InviterState::Completed), conn_sender.get_state());
        ```
        
Not a big deal, consider some extraction - if it turns out problematic for some reason, or not very useful, nevermind.

Patrik-Stas (changes requested):
left a few comments

aries_vcx/tests/test_integration.rs
mirgee: Why? It is a perfectly valid input

aries_vcx/src/handlers/out_of_band/receiver/receiver.rs
mirgee: Sure this should be done, we discussed this scenario yesterday. There are a few potential problems:
* the `id` field of a service is [not necessarily a did](https://w3c.github.io/did-core/#services)
* the original connection should be updated before reusing it
* we do not distinguish between connections created via public and pairwise did (could be fixed by distinguishing between pw and public dids using a prefix - as we should)

I agree with potential problems incoming regarding the lifetimes. We require the returned connection to live as long as every connection in the vector, which is probably not very practical.

aries_vcx/tests/test_integration.rs
mirgee: Sure thing, that's the plan

aries_vcx/tests/test_integration.rs
mirgee: True (I did that just to generate a reasonable presentation request)

aries_vcx/tests/test_integration.rs
mirgee: That's right. Ideally, I think that testing the functionality of `connection_exists` specifically should be a job of a couple unit tests.

aries_vcx/tests/test_integration.rs
mirgee: Yeah, agreed - there quite a bit of duplication in the integration tests in general.

aries_vcx/src/handlers/out_of_band/receiver/receiver.rs
Patrik-Stas: > the original connection should be updated before reusing it

Assuming you maintain association `<public_did> : <vcx_connection_object>`, and you bump into the same public_did in some OOB message - you match it, doesn't mean you need to update service of connection's ddo.

> the bootstrap did doc does not contain public did currently and doing so would probably require assuming the id field of a service

That's right, `@id` can be any URI, not necessarily a DID. Quote from the spec you referred to: 
> The value of the id property MUST be a URI conforming to [RFC3986].

So I think need to remember public DID somewhere aside from DDO. I think on our Connection object level.

Patrik-Stas: > the original connection should be updated before reusing it

Assuming you maintain association `<public_did> : <vcx_connection_object>`, and you bump into the same public_did in some OOB message - you match it, doesn't mean you need to update service of connection's ddo.

> the bootstrap did doc does not contain public did currently and doing so would probably require assuming the id field of a service

That's right, `@id` can be any URI, not necessarily a DID. Quote from the spec you referred to: 
> The value of the id property MUST be a URI conforming to [RFC3986].

So I think need to remember public DID somewhere aside from DDO. I think on our Connection object level.

Patrik-Stas (approved):

PR #347, opened 2021-09-08
Add out of band message to nodejs wrapper + tests

mirgee:
#### OOB creator APIs
- `vcx_out_of_band_create` - Create new empty OOB protocol object.
- `vcx_out_of_band_append_message` - Given OOB protocol object, add message to `requests~attach` list of OOB message.
- `vcx_out_of_band_append_service` - Given OOB protocol object, add a service to `services` list of OOB message.
- `vcx_out_of_band_to_message` - Return full OOB message that should be shared with OOB recipient.
- `vcx_public_agent_get_service` - Given a public agent, return what could be used to setup [`service`](https://github.com/hyperledger/aries-rfcs/blob/main/features/0067-didcomm-diddoc-conventions/README.md#service-conventions) field on DDO. This can as input for `vcx_out_of_band_append_service`, so that OOB responses shall be collected by a particular public agent.

#### OOB receiver APIs:

- `vcx_out_of_band_create_from_message` - Create OOB protocol object using a received OOB message.
- `vcx_out_of_band_connection_exists` - Given OOB message and list of connections handles, determine a connection already exists with the OOB counterparty.
- `vcx_out_of_band_build_connection` - Given OOB message, build Connection protocol object which can be used to establish Aries connection with the counterparty.
- `vcx_out_of_band_extract_message` - Given OOB protocol object, return first valid Aries message found in `requests~attach`.

####  OOB shared methods
- `vcx_out_of_band_serialize` - Serialize OOB protocol object.
- `vcx_out_of_band_deserialize` - Deserialize OOB protocol object.
- `vcx_out_of_band_release` - Release OOB protocol object from memory.


aries_vcx/src/handlers/connection/public_agent.rs
```diff
@@ -43,6 +43,10 @@ impl PublicAgent {
         self.institution_did.clone()
     }
 
+    pub fn service(&self) -> VcxResult<FullService> {
```
Patrik-Stas, line 46: Not being used - luckly :-D Because it wouldn't make much sense to me. The service resolved from ledger might actually refer to a different, newer, public agent.

aries_vcx/src/handlers/out_of_band/mod.rs
```diff
@@ -43,8 +42,41 @@ pub struct OutOfBand {
...
+    }
+
+    pub fn from_string(oob_data: &str) -> VcxResult<OutOfBand> {
+        serde_json::from_str(oob_data)
+            .map_err(|err| VcxError::from_msg(VcxErrorKind::InvalidJson, format!("Cannot deserialize out of band message: {:?}", err)))
+    }
+}
+
+// TODO: Add more tests
+#[cfg(test)]
```
Patrik-Stas, line 64: We could add test to compare serialized OOB against expected string / serde JSON.

aries_vcx/src/handlers/connection/public_agent.rs
Patrik-Stas: Oh I see it's actually used from `libvcx`. 

if I get it right this should return service associated to the given public agent, but as pointed out above, this won't be the case if the data is resolved from ledger. The `self.institution_did` might have written arbitrary value for `service` field on ledger and it won't necessary match `self` public agent.

Patrik-Stas:
I'd suggest we actually create 2 kinds of OOB protocol objects, similar like we have 2 parties for each Aries protocol (invitee / inviter etc.). I've documented the C apis in the PR description as an exercise and realized the creator and receiver of OOB tend to use quite distict APIs, and calling methods not intended to by called by a particular party can cause undefined behaviour. I think we can save users from shooting themselves into foot by calling wrong method, on the wrong side of OOB.

Patrik-Stas:
Committed by mistake `aries_vcx/tails.txt/6GA72MbjhyEwgkb6xKi2HTWzWEEoQqUvgvEnzsH1Anu9`

libvcx/src/api_lib/api_handle/out_of_band.rs
```diff
@@ -0,0 +1,139 @@
...
+pub fn is_valid_handle(handle: u32) -> bool {
+    OUT_OF_BAND_MAP.has_handle(handle)
+}
+
+fn store_out_of_band(oob: OutOfBand) -> VcxResult<u32> {
+    OUT_OF_BAND_MAP.add(oob)
+        .or(Err(VcxError::from(VcxErrorKind::CreateOutOfBand)))
+}
+
+pub fn create_out_of_band(config: &str) -> VcxResult<u32> {
```
Patrik-Stas, line 30: Whould be nicer for config argument to be `OOBConfig`. That way looking at C method `vcx_out_of_band_create` it would also be easier to find out what its `config: *const c_char` is.

Patrik-Stas: This might be analogically applicable to other methods too. Move parsing string a layer up

wrappers/node/src/api/out-of-band.ts
```diff
@@ -0,0 +1,281 @@
...
+export enum GoalCode {
+  IssueVC = 'issue-vc',
+  RequestProof = 'request-proof',
+  CreateAccount = 'create-account',
+  P2PMessaging = 'p2p-messaging',
+}
+
+export enum HandshakeProtocol {
+  ConnectionV1 = 0,
+  DidCommV1 = 1,
```
Patrik-Stas, line 34: Probably DidExchange is what you meant here

Patrik-Stas (changes requested):
Generally looks good, left few comments + the suggestion about splitting oob interface for obb creator / receiver.

mirgee:
@Patrik-Stas What do you think about splitting the OOB interface in another PR?

aries_vcx/src/handlers/out_of_band/mod.rs
mirgee: Added a verification to the existing test.

aries_vcx/src/handlers/connection/public_agent.rs
mirgee: The agent's service is now implied from pairwise and cloud agent info instead from ledger now.

Patrik-Stas (approved):

PR #348, opened 2021-09-14
Expose thread id

mirgee:


aries_vcx/src/handlers/connection/invitee/states/complete.rs
```diff
@@ -15,26 +15,27 @@ pub struct CompleteState {
     pub did_doc: DidDoc,
     pub bootstrap_did_doc: DidDoc,
     pub protocols: Option<Vec<ProtocolDescriptor>>,
+    pub thread_id: String
```
Patrik-Stas, line 18: Please note in PR description that breaking changes to serialized structures are introduced

Patrik-Stas: In the future we should probably be versioning state machines / protocol objects and provide migration functions between them.

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -459,17 +458,38 @@ impl SmConnectionInviter {
     }
 
     pub fn handle_ack(self, ack: Ack) -> VcxResult<Self> {
-        let Self { source_id, pairwise_info, state, send_message } = self;
+        let Self { source_id, pairwise_info, state, send_message } = self.clone();
         let state = match state {
             InviterFullState::Responded(state) => {
-                InviterFullState::Completed((state, ack).into())
+                let thread_id = self.get_thread_id()?;
```
Patrik-Stas, line 464: Can't we do this in `update_state_with_message`? So we wouldn't have to this explicitly on every particular message handler?

aries_vcx/src/handlers/connection/util.rs
```diff
@@ -11,7 +11,7 @@ pub fn handle_ping(ping: &Ping,
 ) -> VcxResult<()> {
     if ping.response_requested {
         let ping_response = PingResponse::create().set_thread_id(
-            &ping.thread.as_ref().and_then(|thread| thread.thid.clone()).unwrap_or(ping.id.0.clone()));
+            &ping.thread.as_ref().and_then(|thread| Some(thread.thid.clone())).unwrap_or(ping.id.0.clone()));
```
Patrik-Stas, line 14: If you extract this statement into function like `build_ping_response(ping: &Ping) -> PingResponse`, it can be unit tested for both cases (when Ping contains `thread` and when it only contains `id`).

aries_vcx/src/messages/proof_presentation/presentation_request.rs
```diff
@@ -12,9 +12,8 @@ pub struct PresentationRequest {
     pub comment: Option<String>,
     #[serde(rename = "request_presentations~attach")]
     pub request_presentations_attach: Attachments,
-    #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "~thread")]
-    pub thread: Option<Thread>,
+    pub thread: Thread
```
Patrik-Stas, line 16: Since PresentationRequest is initial message within protocol (typically), does it have `~thread`? Isn't the value of thread determined by message `@id` once the request is received by Prover?

aries_vcx/src/messages/thread.rs
```diff
@@ -1,9 +1,8 @@
 use std::collections::HashMap;
 
-#[derive(Debug, Deserialize, Serialize, PartialEq, Clone)]
+#[derive(Debug, Deserialize, Serialize, PartialEq, Clone, Default)]
 pub struct Thread {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub thid: Option<String>,
+    pub thid: String,
```
Patrik-Stas, line 5: I suppose you already have reverted this locally

Patrik-Stas (changes requested):
Left a few comments

aries_vcx/src/messages/thread.rs
mirgee: Yes.

aries_vcx/src/handlers/connection/inviter/state_machine.rs
mirgee: In principle, we can do what you suggest, but we don't need the thread in every handler or state, so to me it makes sense to fetch it on demand (it returns result, so this way we don't return error when `get_thread_id()` fails for some reason yet we don't need to use thread id in the state). It also reduced the number of arguments to some of the handlers. But you are right it would maybe remove some lines.

aries_vcx/src/messages/proof_presentation/presentation_request.rs
mirgee: Good catch, I missed that. I also think presentation request should not have `~thread`.

aries_vcx/src/handlers/connection/util.rs
mirgee: Test added.

aries_vcx/src/handlers/connection/inviter/state_machine.rs
Patrik-Stas: Well right now we pretty much only check thread id in received message for `ack` message. Why? Whatever message we try to update the state machine with, it should have expected thread id. You should either add these checks to all other handles, or pull the checking on `update_state_with_message` (which I strongly favor). 

Equally there's forgotten cases to check for thread id on invitee side. For example we check `ack`'s thread_id in inviter side, but not on invitee's side.

aries_vcx/src/handlers/connection/invitee/state_machine.rs
```diff
@@ -400,6 +399,16 @@ impl SmConnectionInvitee {
     pub fn handle_ack(self, _ack: Ack) -> VcxResult<Self> {
         Ok(self)
     }
+
+    pub fn get_thread_id(&self) -> VcxResult<String> {
+        match &self.state {
+            InviteeFullState::Invited(_) => Err(VcxError::from_msg(VcxErrorKind::NotReady, "Thread ID not yet available in this state")),
+            InviteeFullState::Requested(state) => Ok(state.request.id.0.clone()),
+            InviteeFullState::Responded(state) => state.response.thread.thid.clone().ok_or(VcxError::from_msg(VcxErrorKind::UnknownError, "Thread ID missing on connection")),
```
Patrik-Stas, line 407: I understand this error handling arises from general optionality of thid by aries spec, but I think we should not allow this in a first place. Receiving connection response without thread id should not be valid. And if we allow it, we probably only delay issues (making them harder to debug), as for example, `get_thread_id`  will be failing (because we have previously allowed `connection response` without `thid`)

Patrik-Stas: We should have way to express that in some cases, `thid` is required. Maybe have 2 `Thread` structures, where one has `thid` required and other optional? Doesn't sound great but seems to me sometimes we need to be able to express that `thid` is required.

Patrik-Stas: I know we had this discussion before and to be honest I don't remember 100% of it - but would we break something if we just make `thid` required? Do we have case is `aries-vcx` implementation where we construct thread structure without `thid`?

aries_vcx/src/handlers/connection/inviter/state_machine.rs
Patrik-Stas: As per worries about calling `get_thread_id` and getting error, it should generally not happen if received messages are properly validated - this relates to my comments here https://github.com/hyperledger/aries-vcx/pull/348/files#r714713662

aries_vcx/src/handlers/issuance/issuer/state_machine.rs
```diff
@@ -369,23 +372,28 @@ pub mod test {
...
 
-    fn _tails_file() -> String {
+    pub fn _tails_file() -> String {
         String::from("TEST_TAILS_FILE")
     }
 
-    fn _send_message() -> Option<&'static impl Fn(&A2AMessage) -> VcxResult<()>> {
-        Some(&|_: &A2AMessage| VcxResult::Ok(()))
+    pub fn _send_message() -> Option<&'static impl Fn(&A2AMessage) -> VcxResult<()>> {
+        Some(&|_: &A2AMessage| send_message("", &DidDoc::default(), &A2AMessage::BasicMessage(BasicMessage::default())))
```
Patrik-Stas, line 396: I am surprised tests are not failing after this change - this will actually try to call networking code. Why making this change?

Patrik-Stas (changes requested):
left comments

aries_vcx/src/handlers/issuance/issuer/state_machine.rs
Patrik-Stas: I think it would be cleaner to just make new method, something like `_send_message_but_fail` and return `Error`. 
Reasons:
1. It would be easier to read the test, imo
2. This adds dependency on agency code into state machine - I know such dependencies exists today, but we should head toward reducing them

Patrik-Stas:
I went through all protocols and found few places where verification is missing

invitee, `handle_connection_response`
inviter, `handle_connection_request`
holder, code handling receiving credential
issuer, code handling credential request

Additionally I suggest we remove `handle_ack` on invitee side, as you pointed out, invitee does not receive acks within connection protocol, yet it has `handle_ack` which is also missing `thid` verification. But why add it if it's never used.

Patrik-Stas:
The above is conclusion of my finding but for record, I'll share here my private notes while I was going through this:
```
### invitee
- handle_invitation          ( can N, is N )
- handle_connection_response ( can Y, is N ) << TODO: add
- handle_ack                 ( can Y, is N ) << TODO: I suppose we should just remove handle_ack capability for invitee side of connection protocol

- handle_problem_report      ( can N, is N ) // (is using pthid to refer @id if whatever problematic message)
- handle_ping                ( can N, is N ) // (ping has only @id, ping response useds that for ~thread.thid)
- handle_ping_response       ( can Y, is N ) << todo later: abstract the protocol out from connection, then we can handle threads (thid == @id of ping)
- handle_discovery_query     ( can N, is N ) // (has @id, initiates thread)
- handle_disclose            ( can Y, is N ) << todo later: abstract the protocol out from connection, then we can handle threads (thid == @id of discovery_query)

### inviter
- handle_connection_request  ( can Y, is N ) << TODO: thid should be same as @id of invitation
- handle_ack                 ( can Y, is Y ) 
- handle_problem_report      ( can N, is N ) 

- handle_ping                ( can N, is N ) // (ping has only @id, ping response useds that for ~thread.thid)
- handle_ping_response       ( can Y, is N ) << todo later: abstract the protocol out from connection, then we can handle threads (thid == @id of ping)
- handle_discovery_query     ( can N, is N ) // (has @id, initiates thread)
- handle_disclose            ( can Y, is N ) << todo later: abstract the protocol out from connection, then we can handle threads (thid == @id of discovery_query)


----------------------------------------------------------------

### holder
- handle credential offer ( can N, is N ) (assuming the interaction start from issuer, @id begins the thread)
- handle credential       ( can Y, is N ) << add (thid == @id of offer)


### issuer
- handle cred request ( can Y, is N ) << todo: add

----------------------------------------------------------------

### prover
- handle proof request ( can N, is N ) 

### verifier
- handle presentation (can Y, is Y) 
```

Patrik-Stas:
As far as I can tell there's also no verifications on stuff like `handle_ping`,  `handle_ping_response ` but I'd just forget that for now. I was thinking we should *maybe* extract those protocols as separate protocol objects, just like we have issuance, presentation or oob. But that's different discussion

aries_vcx/src/handlers/issuance/issuer/state_machine.rs
mirgee: I like the flexibility of mocking responses, but I did as you suggested.

aries_vcx/src/handlers/connection/invitee/state_machine.rs
mirgee: I don't understand why `thid` being optional is a problem as long as, if it is expected, it is verified immediately after receiving the message, which it is.

Patrik-Stas: It allows for scenarios which should never happen - in messages and scenarios where `thid` is required (basically almost always at normal circumstances, except for initial messages or problems reports), typing `thid` as Option does not reflect the requirement and delegates the responsibility to check value of `Option` to the developer anytime you try to access this data. Just like in the case above:

```
InviteeFullState::Responded(state) => state.response.thread.thid.clone().ok_or(VcxError::from_msg(VcxErrorKind::UnknownError, "Thread ID missing on connection")),
```

we need to throw `Uknown Error` because we don't really know what happened - we as developer must have made mistake if this happens - we know we are checking `Connection Response` for `thid` anytime we use it to update state.   If `thid` in this case was not optional, either:
1. it has `thid`, we continue and we never have to worry about throwing `VcxErrorKind::UnknownError` elsewhere in code
2. it does not have `thid` and we fail quickly. 

I mean - at core making a field mandatory if it is mandatory just reflects reality better and that itself then have positive effects on code cleaniness. 

The threading RFC states that `thid` is optional but most of the time it's not. I am not sure what'd be the best technical solution - the simplest I can think of is having 2 `Thread` structures, one with `thid` optional, the other not - perhaps `GeneralThread` and `ConversationThread`. You would use `Conversation Thread` on messages which are expected to be part of a conversation (must have `thid`) and `GeneralThread` in cases when AriesRFC-wide flexibility is required.

Patrik-Stas (dismissed):

PR #350, opened 2021-09-21
Separate sender and receiver sides in OOB protocol

mirgee:


Patrik-Stas:
@mirgee some conflicts popped up here

Patrik-Stas:
After merging the original oob wrapper, this will need some update

aries_vcx/src/handlers/out_of_band/sender/sender.rs
```diff
@@ -1,53 +1,58 @@
...
     pub fn set_goal(mut self, goal: &str) -> Self {
-        self.goal = Some(goal.to_string());
+        self.oob.goal = Some(goal.to_string());
         self
     }
 
-    pub fn append_service(mut self, service: ServiceResolvable) -> Self {
-        self.services.push(service);
-        self
+    pub fn append_service(&mut self, service: &ServiceResolvable) -> VcxResult<()> {
```
Patrik-Stas, line 34: I think we can just return nothing, or Self? There's no error scenario here

aries_vcx/src/handlers/out_of_band/sender/sender.rs
```diff
@@ -1,53 +1,58 @@
...
+        self.oob.services.push(service.clone());
+        Ok(())
     }
 
-    pub fn append_handshake_protocol(mut self, protocol: HandshakeProtocol) -> VcxResult<Self> {
+    pub fn append_handshake_protocol(&mut self, protocol: &HandshakeProtocol) -> VcxResult<()> {
```
Patrik-Stas, line 39: also don't see err scenario here

libvcx/src/api_lib/api_c/out_of_band.rs
```diff
@@ -0,0 +1,450 @@
...
+use aries_vcx::indy_sys::CommandHandle;
+
+use crate::api_lib::api_handle::out_of_band;
+use crate::api_lib::utils::cstring::CStringUtils;
+use crate::api_lib::utils::runtime::execute;
+use crate::error::prelude::*;
+use aries_vcx::utils::error;
+
+#[no_mangle]
+pub extern fn vcx_out_of_band_create(command_handle: CommandHandle,
```
Patrik-Stas, line 13: I'd suggest `vcx_out_of_band_sender_create`

libvcx/src/api_lib/api_c/out_of_band.rs
```diff
@@ -0,0 +1,450 @@
...
+            }
+        }
+        Ok(())
+    });
+
+    error::SUCCESS.code_num
+}
+
+#[no_mangle]
+pub extern fn vcx_out_of_band_create_from_message(command_handle: CommandHandle,
```
Patrik-Stas, line 43: I'd suggest `vcx_out_of_band_receiver_create_from_message` or even simply `vcx_out_of_band_receiver_create`

libvcx/src/api_lib/api_c/out_of_band.rs
```diff
@@ -0,0 +1,450 @@
...
+            }
+        }
+        Ok(())
+    });
+
+    error::SUCCESS.code_num
+}
+
+#[no_mangle]
+pub extern fn vcx_out_of_band_append_message(command_handle: CommandHandle,
```
Patrik-Stas, line 73: same as above, also applies for next couple methods, I'd distinguish them with `_sender_` and `_receiver_` unless shared for both

Patrik-Stas (changes requested):
only minor changes requested

aries_vcx/src/handlers/out_of_band/sender/sender.rs
mirgee: I like to keep flexibility to add result-returning methods later without necessary changes rippling through every usage of the method.

mirgee:
I just realized fixing some of the conflicts will be a bit more complicated than expected, will have to revert some of the commits

aries_vcx/src/handlers/out_of_band/sender/sender.rs
Patrik-Stas: Uhhh I don't know, by that logic we should return Result everywhere. About the flexibility - I see that as a minus, if you actually add error result into `append_service `, it will compile straight away and you might think you are done, skipping review of how the particular error should be handled. If instead you had to change return type of the `append_service` from `Self` to `Result`, you would also be force to better think about the consequences wherever `append_service ` is called.

Patrik-Stas (dismissed):

aries_vcx/src/handlers/out_of_band/sender/sender.rs
mirgee: Resolved :)

Patrik-Stas (approved):

PR #375, opened 2021-10-29
Store thread id on state machine level for connection

mirgee:


aries_vcx/src/handlers/connection/connection.rs
```diff
@@ -804,15 +772,15 @@ impl<'de> Deserialize<'de> for Connection {
     }
 }
 
-impl Into<(SmConnectionState, PairwiseInfo, CloudAgentInfo, String)> for Connection {
-    fn into(self) -> (SmConnectionState, PairwiseInfo, CloudAgentInfo, String) {
-        (self.state_object(), self.pairwise_info().to_owned(), self.cloud_agent_info().to_owned(), self.source_id())
+impl Into<(SmConnectionState, PairwiseInfo, CloudAgentInfo, String, String)> for Connection {
+    fn into(self) -> (SmConnectionState, PairwiseInfo, CloudAgentInfo, String, String) {
+        (self.state_object(), self.pairwise_info().to_owned(), self.cloud_agent_info().to_owned(), self.source_id(), self.get_thread_id().unwrap()) // TODO: Use TryInto instead
```
Patrik-Stas, line 777: `get_thread_id()` always returns `Ok()` - why not to just remove VcxResult wrapper from the return value?

aries_vcx/src/handlers/connection/invitee/state_machine.rs
```diff
@@ -246,106 +244,88 @@ impl SmConnectionInvitee {
...
-                state.clone()
-            }
+            _ => self.state.clone()
         };
-        Ok(Self { state, pairwise_info, source_id, ..self })
+        Ok(Self { state, ..self })
     }
 
     pub fn handle_connection_response(self, response: SignedResponse) -> VcxResult<Self> {
-        verify_thread_id(&self.get_thread_id()?, &A2AMessage::ConnectionResponse(response.clone()))?;
```
Patrik-Stas, line 278: I think we should be doing this check, why remove?

aries_vcx/src/handlers/connection/invitee/state_machine.rs
```diff
@@ -354,46 +334,40 @@ impl SmConnectionInvitee {
...
+                    Ok(response) => InviteeFullState::Completed((state.clone(), response).into()),
                     Err(err) => {
                         let problem_report = ProblemReport::create()
                             .set_problem_code(ProblemCode::ResponseProcessingError)
                             .set_explain(err.to_string())
-                            .set_thread_id(&state.request.id.0);
-                        send_message(&pairwise_info.pw_vk, &state.did_doc, &problem_report.to_a2a_message()).ok();
-                        InviteeFullState::Initial((state, problem_report).into())
+                            .set_thread_id(&self.thread_id);
+                        (self.send_message)(&self.pairwise_info.pw_vk, &state.did_doc, &problem_report.to_a2a_message()).ok();
```
Patrik-Stas, line 352: nitpick - can't be without the parenthesis around `self.send_message`?

aries_vcx/src/handlers/connection/invitee/state_machine.rs
```diff
@@ -504,27 +472,6 @@ pub mod test {
             }
         }
 
-        mod get_thread_id {
-            use super::*;
-
-            #[test]
-            #[cfg(feature = "general_test")]
-            fn handle_response_fails_with_incorrect_thread_id() {
```
Patrik-Stas, line 512: I think this test should still be valid/working? If we try to process connection response with wrong thread id, we should detect it

aries_vcx/src/handlers/connection/invitee/state_machine.rs
mirgee: No (complaints please send to the Rust compiler :))

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -240,38 +243,36 @@ impl SmConnectionInviter {
...
 
     pub fn handle_connection_request(self,
                                      request: Request,
                                      new_pairwise_info: &PairwiseInfo,
                                      new_routing_keys: Vec<String>,
                                      new_service_endpoint: String) -> VcxResult<Self> {
-        let Self { pairwise_info: bootstrap_pairwise_info, state, send_message, .. } = self.clone();
-        let state = match state {
+        let bootstrap_pairwise_info = self.pairwise_info.clone();
+        let thread_id = request.get_thread_id().ok_or(VcxError::from_msg(VcxErrorKind::InvalidJson, "Missing ~thread decorator field in request"))?;
```
Patrik-Stas, line 274: Also I think we miss matching the thread id from the request against our known `self.thread_id`?

aries_vcx/src/handlers/connection/invitee/state_machine.rs
mirgee: We are doing them multiple times - once in `update_state` and once in the handlers I removed the from.

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -314,164 +313,119 @@ impl SmConnectionInviter {
...
-            InviterFullState::Invited(state) => state.invitation.get_id(), 
-            InviterFullState::Requested(state) => Ok(state.thread_id.clone()),
-            InviterFullState::Responded(state) => state.signed_response.thread.thid.clone().ok_or(VcxError::from_msg(VcxErrorKind::UnknownError, "Thread ID missing on connection")),
-            InviterFullState::Completed(state) => state.thread_id.clone().ok_or(VcxError::from_msg(VcxErrorKind::UnknownError, "Thread ID missing on connection")),
-            InviterFullState::Initial(_) => Ok(String::new())
-        }
+        Ok(self.thread_id.clone())
     }
 
     fn build_response(
```
Patrik-Stas, line 431: We are verifying thread id in request conditionally against our invite - instead we can now just verify `self.thread_id` against the thread_id in the request

aries_vcx/src/handlers/connection/inviter/state_machine.rs
Patrik-Stas: Oh I see we do that in `build_response` - I left a comment there and consquently if we always do thread verification inside `build_response`, we don't have to check if request contains `thread_id` at this place

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -314,164 +313,119 @@ impl SmConnectionInviter {
...
-                    send_message(&pairwise_info.pw_vk, &state.did_doc, &problem_report.to_a2a_message()).ok();
-                    InviterFullState::Initial((state, problem_report).into())
-                } else {
-                    InviterFullState::Completed((state, ack).into())
-                }
-            }
-            _ => {
-                state.clone()
-            }
+        let state = match self.state {
```
Patrik-Stas, line 420: we should still do `thread_id` verification, but the code should look simpler now I believe

aries_vcx/src/handlers/connection/invitee/state_machine.rs
mirgee: As I said in another comment, I disagree with verifying thread id multiple times or in every handler.

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -493,15 +447,14 @@ impl SmConnectionInviter {
             .set_service_endpoint(new_service_endpoint)
             .set_keys(new_recipient_keys, new_routing_keys)
             .ask_for_ack()
-            .set_thread_id(&request.id.0)
+            .set_thread_id(&request.get_thread_id().ok_or(VcxError::from_msg(VcxErrorKind::InvalidJson, "Missing ~thread decorator field in request"))?)
```
Patrik-Stas, line 450: I don't quite remember - can Connection Request have non-set thread?

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -562,28 +515,6 @@ pub mod test {
             }
         }
 
-        mod get_thread_id {
-            use super::*;
-
-            #[test]
-            #[cfg(feature = "general_test")]
-            fn ack_fails_with_incorrect_thread_id() {
```
Patrik-Stas, line 570: I think this is still valid test

aries_vcx/src/handlers/connection/inviter/state_machine.rs
mirgee: It is marked as TODO to make it mandatory.

aries_vcx/src/handlers/connection/inviter/state_machine.rs
mirgee: dtto

ci/libvcx-codecov.dockerfile
```diff
@@ -67,4 +67,4 @@ RUN mv /home/indy/indy-sdk/libnullpay/target/release/*.so \
 
 USER indy
 
-RUN cargo install grcov
+RUN cargo install grcov --version 0.8.3
```
Patrik-Stas, line 70: 👍

Patrik-Stas (changes requested):
left some comments

aries_vcx/src/handlers/connection/invitee/state_machine.rs
Patrik-Stas: Oh yeah! I missed that. However I think it should actually be in `state_machine.rs` - even if it means that it needs to be placed at multiple `handle_msg_<xyz>` functions. The reason being - `thread_id` is owned by the connection state machine itself and should happen regardless of how the upper abstraction layer (`connection.rs`, or perhaps it's agency-less equivalent in future) is implemented.

aries_vcx/src/handlers/connection/connection.rs
mirgee: However, currently, invitee's SM has `thread_id` still undefined on creation - we should eventually create invitee in invited state (replace Invited state with Initial state with invitation and therefore thread id defined).

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -314,164 +313,119 @@ impl SmConnectionInviter {
...
 
     pub fn get_thread_id(&self) -> VcxResult<String> {
-        match &self.state {
-            InviterFullState::Invited(state) => state.invitation.get_id(), 
-            InviterFullState::Requested(state) => Ok(state.thread_id.clone()),
-            InviterFullState::Responded(state) => state.signed_response.thread.thid.clone().ok_or(VcxError::from_msg(VcxErrorKind::UnknownError, "Thread ID missing on connection")),
-            InviterFullState::Completed(state) => state.thread_id.clone().ok_or(VcxError::from_msg(VcxErrorKind::UnknownError, "Thread ID missing on connection")),
-            InviterFullState::Initial(_) => Ok(String::new())
-        }
+        Ok(self.thread_id.clone())
```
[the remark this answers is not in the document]
mirgee: dtto

aries_vcx/src/handlers/connection/inviter/state_machine.rs
mirgee: I agree it seems confusing, maybe we should consolidate `handle_connection_request` and `handle_send_response` (i.e. `Requested` and `Responded` states) into one step.

Patrik-Stas: Yeah I think that's a good idea, that way connection response can be constructed arbitrarily on upper layers without forcing a particular way on SM layer

Patrik-Stas: But back to this particular line of code - especially with the changes you did in `build_response`, I think this line can be completely removed - if `thread_id` is missing in request, it will be detected in `build_response` and problem report will be sent (so in that way it's even more desired behaviour, doesn't happen with line `274`)

Patrik-Stas (approved):

PR #376, opened 2021-11-02
Simplify issuer offer sending API

mirgee:
Before, credential offer was sent in e.g. NodeJS as follows:
```
    const issuerCred = await IssuerCredential.create({
      attr,
      sourceId: 'alice_degree',
      credDefHandle,
      credentialName: 'cred',
      price: '0',
      issuerDid
    })
    await issuerCred.sendOffer(connection)
```
In order to simplify the API (and the internal code), this was changed to (again, using NodeJS only as an example):
```
    const issuerCred = await IssuerCredential.create('alice_degree')
    logger.info(`Per issuer credential ${issuerCredId}, sending cred offer to connection ${connectionId}`)
    await issuerCred.sendOffer(connection, {
      attr,
      credDefHandle
    })

```


wrappers/node/test/suite1/ariesvcx-issuer-credential.test.ts
```diff
@@ -18,54 +18,16 @@ describe('IssuerCredential:', () => {
...
-      const { sourceId, ...data } = await dataIssuerCredentialCreate();
-      const error = await shouldThrow(() => IssuerCredential.create(data as any));
-      assert.equal(error.vcxCode, VCXCode.INVALID_OPTION);
-    });
-
-    it('throws: invalid credDefHandle', async () => {
-      const { credDefHandle, ...data } = await dataIssuerCredentialCreate();
-      const error = await shouldThrow(() => IssuerCredential.create(data as any));
-      assert.equal(error.vcxCode, VCXCode.INVALID_CREDENTIAL_DEF_HANDLE);
-    });
```
mirgee, line 30: Can add similar test for the send offer. Do we want to maintain nodejs unit tests?

wrappers/node/src/api/issuer-credential.ts
```diff
@@ -216,14 +188,16 @@ export class IssuerCredential extends VCXBaseWithState<IIssuerCredentialData, Is
    * await issuerCredential.sendOffer(connection)
    * ```
    */
-  public async sendOffer(connection: Connection): Promise<void> {
+  public async sendOffer(connection: Connection, data: IIssuerCredentialOfferSendData): Promise<void> {
```
mirgee, line 191: Perhaps the argument should be a single object of type
```
export interface IIssuerCredentialOfferSendData {
  credDef: CredentialDef;
  connection: Connection;
  attr: {
    [index: string]: string;
  };
}
```

aries_vcx/src/handlers/issuance/issuer/state_machine.rs
```diff
@@ -264,48 +222,38 @@ impl IssuerSM {
...
             }
             IssuerFullState::ProposalReceived(state_data) => match cim {
-                CredentialIssuanceMessage::CredentialOfferSend(comment) => {
-                    let (cred_def_id, credential_json) = match state_data.offer_info {
-                        None => return Err(VcxError::from_msg(VcxErrorKind::InvalidState, "`set_offer()` must be called before sending credential offer after receiving proposal")),
-                        Some(offer_info) => (offer_info.cred_def_id, offer_info.credential_json)
-                    };
-                    let cred_offer = libindy_issuer_create_credential_offer(&cred_def_id)?;
+                CredentialIssuanceMessage::CredentialOfferSend(offer_info, comment) => {
+                    let cred_offer = libindy_issuer_create_credential_offer(&offer_info.cred_def_id)?;
```
Patrik-Stas, line 246: Maybe in separate PR, i'd suggest this line, and line `248` a layer up, so `CredentialOfferSend` would actually be responsible for passing in `CredentialOffer` - hence remove the dependency on `libindy/vdr-tools` here and talk in pure aries terms (eg. "please pass me valid CredentialOffer message, I don't care if it's based on libindy")

mirgee: Agree, let's do it (in a separate PR, though)

Patrik-Stas (approved):

PR #378, opened 2021-11-05
Make thread id optional

mirgee:

### DIFF
--- a/aries_vcx/src/handlers/connection/invitee/state_machine.rs
+++ b/aries_vcx/src/handlers/connection/invitee/state_machine.rs
@@ -231,12 +231,12 @@ impl SmConnectionInvitee {
                  pairwise_info: &PairwiseInfo,
                  send_message: fn(&str, &DidDoc, &A2AMessage) -> VcxResult<()>) -> VcxResult<Response> {
         let remote_vk: String = did_doc.recipient_keys().get(0).cloned()
-            .ok_or(VcxError::from_msg(VcxErrorKind::InvalidState, "Cannot handle Response: Remote Verkey not found"))?;
+            .ok_or(VcxError::from_msg(VcxErrorKind::InvalidState, "Cannot handle response: remote verkey not found"))?;
 
         let response = response.clone().decode(&remote_vk)?;
 
         if !response.from_thread(&request.get_thread_id()) {
-            return Err(VcxError::from_msg(VcxErrorKind::InvalidJson, format!("Cannot handle Response: thread id does not match: {:?}", response.thread)));
+            return Err(VcxError::from_msg(VcxErrorKind::InvalidJson, format!("Cannot handle response: thread id does not match: {:?}", response.thread)));
         }
 
         let message = Ack::create()
--- a/aries_vcx/src/handlers/connection/invitee/state_machine.rs
+++ b/aries_vcx/src/handlers/connection/invitee/state_machine.rs
@@ -33,7 +33,7 @@ pub struct SmConnectionInvitee {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub enum InviteeFullState {
     Initial(InitialState),
-    Invited(InvitedState), // TODO: Remove this state
+    Invited(InvitedState),
     Requested(RequestedState),
     Responded(RespondedState),
     Completed(CompleteState),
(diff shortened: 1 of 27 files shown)


aries_vcx/src/handlers/connection/invitee/state_machine.rs
```diff
@@ -369,13 +368,13 @@ impl SmConnectionInvitee {
         Ok(Self { state, ..self })
     }
 
-    pub fn handle_problem_report(self, problem_report: ProblemReport) -> VcxResult<Self> {
+    pub fn handle_problem_report(self, _problem_report: ProblemReport) -> VcxResult<Self> {
```
Patrik-Stas, line 371: I was about to say I strongly oppose not storing problem report after it's received - only to find out we didn't do it previously either and just ignored it in `impl From<(InvitedState, ProblemReport)> for InitialState {` :-) 

So it's good that this modification pointed that out - I think we should store it, what do you think?

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -451,15 +450,14 @@ impl SmConnectionInviter {
         new_service_endpoint: String,
     ) -> VcxResult<SignedResponse> {
         request.connection.did_doc.validate()?;
-        let thread_id = &request.get_thread_id().ok_or(VcxError::from_msg(VcxErrorKind::InvalidJson, "Missing ~thread decorator field in request"))?;
-        verify_thread_id(&thread_id, &A2AMessage::ConnectionRequest(request.clone()))?;
+        verify_thread_id(&request.get_thread_id(), &A2AMessage::ConnectionRequest(request.clone()))?;
```
Patrik-Stas, line 453: Seems like we are validating request thread_id against itself, instead of `self.thread_id`? I think this is verification will always pass

aries_vcx/src/handlers/connection/inviter/state_machine.rs
```diff
@@ -451,15 +450,14 @@ impl SmConnectionInviter {
...
         let new_recipient_keys = vec!(new_pairwise_info.pw_vk.clone());
         Response::create()
             .set_did(new_pairwise_info.pw_did.to_string())
             .set_service_endpoint(new_service_endpoint)
             .set_keys(new_recipient_keys, new_routing_keys)
             .ask_for_ack()
-            .set_thread_id(&thread_id)
+            .set_thread_id(&request.get_thread_id())
```
Patrik-Stas, line 460: Would feel safer to always copy from `self.thread_id`, though after verification those values should be the same (except when the verification of the message thread id is forgotten or incorrect like happened above)

aries_vcx/src/messages/connection/request.rs
```diff
@@ -9,6 +9,7 @@ pub struct Request {
     pub label: String,
     pub connection: ConnectionData,
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "~thread")]
```
Patrik-Stas, line 12: 😮  surprised integration with acapy been working

Patrik-Stas (changes requested):
few comments

aries_vcx/src/handlers/connection/inviter/state_machine.rs
mirgee: Of course, that's a typo.

aries_vcx/src/handlers/connection/invitee/state_machine.rs
mirgee: I am not so sure. Maybe we can start storing it when we have a use case for it?

aries_vcx/src/handlers/connection/inviter/state_machine.rs
mirgee: I don't think it should matter at all, but will make it `self.thread_id` if you think it's better.

aries_vcx/src/messages/connection/request.rs
mirgee: Me too.

aries_vcx/src/handlers/connection/inviter/state_machine.rs
mirgee: Actually, `self.thread_id` can't be used here because it is undefined in the scenario where the inviter is creating a connection from received request (the thread id is set after the response is built). There are ways to satisfy your demand to use `self.thread_id`, but I don't see the upside of that.

Patrik-Stas (approved):

PR #383, opened 2021-11-22
Do not duplicate openssl and indy symbols in libvcx ios libraries

Patrik-Stas:
- Previously some symbols were probably duplicated in final binary, because openssl (libssl, libcrypto) and indy (libindy) were first statically linked to `libvcx.a` during libvcx build and then secondly these libraries were again combined into single `.a` in step `copy_libvcx_architectures` of ios build. Hence I preserved `libvcx` build process, but removed combining openssl, indy symbols in ios build.
- I removed `stripping` phase in ios build which was never utilized. This was condition on `DEBUG_SYMBOLS` variable which we never set. This step is actually useful because it decreases size of final binary by about 20-25%, but I'll rather try to introduce this feature in separate PR. In this PR I am basically just removing it as dead code.
- I added few logs to iOS CI
- Updated git ignore file

### DIFF
--- a/wrappers/ios/ci/build.sh
+++ b/wrappers/ios/ci/build.sh
@@ -212,11 +212,8 @@ copy_libvcx_architectures() {
 copy_libs_to_combine() {
     mkdir -p $OUTPUT_DIR/cache/arch_libs
 
-    copy_lib_tocombine openssl libssl
-    copy_lib_tocombine openssl libcrypto
     copy_lib_tocombine sodium libsodium
     copy_lib_tocombine zmq libzmq
-    copy_lib_tocombine indy libindy
     copy_lib_tocombine vcx libvcx
 }
 
@@ -227,7 +224,7 @@ copy_lib_tocombine() {
     ARCHS="arm64 x86_64"
 
     for ARCH in ${ARCHS[*]}; do
-        cp -v $OUTPUT_DIR/libs/$LIB_NAME/$ARCH/$LIB_FILE_NAME.a $OUTPUT_DIR/cache/arch_libs/${LIB_FILE_NAME}_$ARCH.a
+        cp -v "$OUTPUT_DIR/libs/$LIB_NAME/$ARCH/$LIB_FILE_NAME.a" "$OUTPUT_DIR/cache/arch_libs/${LIB_FILE_NAME}_$ARCH.a"
     done
 }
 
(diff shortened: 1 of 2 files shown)


wrappers/ios/ci/build.sh
```diff
@@ -405,7 +368,7 @@ abspath() {
 
 # Setup environment
 setup
- 
+
 # Build 3rd party libraries
 build_crypto
```
mirgee, line 373: Is this step still necessary? When is the iOS build of openssl used? Does it replace openssl installed via brew?

wrappers/ios/ci/build.sh
```diff
@@ -227,79 +225,44 @@ copy_lib_tocombine() {
...
     done
 
-    echo "Using source_combined: ${source_combined} to create ${COMBINED_LIB}.a"
-    # Merge the combined library for each architecture into a single fat binary
-    lipo -create $source_combined -o $OUTPUT_DIR/${COMBINED_LIB}.a
+    for arch in ${ARCHS[*]}; do
+        COMBINED_LIB_PATH=${BUILD_CACHE}/arch_libs/${COMBINED_LIB}_${arch}.a
+        echo "Lipo info about combined library ${COMBINED_LIB_PATH}:"
+        lipo -info "$COMBINED_LIB_PATH"
+    done
```
mirgee, line 258: I really like how much unnecessary code is removed, it makes this piece much more readable. However, I think it would be even more readable if we remove code useful primarily for debugging, like the usages of `lipo -info ...` :)

wrappers/ios/ci/build.sh
```diff
@@ -227,79 +225,44 @@ copy_lib_tocombine() {
...
-    rm -rf ${source_combined}
+    FAT_COMBINED_LIB_PATH="$OUTPUT_DIR/${COMBINED_LIB}.a"
+    echo "Using combined_libs_paths: ${combined_libs_paths} to combine them into single fat library: ${FAT_COMBINED_LIB_PATH}"
+    lipo -create ${combined_libs_paths} -o "${FAT_COMBINED_LIB_PATH}"
 
-    # Show info on the output library as confirmation
-    echo "Combination complete."
-    lipo -info $OUTPUT_DIR/${COMBINED_LIB}.a
+    echo "Lipo info about combined library ${FAT_COMBINED_LIB_PATH}:"
+    lipo -info "${FAT_COMBINED_LIB_PATH}"
```
mirgee, line 265: dtto

wrappers/ios/ci/build.sh
```diff
@@ -427,7 +390,10 @@ build_libvcx
 copy_libvcx_architectures
 
 # Copy libraries to combine
+
+tree "$OUTPUT_DIR/libs/"
 copy_libs_to_combine
+tree "$OUTPUT_DIR/cache/arch_libs"
```
mirgee, line 396: dtto

mirgee (changes requested):

wrappers/ios/ci/build.sh
```diff
@@ -227,79 +225,44 @@ copy_lib_tocombine() {
     ARCHS="arm64 x86_64"
 
     for ARCH in ${ARCHS[*]}; do
-        cp -v $OUTPUT_DIR/libs/$LIB_NAME/$ARCH/$LIB_FILE_NAME.a $OUTPUT_DIR/cache/arch_libs/${LIB_FILE_NAME}_$ARCH.a
+        cp -v "$OUTPUT_DIR/libs/$LIB_NAME/$ARCH/$LIB_FILE_NAME.a" "$OUTPUT_DIR/cache/arch_libs/${LIB_FILE_NAME}_$ARCH.a"
```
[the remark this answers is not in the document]
Patrik-Stas: It's best practice to always enclose strings containing variable interpolation into double quotes, otherwise it can cause undesired behaviour if the variable contains whitespace (what you expected to be 1 literal token might become 2 or more separate tokens). That's most likely won't be the case here even in the future, but it is best practice.

wrappers/ios/ci/build.sh
Patrik-Stas: This step is required by `build_libvcx`, because `libvcx/build.rs` requires variable `OPENSSL_LIB_DIR` to be passed.

.github/workflows/main.yml
```diff
@@ -855,7 +855,7 @@ jobs:
 
   # TODO: Add tests of iOS build
   publish-ios-wrapper:
-    needs: [workflow-setup, build-image-libvcx]
+    needs: [workflow-setup]
```
mirgee, line 858: Braces are unnecessary here.

wrappers/ios/ci/build.sh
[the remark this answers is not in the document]
mirgee: OK

wrappers/ios/ci/build.sh
```diff
@@ -50,6 +50,7 @@ setup() {
     brew list openssl &>/dev/null || brew install openssl@1.1
     brew list zmq &>/dev/null || brew install zmq
     brew list libzip &>/dev/null || brew install libzip
+    brew list tree &>/dev/null || brew install tree
```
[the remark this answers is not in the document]
Patrik-Stas: removed

wrappers/ios/ci/build.sh
Patrik-Stas: removed

PR #398, opened 2021-11-27
iOS Wrapper abstraction

kukgini:
Some information, such as handler types (u32, i32) or callback definitions, is only useful within vcx and the client code that uses them doesn't need to know the specifics. So I think hiding it from the client code will improve maintainability.

Signed-off-by: kukgini <kukgini@gmail.com>


kukgini:
I will close this PR and try again to fix the DCO issue.

PR #399, opened 2021-11-29
Ugprade/reqwest 0.10.10

Patrik-Stas:


Patrik-Stas:
Runs fine on ios, had reported problem with error from reqwest `Cannot drop a runtime in a context where blocking is not allowed.`

PR #403, opened 2021-11-29
Revert reqwest, futures, tokio upgrade (revision 5502176e6)

Patrik-Stas:


.github/workflows/main.yml
```diff
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - main
+      - master
```
mirgee, line 6: Incorrect branch

agency_client/Cargo.toml
```diff
@@ -16,7 +16,7 @@ lazy_static = "1.3"
 serde = "1.0.97"
 serde_json = "1.0.40"
 serde_derive = "1.0.97"
-reqwest = { version = "0.9.5" }
+reqwest = "0.9.5"
```
mirgee, line 19: Doesn't seem to revert reqwest version.

agents/node/vcxagent-core/package.json
```diff
@@ -69,6 +69,6 @@
     "winston": "^3.3.3"
   },
   "peerDependencies": {
-    "@hyperledger/node-vcx-wrapper": "^0.24.1"
+    "@hyperledger/node-vcx-wrapper": "^0.24.2"
```
mirgee, line 72: I would prefer to have release PR separate as we usually do.

agency_client/Cargo.toml
Patrik-Stas: It doesn't revert reqwest specifically, but it does revert dependencies which previous upgrade of reqwest has caused bringing in in lock file. When the reqwest was previously downgraded from 11 to 10, then from 10 to 9, in neither of these downgrades we reverted dependencies brought in by the first `9->11` upgrade.

mirgee (approved):

PR #408, opened 2021-12-02
Use GITHUB_TOKEN, GITHUB_ACTOR in ci

Patrik-Stas:
Use GITHUB_TOKEN, GITHUB_ACTOR to publish libvcx image to github registry

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1015,8 +1015,8 @@ jobs:
         uses: azure/docker-login@v1
         with:
           login-server: ghcr.io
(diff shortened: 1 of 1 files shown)


Patrik-Stas:
According to [this](https://github.community/t/unable-to-push-to-ghcr-io-from-github-actions/191761), personal access token was the only way to authenticate to dockre registry beta. Since GITHUB_ACTOR still doesn't work, I guess we have to wait some more. 
The problem with using PAT is that `publish-libvcx` step is failing for PRs from forks.

PR #464, opened 2022-03-28
Upgrade node version to 17.8.0

mirgee:


kukgini:
IMHO. It is safer to use lts version (16.14.2) if possible unless there is a specific reason.

mirgee:
Postponing update due to significant [slowdown](https://github.com/nodejs/node/issues/32226) on node 17.

Patrik-Stas:
Closing in favor of recently merged
https://github.com/hyperledger/aries-vcx/pull/643 
which enables upgrade to recent versions of node, however that comes with significant perf cost compared to nodejs 12.

PR #484, opened 2022-05-11
Add get txn feature

mirgee:

### DIFF
--- a/aries_vcx/src/libindy/utils/anoncreds.rs
+++ b/aries_vcx/src/libindy/utils/anoncreds.rs
@@ -649,12 +649,44 @@ pub async fn libindy_to_unqualified(entity: &str) -> VcxResult<String> {
         .map_err(VcxError::from)
 }
 
+async fn libindy_build_get_txn_request(submitter_did: Option<&str>, seq_no: i32) -> VcxResult<String> {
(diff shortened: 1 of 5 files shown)


aries_vcx/src/libindy/utils/anoncreds.rs
```diff
@@ -649,6 +649,31 @@ pub async fn libindy_to_unqualified(entity: &str) -> VcxResult<String> {
...
+}
+
+pub async fn build_get_txn_request(did: Option<&str>, seq_no: i32) -> VcxResult<String> {
+    trace!("build_get_txn_request >>> did: {:?}, seq_no: {}", did, seq_no);
+    let request = libindy_build_get_txn_request(did, seq_no).await?;
+    let request = append_txn_author_agreement_to_request(&request).await?;
+    Ok(request)
+}
+
+pub async fn get_ledger_txn(did: Option<&str>, seq_no: i32) -> VcxResult<String> {
```
Patrik-Stas, line 665: Is there any case where you'd want to submit read transactions signed by a specific did? If it doesn't needed, I'd suggest using only `libindy_submit_request(&req).await?` and removing the `did` parameter. 
If there's case to keep it, I'd suggest rename here and on `_c` layer to `submitter_did` to avoid confusion

Patrik-Stas (approved):

PR #491, opened 2022-05-23
Parametrize issuance with rev reg

mirgee:


aries_vcx/src/libindy/credential_def/mod.rs
```diff
@@ -149,6 +174,10 @@ impl CredentialDef {
         self.state == PublicEntityStateType::Published
     }
 
+    pub fn get_support_revocation(&self) -> bool {
```
Patrik-Stas, line 177: not used

aries_vcx/src/libindy/credential_def/mod.rs
```diff
@@ -149,6 +174,10 @@ impl CredentialDef {
...
+        self.support_revocation
```
Patrik-Stas, line 178: Seems like practically not used either

PR #495, opened 2022-06-06
Include @id on forward message

Patrik-Stas:


niallshaw-absa (approved):

agency_client/src/lib.rs
```diff
@@ -269,18 +270,20 @@ pub struct ForwardV2 {
     fwd: String,
     #[serde(rename = "@msg")]
     msg: Value,
+    #[serde(rename = "@id")]
+    id: String,
```
mirgee, line 274: Let's use a special `MessageId` type  for the `@id` field the same way we do in the `aries_vcx` crate

PR #558, opened 2022-09-09
Ledger comm/validate response

Patrik-Stas:
- Add validations to ledger responses. Return error if transaction was rejected
- Update `publish_local_revocations` implementation to assure that accumulated deltas are deleted only if publishing to ledger has been successful. 
- Removes function to revoke single credential `revoke_credential` - if local revocations has been performed (eg `get_rev_reg_delta` resolved some unpublished accumulated delta), calling this function will fail, as `prevAccum` in delta returned by `libindy_issuer_revoke_credential` actually refers to state of accumulator AFTER the local revocations, not the state of accumulator on the ledger. Attempt to publish this revocation delta will fail due to mismatch of ledger state of `accum` and value of `prevAccum` in delta returned by `libindy_issuer_revoke_credential`. With old implementation of `revoke_credential`, the delta is lost. That's a problem because, the revocation has changes state of internal accumulator, however the delta was not merged into other accumulated delta. This makes further revocation in this revocation registry, least to say difficult - with current implementation, any attempt to perform further revocations would fail. 
This can be prevented if we always accumulate deltas returned by `libindy_issuer_revoke_credential`, which is what this PR does. 
The migration strategy is simply use combination of `revoke_credential_local` and `publish_local_revocations` instead. That way the calling code is more transparent about the fact that ALL local revocations will be published.

- Minor variables/function renames, but particularly removal of "cache" vocabulary for storing revocation deltas. Rev delta storage is not cache, as it does not serve for optimization, but rather storing quite quite important data.


libvcx/src/api_lib/api_handle/issuer_credential.rs
```diff
@@ -188,6 +188,7 @@ pub async fn send_credential(handle: u32, connection_handle: u32) -> VcxResult<u
     Ok(error::SUCCESS.code_num)
 }
 
+#[deprecated(note = "Use combination of issuer_credential::revoke_credential_local and revocation_registry::publish_revocations instead")]
```
mirgee, line 191: This function also publishes locally revoked credentials, so it doesn't work as intended, and that is the reason for deprecation? But the suggested solution doesn't offer a different behavior...

aries_vcx/src/libindy/utils/anoncreds.rs
```diff
@@ -852,19 +855,25 @@ pub async fn revoke_credential_local(
     rev_reg_id: &str,
     cred_rev_id: &str,
 ) -> VcxResult<()> {
-    let mut new_delta = libindy_issuer_revoke_credential(wallet_handle, tails_file, rev_reg_id, cred_rev_id).await?;
-    if let Some(old_delta) = get_rev_reg_delta_cache(wallet_handle, rev_reg_id).await {
-        new_delta = libindy_issuer_merge_revocation_registry_deltas(old_delta.as_str(), new_delta.as_str()).await?;
+    let mut new_delta_json = libindy_issuer_revoke_credential(wallet_handle, tails_file, rev_reg_id, cred_rev_id).await?;
+    info!("revoke_credential_local >>> new_delta_json: {}", new_delta_json);
```
mirgee, line 859: Maybe some of these logs can be a `debug` log level

aries_vcx/src/libindy/utils/anoncreds.rs
```diff
@@ -852,19 +855,25 @@ pub async fn revoke_credential_local(
...
-        match clear_rev_reg_delta_cache(wallet_handle, rev_reg_id).await {
-            Ok(_) => publish_rev_reg_delta(wallet_handle, pool_handle, &submitter_did, rev_reg_id, &delta).await,
-            Err(err) => Err(err),
+pub async fn publish_local_revocations(wallet_handle: WalletHandle, pool_handle: PoolHandle, submitter_did: &str, rev_reg_id: &str) -> VcxResult<()> {
+    if let Some(delta) = get_rev_reg_delta(wallet_handle, rev_reg_id).await {
+        publish_rev_reg_delta(wallet_handle, pool_handle, &submitter_did, rev_reg_id, &delta).await?;
+        info!("publish_local_revocations >>> rev_reg_delta published for rev_reg_id {}", rev_reg_id);
+        match clear_rev_reg_delta(wallet_handle, rev_reg_id).await {
+            Ok(_) => info!("publish_local_revocations >>> rev_reg_delta cache cleared for rev_reg_id {}", rev_reg_id),
+            Err(err) => error!("publish_local_revocations >>> rev_reg_delta cache clear failed for rev_reg_id {}: {}", rev_reg_id, err)
```
mirgee, line 874: Perhaps we should return an error in this case - it gets the library into an inconsistent state, which is quite severe

mirgee (changes requested):
Can you please add a test where you revoke credentials locally and then revoke publicly?

libvcx/src/api_lib/api_handle/issuer_credential.rs
mirgee: I see, this is addressed in #565.

PR #561, opened 2022-09-14
Refactor building problem reports, add logging

Patrik-Stas:

### DIFF
--- a/aries_vcx/src/protocols/common.rs
+++ b/aries_vcx/src/protocols/common.rs
@@ -1,0 +1,31 @@
+use crate::messages::error::ProblemReport;
+
+pub fn build_problem_report_msg(comment: Option<String>, thread_id: &str) -> ProblemReport {
+    ProblemReport::create()
+        .set_out_time()
+        .set_comment(comment)
+        .set_thread_id(thread_id)
+}
+
+#[cfg(test)]
+#[cfg(feature = "general_test")]
+mod test {
+    use crate::messages::a2a::MessageId;
+    use crate::protocols::common::build_problem_report_msg;
+    use crate::utils::devsetup::{was_in_past, SetupMocks};
+
+    #[test]
+    #[cfg(feature = "general_test")]
+    fn test_holder_build_problem_report_msg() {
+        let _setup = SetupMocks::init();
+        let msg = build_problem_report_msg(Some("foo".into()), "12345");
+
+        assert_eq!(msg.id, MessageId::default());
+        assert_eq!(msg.thread.unwrap().thid.unwrap(), "12345");
+        assert!(was_in_past(
+            &msg.timing.unwrap().out_time.unwrap(),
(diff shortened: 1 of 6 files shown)


aries_vcx/src/protocols/issuance/holder/state_machine.rs
```diff
@@ -517,6 +515,7 @@ pub async fn create_credential_request(
 
 async fn _make_credential_request(
     wallet_handle: WalletHandle,
+    thread_id: String,
```
mirgee, line 518: It seems that `&str` would be sufficient and avoid cloning, same for `my_pw_did` argument

aries_vcx/src/protocols/proof_presentation/prover/state_machine.rs
```diff
@@ -52,21 +53,14 @@ pub enum ProverFullState {
     Finished(FinishedState),
 }
 
-fn build_presentation_message(thread_id: &str, presentation_attachment: String) -> VcxResult<Presentation> {
+fn build_presentation_msg(thread_id: &str, presentation_attachment: String) -> VcxResult<Presentation> {
     Ok(Presentation::create()
         .ask_for_ack()
         .set_thread_id(&thread_id)
```
mirgee, line 59: No need for reference here

aries_vcx/src/protocols/common.rs
```diff
@@ -0,0 +1,31 @@
+use crate::messages::error::ProblemReport;
+
+pub fn build_problem_report_msg(comment: Option<String>, thread_id: &str) -> ProblemReport {
```
mirgee, line 3: A benefit of using `Option<String>` is that it gives the user flexibility to avoid cloning unnecessarily in cases when they already have a `String` or `Option<String>` to work with; but it seems rather rare in our case... Otherwise, when cloning is necessary and the user usually has a reference, I think it is better done at "the lower level"... I wonder what are your thoughts here?

mirgee (changes requested):

aries_vcx/src/protocols/issuance/holder/state_machine.rs
Patrik-Stas: I can make this &str, but it doesn't really help, because this is used to set 
```
CredentialRequest::create()
        .set_thread_id(thread_id)
```
and `set_thread_id` run clone anyway. I'd suggest in another PR we change macros like `threadlike_optional` etc. to require `String` instead of `&str`, as it's anyway ends up cloning the `str`. I think this will then trickle-up on over layers of calling code, as well and will probably lead to some unification/clearing of how we pass `&str`/`String`, as it's fairly inconsistent now.

aries_vcx/src/protocols/common.rs
Patrik-Stas: my thoughts in the reply above

mirgee (approved):

PR #565, opened 2022-09-14
Remove function to revoke+publish individual credential

Patrik-Stas:
After changes done on https://github.com/hyperledger/aries-vcx/pull/558 it became clear that `vcx_revoke_credential` is problematic function as:
- before #558 it could cause entering problematic state after losing revocation delta
- changes in #558 prevent problematic behaviour, but causes `vcx_revoke_credential` to have side effects - if other credentials in same revocation registry has been revoked locally before and their deltas are accumulated, it would publish revocation for those credential as well. 
The pre #558 approach in unacceptable, the post #558 approach is not transparent. The right way is to revoke credential locally, then published the delta, accounting for the fact that can lead to actually publishing merged revocation delta of multiple local revocations.

### DIFF
--- a/aries_vcx/src/handlers/issuance/issuer.rs
+++ b/aries_vcx/src/handlers/issuance/issuer.rs
@@ -6,6 +6,7 @@ use agency_client::agency_client::AgencyClient;
 
 use crate::error::prelude::*;
 use crate::handlers::connection::connection::Connection;
+use crate::libindy::utils::anoncreds;
 use crate::libindy::utils::anoncreds::libindy_issuer_create_credential_offer;
 use crate::messages::a2a::A2AMessage;
 use crate::messages::issuance::credential_offer::OfferInfo;
@@ -13,7 +14,7 @@ use crate::messages::issuance::credential_proposal::CredentialProposal;
 use crate::messages::issuance::CredentialPreviewData;
 use crate::messages::mime_type::MimeType;
 use crate::protocols::issuance::actions::CredentialIssuanceAction;
-use crate::protocols::issuance::issuer::state_machine::{IssuerSM, IssuerState};
+use crate::protocols::issuance::issuer::state_machine::{IssuerSM, IssuerState, RevocationInfoV1};
 use crate::protocols::SendClosure;
 
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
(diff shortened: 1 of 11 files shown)


aries_vcx/src/handlers/issuance/issuer.rs
```diff
@@ -163,9 +164,24 @@ impl Issuer {
         self.issuer_sm.find_message_to_handle(messages)
     }
 
-    // todo: eliminate publish parameter - only revoke locally, leave it to caller to publish accumulated deltas
-    pub async fn revoke_credential(&self, wallet_handle: WalletHandle, pool_handle: PoolHandle, issuer_did: &str, publish: bool) -> VcxResult<()> {
-        self.issuer_sm.revoke(wallet_handle, pool_handle, issuer_did, publish).await
+    pub async fn revoke_credential_local(&self, wallet_handle: WalletHandle) -> VcxResult<()> {
+        let revocation_info: RevocationInfoV1 = self.issuer_sm.get_revocation_info().ok_or(VcxError::from_msg(
+            VcxErrorKind::InvalidState,
+            "Credential is not revocable, not revocation info has been resolved.",
```
mirgee, line 170: "no revocation info has been found."

libvcx/src/api_lib/api_c/issuer_credential.rs
```diff
@@ -886,23 +825,18 @@ pub extern "C" fn vcx_issuer_revoke_credential_local(
...
 
     let source_id = issuer_credential::get_source_id(credential_handle).unwrap_or_default();
-    let issuer_did: String = match settings::get_config_value(settings::CONFIG_INSTITUTION_DID) {
-        Ok(err) => err,
-        Err(err) => return err.into(),
-    };
     info!(
-        "vcx_issuer_revoke_local(command_handle: {}, credential_handle: {}, issuer_did: {}) source_id: {}",
-        command_handle, credential_handle, issuer_did, source_id
+        "vcx_issuer_revoke_local(command_handle: {}, credential_handle: {}) source_id: {}",
```
mirgee, line 829: The function is called `vcx_issuer_revoke_credential_local` (same name makes it easier for fulltext search it based on logs)

mirgee (changes requested):

PR #577, opened 2022-09-21
Separate messages module into a crate

mirgee:


aries_vcx/src/libindy/utils/ledger.rs
```diff
@@ -352,6 +354,40 @@ pub async fn get_service(pool_handle: PoolHandle, did: &Did) -> VcxResult<AriesS
     })
 }
 
+pub async fn into_did_doc(pool_handle: PoolHandle, invitation: &Invitation) -> VcxResult<DidDoc> {
```
Patrik-Stas, line 357: I think this actually makes sense to be under `messages` module rather, however instead of 
```
           let service = get_service(pool_handle, &invitation.did).await.unwrap_or_else(|err| {
                error!("Failed to obtain service definition from the ledger: {}", err);
                AriesService::default()
            });
            (service.service_endpoint, service.recipient_keys, service.routing_keys)
```
we should imo have something like 
```
        Invitation::Public(invitation) => {
            did_doc.set_id(invitation.did.to_string());
            return indy_resolver::resolve_ddo(invitation.did);
        }
```
And ddo_resolver could be trait, I imagine we could have our own ddo resolution, or we could also implement that trait experimentally by calling http://uniresolver.io/

Patrik-Stas: Doing this, `into_did_doc` could then stay in messages

Patrik-Stas: But creating ddo resolver interface and start using it in the codebase might be out of scope here. So just a suggestion for a next steps :-)

mirgee: I agree, but my goal here is just to remove all mentions of handle as simply as possible and then move code back in when it makes sense.

aries_vcx/src/error.rs
```diff
@@ -358,6 +359,49 @@ impl From<AgencyClientErrorKind> for VcxErrorKind {
...
 }
 
+impl From<messages::error::VcxError> for VcxError {
+    fn from(msg_err: messages::error::VcxError) -> VcxError {
+        let vcx_error_kind: VcxErrorKind = msg_err.kind().into();
+        VcxError::from_msg(vcx_error_kind, msg_err.to_string())
+    }
+}
+
+impl From<MessagesErrorKind> for VcxErrorKind {
```
Patrik-Stas, line 369: Maybe for separate PR, but this should be quite reworked - numeric error codes, many of the error kinds, C string definitions, all these things don't really make sense for this crate

aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -51,7 +52,7 @@ impl OutOfBandReceiver {
                                 return Ok(Some(connection));
                             }
                         };
-                        if did_doc.resolve_service()? == service.resolve(pool_handle).await? {
+                        if did_doc.resolve_service()? == resolve_service(pool_handle, &service).await? {
```
Patrik-Stas, line 55: how about we rename `did_doc.resolve_service` to  `did_doc.get_service`? As that function is only trying to construct service locally.

aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -151,7 +152,7 @@ impl OutOfBandReceiver {
     }
 
     pub fn to_string(&self) -> VcxResult<String> {
-        self.oob.to_string()
+        self.oob.to_string().map_err(|err| err.into())
```
Patrik-Stas, line 155:
```
json!(self.oob).to_string()
```
doesn't ever return error - then we could also change return value of this function to `String`. 

I've noticed we have many functions which return `VcxResult` whereas they actually never return Error. Would be nice to clean this up

aries_vcx/src/handlers/out_of_band/sender.rs
```diff
@@ -88,7 +89,7 @@ impl OutOfBandSender {
     }
 
     pub fn to_string(&self) -> VcxResult<String> {
-        self.oob.to_string()
+        self.oob.to_string().map_err(|err| err.into())
```
Patrik-Stas, line 92: same here

aries_vcx/src/handlers/proof_presentation/prover.rs
```diff
@@ -134,6 +134,7 @@ impl Prover {
             .presentation_request()?
             .request_presentations_attach
             .content()
+            .map_err(|err| err.into())
```
Patrik-Stas, line 137: seems superflous, converting VcxError to VcxError

Patrik-Stas: ah I see ,that relates to my first comment about reworking Error in messages crate. Can we at least rename this "messages" VcxError/VcxResult to something like `MessagesError/MessagesResult`? FYI Intelij can do this mass rename on 2 clicks.

aries_vcx/src/libindy/proofs/proof_request.rs
```diff
@@ -1,290 +0,0 @@
...
-
-use serde_json;
-
-use crate::error::prelude::*;
-use crate::libindy::proofs::proof_request_internal::{AttrInfo, NonRevokedInterval, PredicateInfo};
-use crate::libindy::utils::anoncreds;
-
-#[derive(Serialize, Deserialize, Builder, Debug, PartialEq, Clone)]
-#[builder(setter(into), default)]
-pub struct ProofRequestData {
```
Patrik-Stas, line 12: These are indy specific structures which I think should stay in `aries-vcx`. From messages crate perspective, we can work with this just as attachment. I think with few modifications to code, messages crate does not need to know this `ProofRequestData` exists

messages/src/error.rs
```diff
@@ -0,0 +1,739 @@
...
+use failure::{Backtrace, Context, Fail};
+
+use crate::utils::error;
+
+pub mod prelude {
+    pub use super::{err_msg, VcxError, VcxErrorExt, VcxErrorKind, VcxResult, VcxResultExt};
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Fail)]
+pub enum VcxErrorKind {
```
Patrik-Stas, line 173: The `ProblemReport` as aries message and Rust errors should be 2 separate files

messages/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+rand = "0.7.3"
+serde = "1.0.97"
+serde_json = "1.0.40"
+serde_derive = "1.0.97"
+url = "1.5.1"
+regex = "1.1.0"
+rust-base58 = "0.0.4"
+rmp-serde = "0.13.7"
+base64 = "0.8.0"
+openssl = { version = "0.10.35", features = ["vendored"] }
```
Patrik-Stas, line 26: It seems there's number of extra dependencies

Patrik-Stas (changes requested):
pretty cool! Left some comments but this is definitely good step forwards

PR #584, opened 2022-09-29
Derive Serialize / Deserialize traits for AgencyClient

mirgee:

### DIFF
--- a/agency_client/src/agency_client.rs
+++ b/agency_client/src/agency_client.rs
@@ -5,7 +5,7 @@ use crate::configuration::AgencyClientConfig;
 use crate::error::AgencyClientResult;
 use crate::{validation, AgencyClientError, AgencyClientErrorKind};
 
-#[derive(Clone)]
+#[derive(Clone, Serialize, Deserialize)]
 pub struct AgencyClient {
     wallet_handle: WalletHandle,
     pub agency_url: String,


agency_client/src/agency_client.rs
```diff
@@ -5,7 +5,7 @@ use crate::configuration::AgencyClientConfig;
 use crate::error::AgencyClientResult;
 use crate::{validation, AgencyClientError, AgencyClientErrorKind};
 
-#[derive(Clone)]
+#[derive(Clone, Serialize, Deserialize)]
```
Patrik-Stas, line 8: - This adds alternative approach how to serialize and deserialize `AgencyClient`. Current approach is to create the cloud agent, receive `AgencyClientConfig` and use that to build AgencyClient. Currently we need to care about compatibility changes made to of `AgencyClientConfig`, but if we add this alternative approach, suddenly we should also mind compatibility changes in AgencyClient itself.
- Additionally `AgencyClient` should not be serialized with `wallet_handle`, as that is runtime information. Well, that can be easily solved with serde's `skip`, but then my first argument is still intact.

Patrik-Stas (changes requested):

agency_client/src/agency_client.rs
mirgee: This is a crude workaround, my bad making a PR out of it.

PR #608, opened 2022-10-20
Revocation notification

mirgee:
https://github.com/hyperledger/aries-rfcs/tree/main/features/0721-revocation-notification-v2


aries_vcx/src/handlers/issuance/issuer.rs
```diff
@@ -164,6 +165,23 @@ impl Issuer {
         .await
     }
 
+    pub async fn send_revocation_notification(&mut self, comment: Option<String>, send_message: SendClosure) -> VcxResult<()> {
+        // TODO: Check if actually revoked
+        if self.issuer_sm.is_revokable() {
+            let rev_msg = RevocationNotification::create()
+                .set_credential_id(self.get_rev_reg_id()?, self.get_rev_id()?)
+                .set_comment(comment)
+                .set_thread_id(&self.get_thread_id()?)
```
Patrik-Stas, line 174: As per private discussion, I think this message should not have `thread_id`, only `@id`

aries_vcx/src/handlers/issuance/issuer.rs
```diff
@@ -164,6 +165,23 @@ impl Issuer {
         .await
     }
 
+    pub async fn send_revocation_notification(&mut self, comment: Option<String>, send_message: SendClosure) -> VcxResult<()> {
+        // TODO: Check if actually revoked
+        if self.issuer_sm.is_revokable() {
+            let rev_msg = RevocationNotification::create()
```
Patrik-Stas, line 171: I am fine keeping `send_revocation_notification` directly on issuer handler per practicality, but I think it'd be cleaner to have the building of notification message extracted into `protocols/revocation_notification/mod.rs`, similarly like we currently do for ping or oob protocols. 
Then analogically add tests there (as we do for ping) for building both notification and response ack message.

aries_vcx/src/protocols/issuance/holder/state_machine.rs
```diff
@@ -288,9 +302,21 @@ impl HolderSM {
                     HolderFullState::RequestSent(state_data)
                 }
             },
-            HolderFullState::Finished(state_data) => {
-                warn!("Unable to process received message in this state");
-                HolderFullState::Finished(state_data)
+            HolderFullState::Finished(state_data) => match cim {
+                CredentialIssuanceAction::RevocationNotification(notification) => {
+                    if notification.ack_on(AckOn::Receipt) {
+                        let ack = build_revocation_ack(&notification.get_thread_id());
```
Patrik-Stas, line 308: if you agree with my previous comment that notification should have no `thread_id`, we would have to build thread_id here from `@id` of received message

Patrik-Stas (approved):

Patrik-Stas (approved):

Issue #583, opened 2022-09-27
Unify aries FSM approach


Patrik-Stas:
Currently
- connection protocol is using new "function" approach, where attempt to invoke certain state transition can be done by calling a method
- issuance and proof FSMs are using original "message" approach, where there's 1 big "step" function which performs state transition based on custom "message" abstraction (which sometimes is a wrap around an actual aries message, sometimes it's just a empty enumeration signalling execution of some action)

PR #610, opened 2022-10-21
Refactor issuer / holder SMs

mirgee:
Holder handler calls specific SM methods instead of relying on a step function. This way, if an explicit "action" is performed in an invalid state, an error is returned, instead of a noop.

Addresses #583 .

### DIFF
--- a/aries_vcx/src/indy/primitives/credential_definition.rs
+++ b/aries_vcx/src/indy/primitives/credential_definition.rs
@@ -208,7 +208,7 @@ impl CredentialDef {
 
     pub fn get_data_json(&self) -> VcxResult<String> {
         serde_json::to_string(&self)
-            .map_err(|err| VcxError::from_msg(VcxErrorKind::SerializationError, "Failed to serialize credential definition"))
+            .map_err(|_| VcxError::from_msg(VcxErrorKind::SerializationError, "Failed to serialize credential definition"))
     }
 
     pub fn get_source_id(&self) -> &String {
--- a/aries_vcx/src/protocols/issuance/holder/state_machine.rs
+++ b/aries_vcx/src/protocols/issuance/holder/state_machine.rs
@@ -153,14 +153,6 @@ impl HolderSM {
         None
     }
 
-    pub fn step(state: HolderFullState, source_id: String, thread_id: String) -> Self {
-        HolderSM {
-            state,
-            source_id,
-            thread_id,
-        }
-    }
-
     pub async fn handle_message(
         self,
         wallet_handle: WalletHandle,
--- a/aries_vcx/src/protocols/issuance/holder/state_machine.rs
+++ b/aries_vcx/src/protocols/issuance/holder/state_machine.rs
@@ -249,9 +249,17 @@ impl HolderSM {
     }
 
     pub async fn send_proposal(self, proposal_data: CredentialProposalData, send_message: SendClosure) -> VcxResult<Self> {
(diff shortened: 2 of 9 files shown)


aries_vcx/src/indy/primitives/credential_definition.rs
```diff
@@ -208,7 +208,7 @@ impl CredentialDef {
 
     pub fn get_data_json(&self) -> VcxResult<String> {
         serde_json::to_string(&self)
-            .map_err(|err| VcxError::from_msg(VcxErrorKind::SerializationError, "Failed to serialize credential definition"))
+            .map_err(|_| VcxError::from_msg(VcxErrorKind::SerializationError, "Failed to serialize credential definition"))
```
Patrik-Stas, line 211: The entire implementation can be replaced by error-free
```
json!(&self).to_string()
```
which begs for question if this function need to actually exist 🤔

aries_vcx/src/protocols/issuance/holder/state_machine.rs
```diff
@@ -296,6 +248,91 @@ impl HolderSM {
         Ok(HolderSM::step(state, source_id, thread_id))
```
Patrik-Stas, line 248: Seems like this `step` function is doing nothing and can be removed

aries_vcx/src/protocols/issuance/holder/state_machine.rs
```diff
@@ -172,18 +172,16 @@ impl HolderSM {
...
-                    let proposal = CredentialProposal::from(proposal_data).set_id(&thread_id);
-                    send_message.ok_or(VcxError::from_msg(
+                    let send_message = send_message.ok_or(VcxError::from_msg(
                         VcxErrorKind::InvalidState,
                         "Attempted to call undefined send_message callback",
-                    ))?(proposal.to_a2a_message())
-                    .await?;
-                    HolderFullState::ProposalSent(ProposalSentState::new(proposal))
+                    ))?;
+                    self.send_proposal(proposal_data, send_message).await?.state
```
Patrik-Stas, line 184: This is quite a code smell, if nothing else changes than state, why only state is not removed? I know the answer, because we use `send_proposal` from `holder.rs` where the entire `SmHolder` is what we need. 
The clean solution I hence advocate here, is to modify the 
```
let state = match state {
```
on line `177` such that the result of the match is entire `HolderSm`, not just state. That way we don't have to do any field cherry picking in situations like on this line. The simpler branches, like `186` will be more verbose but it'll be much cleaner and less error prone.

That being said tho... ideally `handle_message` would not exist in state machine, just like it does not exist for connection state machines.

Patrik-Stas (changes requested):
Good steps forward, but it's yet only halfway towards addressing #583 . In connection protocol, `handle_message` function doesn't exist, any possible state transition is a function. Also currently there's 2 pathways to do the same thing, like sending credential request (either directly or via `handle_message(CredentialIssuanceAction::CredentialRequest(CredentialRequest))`

But it's still improvement toward #583 , so we can merge it after after at least addressing the comments left

Patrik-Stas:
Why close?

aries_vcx/src/indy/primitives/credential_definition.rs
mirgee: This is out of scope of this PR

aries_vcx/src/protocols/issuance/holder/state_machine.rs
mirgee: Simplified the function as discussed

Patrik-Stas (approved):

PR #613, opened 2022-10-22
Ci/improvements

Patrik-Stas:
- Fix CI warnings (usage of set-output)
- As `build-image` action was trying to pull images from `docker.pkg.github.com`, it never worked because that's not docker registry we are logged into - so we didn't reuse previously built docker layer
- Generalized `construct-run-info` to refer to on `$MAIN_BRANCH` instead of hardcoding `main`
- Added action template for detecting skip PR tags (currently supports opt-in PR tags `skip-android` and `skip-ios`)
- Fixed information passing into `azure/docker-login@v1`
- Fixed publishing jobs - images are published under tag`$PUBLISH_VERSION` 
- Modified `build-image` such that it publishes image `<docker-img>:<branch>` to docker registry for caching purposes
- Adjusted publish-image more robust so it doesn't try to load image from cache if it already exist in local docker repo

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,9 +85,9 @@ jobs:
       - name: "Docker Login"
         uses: azure/docker-login@v1
         with:
-          login-server: env.URL_DOCKER_REGISTRY
-          username: env.GITHUB_ACTOR
-          password: secrets.GITHUB_TOKEN
+          login-server: ${{ env.URL_DOCKER_REGISTRY }}
+          username: ${{ env.GITHUB_ACTOR }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: "Build and cache image"
         uses: ./.github/actions/build-image
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -605,14 +612,16 @@ jobs:
...
       - name: "Load android image"
+        if: ${{ needs.workflow-setup.outputs.SKIP_ANDROID != 'true' }}
         uses: ./.github/actions/load-image
         with:
           docker-img: ${{ env.DOCKER_IMG_CACHED_ANDROID }}
       - name: "Run android tests"
+        if: ${{ needs.workflow-setup.outputs.SKIP_ANDROID != 'true' }}
         run: |
           rm -rf /tmp/imgcache
           docker run --rm -i  $DOCKER_IMG_CACHED_ANDROID \
```
mirgee, line 627: How about we add `|| 1` here so that the intermittently failing android test is not failing the whole CI run until we address the issue?

Patrik-Stas: I think we should bear the pain to make sure we actually address that :-D I worry we add `|| 1` and it'll stay there longer than it should.

mirgee (approved):

PR #633, opened 2022-10-30
Cleanup: delete and ignore temporary OS file (.DS_Store)

jodaAtGithub:
A `.DS_Store` file is specific temporary local file, generated by Mac OSX, containing specific settings for a local setup. It is common practice to prevent them from pushing into the VCS.

The `.DS_Store` filename is already listed in `.gitignore`, I guess that those file was already listed before they got ignored.


Patrik-Stas:
Thank you @jodaAtGithub

PR #635, opened 2022-10-31
Revocation status "self-check"

mirgee:
Enables credential revocation status verification for issuer and holder based on revocation registry delta obtained from the ledger, i.e. without generation of a non-revocation proof.

### DIFF
--- a/aries_vcx/src/indy/credentials/mod.rs
+++ b/aries_vcx/src/indy/credentials/mod.rs
@@ -45,7 +45,7 @@ pub async fn is_cred_revoked(
     let from = None;
     let to = Some(get_time().sec as u64 + 100);
     let rev_reg_delta = RevocationRegistryDelta::create_from_ledger(pool_handle, rev_reg_id, from, to).await?;
-    Ok(rev_reg_delta.revoked().iter().any(|s| s.to_string().eq(&rev_id)))
+    Ok(rev_reg_delta.revoked().iter().any(|s| s.to_string().eq(rev_id)))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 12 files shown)


aries_vcx/src/indy/credentials/mod.rs
```diff
@@ -4,68 +4,146 @@ pub mod issuer;
...
+    ))
+}
+
+pub async fn is_cred_revoked(
+    pool_handle: PoolHandle,
+    rev_reg_id: &str,
+    rev_id: &str,
+) -> VcxResult<bool> {
+    let from = None;
+    let to = Some(get_time().sec as u64 + 100);
```
Patrik-Stas, line 46: Why is it needed to go "into future"? Wouldn't current timestamp do just fine?

aries_vcx/src/indy/credentials/mod.rs
```diff
@@ -4,68 +4,146 @@ pub mod issuer;
...
+        "Credenial revocation id missing on credential - is this credential revokable?",
+    ))
+}
+
+pub async fn is_cred_revoked(
+    pool_handle: PoolHandle,
+    rev_reg_id: &str,
+    rev_id: &str,
+) -> VcxResult<bool> {
+    let from = None;
```
Patrik-Stas, line 45: I previously understood we would be checking delta from something like `(now - 1 sec, now)`, but it seem here we are doing interval `(ever-since, ~now)`.
Is that the case, or am I missing something?

aries_vcx/src/indy/credentials/mod.rs
mirgee: As per PM, decided add some buffer in case the interval was open. But through experimentation, it seems the current timestamp would indeed do fine. Do you insist on using the current timestamp with no buffer?

Patrik-Stas (approved):

PR #642, opened 2022-11-09
create prefix did:sov

guijd3p:
this task was performed implementing did:sov in out-of-band message generation, now every time generating a public out-of-band invite will contain did:sov, it also managed to process messages without the did:sov prefix

### DIFF
--- a/libvcx/src/api_lib/api_handle/out_of_band.rs
+++ b/libvcx/src/api_lib/api_handle/out_of_band.rs
@@ -102,9 +102,7 @@ pub fn append_service(handle: u32, service: &str) -> VcxResult<()> {
 pub fn append_service_did(handle: u32, did: &str) -> VcxResult<()> {
     trace!("append_service_did >>> handle: {}, did: {}", handle, did);
     let mut oob = OUT_OF_BAND_SENDER_MAP.get_cloned(handle)?;
-    let mut did_sov = "did:sov:".to_owned();
-    did_sov.push_str(&did);
-    oob = oob.clone().append_service(&ServiceResolvable::Did(Did::new(&did_sov)?));
+    oob = oob.clone().append_service(&ServiceResolvable::Did(Did::new(did)?));
     OUT_OF_BAND_SENDER_MAP.insert(handle, oob)
 }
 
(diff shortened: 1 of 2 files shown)


libvcx/src/api_lib/api_handle/out_of_band.rs
```diff
@@ -102,7 +102,9 @@ pub fn append_service(handle: u32, service: &str) -> VcxResult<()> {
 pub fn append_service_did(handle: u32, did: &str) -> VcxResult<()> {
     trace!("append_service_did >>> handle: {}, did: {}", handle, did);
     let mut oob = OUT_OF_BAND_SENDER_MAP.get_cloned(handle)?;
-    oob = oob.clone().append_service(&ServiceResolvable::Did(Did::new(did)?));
+    let mut did_sov = "did:sov:".to_owned();
```
Patrik-Stas, line 105: Let's please leave caller of `append_service_did` to decide on format. With this change, we are forcing caller to supply the legac / non-prefixed format, and do some work for him and it's not obvious. 

Also this is forces OOB sender to dispatch `did:sov` prefixed DID, which wouldn't be understood on aries-vcx clients which have not yet been updated to  revision containing this commit. The suggestion above also solves this problem, as OOB sender can decide himself to use non/prefixed DIDs in OOBs.

Consequently please adjust `test_build_oob_sender_append_services` (or create new test) to cover both cases -when OOB sender produces prefixed/non-prefixed DID.

Patrik-Stas (changes requested):

Patrik-Stas:
Thank you @guijd3p , minor change requested but after you address that I believe we can merge. 

I can image a lot refactoring which could be done in regard to working with DIDs, as right now we don't really have them typed, instead we work with `str/string`. But that's just a side note, not in scope of this PR.

libvcx/src/api_lib/api_handle/out_of_band.rs
guijd3p: IIt makes perfect sense what you said, so I think leaving the method the same as before, and creating a new test to add the did:sov.

Patrik-Stas (approved):

mirgee (approved):

PR #645, opened 2022-11-11
Nonmediated connection

mirgee:
Introduces a new, generic connection handler type, which will eventually replace the current - strictly mediated, tightly coupled with agency client and unrelated protocols (e.g. OOB, disclose, ping, ...) - connection handler type.

### DIFF
--- a/aries_vcx/src/handlers/out_of_band/receiver.rs
+++ b/aries_vcx/src/handlers/out_of_band/receiver.rs
@@ -4,7 +4,7 @@ use vdrtools_sys::PoolHandle;
 use agency_client::agency_client::AgencyClient;
 
 use crate::error::prelude::*;
-use crate::handlers::connection::connection::Connection;
+use crate::handlers::connection::mediated_connection::MediatedConnection;
 use messages::out_of_band::invitation::OutOfBandInvitation;
 use messages::a2a::A2AMessage;
 use messages::attachment::AttachmentId;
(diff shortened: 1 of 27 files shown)


aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -41,8 +41,8 @@ impl OutOfBandReceiver {
     pub async fn connection_exists<'a>(
         &self,
         pool_handle: PoolHandle,
-        connections: &'a Vec<&'a Connection>,
-    ) -> VcxResult<Option<&'a Connection>> {
+        connections: &'a Vec<&'a MediatedConnection>,
+    ) -> VcxResult<Option<&'a MediatedConnection>> {
```
Patrik-Stas, line 45: Wondering if we should delete this method and let upper layers track which connection "is who". 
As it's kind of code smell that we should now probably add similar support also for `Connection` instances, while the 2 are quite similar.

aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -133,12 +133,12 @@ impl OutOfBandReceiver {
         Ok(None)
     }
 
-    pub async fn build_connection(&self, agency_client: &AgencyClient, did_doc: DidDoc, autohop_enabled: bool) -> VcxResult<Connection> {
+    pub async fn build_connection(&self, agency_client: &AgencyClient, did_doc: DidDoc, autohop_enabled: bool) -> VcxResult<MediatedConnection> {
```
Patrik-Stas, line 136: Same thoughts here, we could remove this and let the upper layer to choose to build `MediatedConnection` or `Connection` (or perhaps "Connection" but using did-exchange protocol etc.).

aries_vcx/src/utils/mod.rs
```diff
@@ -84,6 +84,7 @@ pub async fn send_message(
         &did_doc)
         .await?;
 
+    // TODO: Extract from agency client
```
Patrik-Stas, line 87: 👍

Patrik-Stas (approved):

Patrik-Stas (approved):

PR #646, opened 2022-11-14
Run nodejs tests with multiple nodejs versions

Patrik-Stas:


.github/workflows/main.yml
```diff
@@ -292,24 +290,33 @@ jobs:
   test-node-wrapper:
     needs: workflow-setup
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        node-version: [12.x, 18.x]
```
mirgee, line 295: What is the issue with node wrapper unit tests on node 19? The integration tests seem to run fine on node 19.

Patrik-Stas: Had some issues .... I'll try to add it again

mirgee:
Running tests for these node versions might seem to suggest that they are "supported", while in fact the slowdown makes the library unusable with a node version > 12.

Patrik-Stas:
@mirgee we might wait another 2 years for that issue to be resolved ^_^ It's open since 2020

mirgee:
@Patrik-Stas Let's merge it - we can undo once it starts causing issues, and at least we will know which change started to cause them.

mirgee (approved):

PR #665, opened 2022-11-25
Generate node bindings using napi-rs - separate crate

mirgee:
- Generate nodejs binding using `napi-rs`
- Running this on Nodejs18 comes with 20x performance improvement compared to old FFI with Nodejs12, 400x performance improvement compared to old FFI with Nodejs18
- Significantly reduced boilerplate in NodeJS classes wrapping the bindings
- Improved error handling - if we throw error from Rust, JS exception is thrown which contains following attributes
  - `napiCode` - error code in NAPI terms, for errors throw from Rust code we always set `GenericFailure`
  - `vcxErrKind` - error kind as string, eg `InvalidConfiguration`
  - `vcxErrCode` - error kind in its numeric representation, eg `1016`
  - `vcxErrMessage`- error message created within Rust code, eg `"Can not updated state with message: Message deserialization failed: {:?}"`


Patrik-Stas (changes requested):
as per discussion, just sort out making the published node-wrapper functional (or otherwise keeping original node wrapper temporarily, and having the new wrapper alongside for testing and dev)

PR #648, opened 2022-11-15
Modularize dependency on vdrtools/indy with option for indy-Vdr and Indy-Credx in place

gmulhearn-anonyome:
# Summary of Changes
* Created `Profile`, `BaseWallet`, `BaseLedger`, and `BaseAnoncreds` traits/interfaces. Where Wallet Ledger and Anoncreds APIs signatures are based mostly on indy/vdrtools APIs for the respective sections. A Profile represents a collection of these 3 traits/dependencies bundled together - Profile is derived from ACA-py's approach.
* Created "IndySdk" (ACA-py terms) implementations of Wallet/Ledger/Anoncreds - which mostly just calls the `/indy` directory functionality directly.
* Created an `IndySdkProfile` implementation which bundles together those 3 Indy impls ^.
* Created `IndyVdr` implementation of Ledger, which uses indy-vdr for ledger request/response formation and sending. The profile's wallet is used for signing transactions. The implementation of most functions is derived from ACA-py.
* Created `IndyCredx` implementation of Anoncreds, which uses indy-credx for the formation of anoncreds related objects - e.g. creating proofs and link secrets. The profile's wallet is used for storing/searching credentials and handling master secrets. The implementation of most functions is derived from ACA-py, especially around the storage of creds as wallet records.
* Created a `ModularWalletProfile` which bundles vdr Ledger and credx Anoncreds with any user provided `BaseWallet` (I.e. a pluggable wallet). This is similar to ACA-py's `AskarProfile` - which we may wish to impl in the future.
* Updated integration tests (where possible) to dynamically pick between using an `IndySdk` or `ModularWallet` Profile (by setting the `modular_deps` flag)
* Created a `MockProfile` which is used by unit tests. Currently MockProfile contains Mock[Ledger|Wallet|Anoncreds], all of which return the mock data which you'd expect to be returned if indy_mocks/did_mocks were enabled.
* Add a new CI step to rerun integration tests with coverage using the 'modular_profile' flag, which will trigger ~50 int tests to run using the modularWalletProfile instead of IndySdkProfile. (note that in the future ALL tests should be integration tests should be triggered by this flag - currently only 50 due to missing implementations)

# Future Work
* IndyVdrLedger and IndyCredx implementations were made with a focus on Holder/Prover functionality. As such, some 'institutional/verifier/issuer' functionality has stubbed implementations and needs to be populated.
* Libvcx would ideally support the ModularProfile, not just default to Indy. More thought and planning will be needed
[...]

PR #666, opened 2022-11-27
Set indy wallet master secret

pomfar:
Set indy wallet master secret
master_secret parameter has been added to wallet config
Master secret is used in some methods where the wallet config is not there,
then has been added a global access to master secret of each wallet

### DIFF
--- a/aries_vcx/src/indy/proofs/prover/prover.rs
+++ b/aries_vcx/src/indy/proofs/prover/prover.rs
@@ -65,11 +65,13 @@ pub async fn generate_indy_proof(
         build_cred_defs_json_prover(wallet_handle, pool_handle, &credentials_identifiers)
         .await?;
 
+    let master_secret = settings::get_master_secret(&wallet_handle)?;
+
     let proof = libindy_prover_create_proof(
         wallet_handle,
         proof_req_data_json,
         &requested_credentials,
-        settings::DEFAULT_LINK_SECRET_ALIAS,
+        &master_secret,
         &schemas_json,
         &credential_defs_json,
         Some(&revoc_states_json),
--- a/aries_vcx/src/indy/wallet.rs
+++ b/aries_vcx/src/indy/wallet.rs
@@ -33,6 +33,8 @@ pub struct WalletConfig {
     pub rekey: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub rekey_derivation_method: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub master_secret: Option<String>,
 }
 
 #[derive(Clone, Debug, Default, Builder, Serialize, Deserialize)]
@@ -565,7 +567,7 @@ pub async fn create_wallet_with_master_secret(config: &WalletConfig) -> VcxResul
     trace!("Created wallet with handle {:?}", wallet_handle);
 
     // If MS is already in wallet then just continue
-    holder::libindy_prover_create_master_secret(wallet_handle, settings::DEFAULT_LINK_SECRET_ALIAS)
+    holder::libindy_prover_create_master_secret(wallet_handle, &config.master_secret.clone().unwrap_or(settings::DEFAULT_LINK_SECRET_ALIAS.to_string()))
         .await
         .ok();
 
--- a/agents/rust/aries-vcx-agent/src/agent/init.rs
+++ b/agents/rust/aries-vcx-agent/src/agent/init.rs
@@ -62,6 +62,7 @@ impl Agent {
             storage_credentials: None,
             rekey: None,
             rekey_derivation_method: None,
+            master_secret: None,
         };
 
         create_wallet_with_master_secret(&config_wallet).await.unwrap();
--- a/aries_vcx/src/global/settings.rs
+++ b/aries_vcx/src/global/settings.rs
@@ -85,18 +85,22 @@ pub fn add_master_secret(wallet_handle: WalletHandle, value: &String) -> VcxResu
 pub fn get_master_secret(wallet_handle: &WalletHandle) -> VcxResult<String> {
     trace!("get_config_value >>> ");
 
-    MASTER_SECRETS
+
+    let ms = MASTER_SECRETS
         .read()
         .or(Err(VcxError::from_msg(
             VcxErrorKind::InvalidConfiguration,
             "Cannot read settings",
         )))?
         .get(wallet_handle)
-        .map(|v| v.to_string())
-        .ok_or(VcxError::from_msg(
-            VcxErrorKind::InvalidConfiguration,
-            format!("Cannot read wallet_handle from settings"),
-        ))
+        .map(|v| v.to_string());
+    match ms {
+        Some(value) => {
+            trace!("get_config_value <<< value: {}", value);
+            Ok(value)
+        }
+        None => Ok(DEFAULT_LINK_SECRET_ALIAS.to_string()),
+    }    
 }
 
 pub fn get_config_value(key: &str) -> VcxResult<String> {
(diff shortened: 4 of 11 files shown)


pomfar:
Hi Patrik,
the change was made to allow setting the master_secret_name to manage wallets created with a different value than the current one "main".

I made a mistake in renaming, I was referring to master_secret_name not at the master_secret. I can bring it back.
For the use of global storage, I agree that it should be avoided, but not knowing the code well enough I didn't find another solution.
The master_secret_name is provided in the wallet opening and then it is requested in the libindy_prover_create_credential_req and libindy_prover_create_proof, where there is no walletconfig.

Each master_secret_name must be associated with its wallethandle, but where can we keep this mapping?

thanks

Patrik-Stas:
I see
>  to manage wallets created with a different value than the current one "main".
that's a valid point.

It's late now, I'll get back to this in the morning (UTC)

aries_vcx/src/indy/wallet.rs
```diff
@@ -33,6 +33,8 @@ pub struct WalletConfig {
     pub rekey: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub rekey_derivation_method: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub master_secret: Option<String>,
```
Patrik-Stas, line 37: - This structure is currently shared for creating and opening wallets, however the `master_secret` would only be used with call `create_wallet_with_master_secret`
- Putting it into the structure would indicate that one wallet can only have 1 master secret. Yes, that's kind of the default right now anyway with the `DEFAULT_LINK_SECRET_ALIAS`, but since we are at it, let's not bake-in this "singleton-ism" into `WalletConfig` structure
- Instead I suggest to change signature of 
```
create_wallet_with_master_secret(config: &WalletConfig)
```
to 
```
create_wallet_with_master_secret(config: &WalletConfig, master_secret_alias: String)
```

aries_vcx/src/indy/proofs/prover/prover.rs
```diff
@@ -65,11 +65,13 @@ pub async fn generate_indy_proof(
         build_cred_defs_json_prover(wallet_handle, pool_handle, &credentials_identifiers)
         .await?;
 
+    let master_secret = settings::get_master_secret(&wallet_handle)?;
```
Patrik-Stas, line 68: - It seems that this is the only place where `master_secret_alias` is being read. Creating the global storage made the reading the `master_secret_alias` somewhat convenient this way, but global variables are always taxing codebase eventually.
- I suggest we remove `get_master_secret`, `set_master_secret`, and instead again just expand the signature of `generate_indy_proof` such that it must be injected by the caller of the method. 
   - In practice, this method is being called from method `build_presentation` in `protocols/proof_presentation/prover/states/presentation_request_received` and there's a few levels of calls, so this can be a bit cumbersome, you will have to add the parameter through number of methods.
   
This way `aries-vcx` will stay generic and will provide freedom in choice of `master_secret_alias`. To maintain the functionality, from `libvcx` you can for now simply keep injecting the original `DEFAULT_LINK_SECRET_ALIAS`

Patrik-Stas:
Well I got to review it earlier than expected :-) Left some comments, mostly I want to loose the globals. 

**Important note:** similarly like with https://github.com/hyperledger/aries-vcx/pull/668, I'd like to ask you for rebase on top of https://github.com/hyperledger/aries-vcx/pull/648.

pomfar:
I saw the branch of #648 and it seems clear to me.

I agree to remove the globals, but it's not entirely clear to me how it should be done.

We can add the _master_secret_id_ parameter to the **aries_vcx** calls and make it generic but then how do we handle _master_secret_id_ in **libvcx**?

Should you pass it when creating the wallet, by adding it as a parameter to _create_wallet_ or by adding a _create_wallet_with_master_secret_id_?
Then we store it internally or add it as a parameter in _send_credential_request_ and _generate_proof_, but that's not a nice solution.
If we store it how? Could we do this as a wallet record?


For **aries_vcx**, _master_secret_id_ is used twice:
- <code>/aries_vcx/src/common/proofs/prover/prover.rs:generate_indy_proof</code>, which is called from <code>/aries_vcx/src/handlers/proof_presentation/prover.rs:generate_presentation</code>, ok
- <code>/aries_vcx/src/protocols/issuance/holder/state_machine.rs:create_credential_request</code> called from <code>/aries_vcx/src/protocols/issuance/holder/state_machine.rs:send_request</code> which is called from <code>/aries_vcx/src/handlers/issuance/holder.rs:send_request</code>, ok, 
but also from <code>/aries_vcx/src/protocols/issuance/holder/state_machine.rs:handle_message</code>.
In this method I don't think it's possible to add _master_secret_id_ parameter, right?

thanks for the help!

Patrik-Stas:
> but also from /aries_vcx/src/protocols/issuance/holder/state_machine.rs:handle_message.
In this method I don't think it's possible to add master_secret_id parameter, right?

It's possible, you will have to extend `CredentialRequestSend(String)` variant of `CredentialIssuanceAction` by new value, turning it into smthng like `CredentialRequestSend(String, String)` - this "Action enumeration" is old approach of interacting with state machines. We are planning to get rid of it, but for now you'd pass the value this way.

> We can add the `master_secret_id` parameter to the aries_vcx calls and make it generic but then how do we handle master_secret_id in libvcx?
Should you pass it when creating the wallet, by adding it as a parameter to create_wallet or by adding a create_wallet_with_master_secret_id?
Then we store it internally or add it as a parameter in send_credential_request and generate_proof, but that's not a nice solution.


Well, if you want to use custom values of `master_secret_alias` (or perhaps manage multiple `master_secret_alias`-es for 1 wallet), you will either way have to store it in some persistent storage. So now it's just question of whether we want to let user specify it during wallet opening, or at the point of using the secret (`send_credential_request` and `generate_proof`) - I am strongly for the second option, adding optional parameter to these methods on libvcx layer - if not specified, assume current default alias, if specified, use that. 

> Should you pass it when creating the wallet, by adding it as a parameter to create_wallet or by adding a create_wallet_with_master_secret_id?
- As for the creation of a new `master_secret` with certain alias, the best approach here I think would be adding a new FFI method `vcx_create_master_secret( <alias> )` - it's generic approach and also:
- This way there's no need to modify `WalletConfig`, which is in `aries-vcx`, to enable this `libvcx` use-case. 


I know this is a bit more work as it involves tweaking FFI, but we've learned a lesson about globals the hard-way and it took a lot of effort to remove them

gmulhearn:
hey @Patrik-Stas , Noticed this PR and thought it might be a good place to raise awareness - there's a push in Hyperledger to transition to using the term "link secret" as it fits in line with the anoncreds spec, whereas the term "master secret" is a depreciated term used by indy. I wonder if this PR or another in the future would be a good chance for aries-vcx to adopt the term "link secret".

Might be good to think about especially if we're adding a new FFI function which might need to change later (`vcx_create_master_secret` -> `vcx_create_link_secret`) (if aries-vcx wants to move towards the new term).

glossary in anoncreds spec:
https://hyperledger.github.io/anoncreds-spec/#term:link-secret

pomfar:
ok
I'll add the optional parameter **link_secret** to <code>send_credential_request</code> and <code>generate_proof</code>.
For the creation of the wallet I think it might be better to use the same approach, i.e. add the optional parameter.
Adding the new method <code>vcx_create_link_secret</code>  would require that it be called before create, otherwise the secret is not set.
I think this might create errors in usage.

Patrik-Stas:
Hi @pomfar, this has been opened for a while. Do you plan to further dedicate effort on this, or shall we close this PR?

@gmulhearn thanks for heads up on the terminology. We should definitely align with the anoncreds vocabulary.

pomfar:
Hi,
sorry but I can not at the moment, you can close it.

PR #668, opened 2022-11-28
revealed parameter in prover building proof

pomfar:
Signed-off-by: Pompeo Faruolo <pompeo.faruolo@etuitus.it>


Patrik-Stas:
hi @pomfar can you please leave write up some description about these changes? Is this in order to be more explicit, or does this have any functional effect?

Thank you.

pomfar:
Hi 
the change is a fix in the generating of the proof response, where the parameter unrevealed was ignored and set to true.
Then the proof response always revealed all attribute values, even when revealed parameter in input was false.

Patrik-Stas:
Hi @pomfar, this looks reasonable. 👍 

I've never used `revealed: false`; if prover specifies this, what happens? I suppose this way the prover only proves ownership of the attribute, but the value?

Before we merge this, I would like to ask you for a favor - can you please rebase this on top of https://github.com/hyperledger/aries-vcx/pull/648 ? This is currently sort of top priority PR and @gmulhearn-anonyome has been already rebasing many times, I hope you understand. In that PR, the files you have modified has been shuffled a bit. Please reach out to @gmulhearn-anonyome on discord if it turns our tricky and a guidance is needed. 

Thank you!

pomfar:
Hi
Yes, with revealed=false the values is not sent to the verifier.
This is a very powerful feature!

For the rebase, it's ok.

thanks

aries_vcx/src/common/proofs/prover/prover_internal.rs
```diff
@@ -114,6 +115,11 @@ pub fn credential_def_identifiers(credentials: &str, proof_req: &ProofRequestDat
 
                 let tails_file = value["tails_file"].as_str().map(|x| x.to_string());
 
+                let mut revealed = true;
```
Patrik-Stas, line 118: This block of added code can be done more safer, simpler:
```
let revealed = value["credential"]["cred_info"]["revealed"].as_bool().unwrap_or_default(true);
```
Although your code seems fine, this 1 liner is less error prone for 3 reasons:
- no `mut`
- no risk of mismatching strings in conditions and assignment, no duplication
- no panicking `unwrap()`

aries_vcx/src/common/proofs/prover/prover_internal.rs
```diff
@@ -15,6 +15,7 @@ pub struct CredInfoProver {
     pub revocation_interval: Option<NonRevokedInterval>,
     pub tails_file: Option<String>,
     pub timestamp: Option<u64>,
+    pub revealed: bool
 }
 
 pub async fn build_schemas_json_prover(
```
Patrik-Stas, line 21: bonus points to add test checking that `build_schemas_json_prover` produces `revealed: true` by default and `revealed: false` if overriden

Patrik-Stas (changes requested):
Hi @pomfar sorry for taking long time to merge this, we can speed things up now that the big PR from George has been merged into 0.50.0.
Just small change requested in the comment and this can be merged.

aries_vcx/src/common/proofs/prover/prover_internal.rs
```diff
@@ -124,6 +127,7 @@ pub fn credential_def_identifiers(credentials: &str, proof_req: &ProofRequestDat
                     rev_reg_id,
                     cred_rev_id,
                     tails_file,
+                    revealed: revealed,
```
Patrik-Stas, line 130: syntax allows you to write simply 
```
revealed
```
here instead of 
```
revealed: revealed
```

Patrik-Stas (approved):
I think you misunderstood my previous comment a little, the `CredInfoProver`'s `revealed` did not have to `Option`, instead it'd be better if line 118 `credential_def_identifiers` defaulted to true
```
let revealed = value["credential"]["cred_info"]["revealed"].as_bool().unwrap_or(true);
```
Now injecting default value is deferred to latter step `build_requested_credentials_json`. It's no biggie, I'll merge it as it is. If you'd like, you can refactor in separate PR.

PR #670, opened 2022-11-28
Remove ios wrapper indysdk

left-arm:


Patrik-Stas:
@PaulRps @pomfar @guijd3p @kukgini any objections removing these functions? I think these won't be missed, but let me know if otherwise is the case

PaulRps:
IMO it's not a problem to remove IndySdk (.h/m) files because there is no usage of their functions. However, it needed to make sure all references were removed and dependencies were resolved.

PR #675, opened 2022-12-01
Implement a basic connection test using Actix framework

mirgee:
This was written purely to explore the Actix framework and its viability in aries-vcx integration tests.

As you can see, it can be used, but the benefits outside of perhaps saving a few lines per integration test are unclear. On the other hand, there are glaring drawbacks:
* async code in `handle` method [can't use](https://github.com/actix/actix/issues/308) (or at least trivially) the reference to `self` and thus a necessity for cloning in the handler, or
* having to clone the agent in the tests (since the `start()` method consumes the agent), or
* having to store handled messages in the agent itself (unlike when using a channel with a capacity).

In addition, it seems far from certain that actix would stay maintained for long. Here is a citation from the actix docs:
> Long ago, Actix Web was built on top of the actix actor framework. Now, Actix Web is largely unrelated to the actor framework and is built using a different system. Though actix is still maintained, its usefulness as a general tool is diminishing as the futures and async/await ecosystem matures. At this time, the use of actix is only required for WebSocket endpoints.

Again, this was just an exploration, so if you don't think the benefits outweigh the costs, feel free to close this PR.

### DIFF
--- a/messages/Cargo.toml
+++ b/messages/Cargo.toml
@@ -33,3 +33,4 @@ strum = "0.16.0"
 strum_macros = "0.16.0"
 derive_builder = "0.10.2"
 tokio = { version = "1.15.0" }
+actix = "0.13.0"
--- a/messages/Cargo.toml
+++ b/messages/Cargo.toml
@@ -33,4 +33,3 @@ strum = "0.16.0"
 strum_macros = "0.16.0"
 derive_builder = "0.10.2"
 tokio = { version = "1.15.0" }
-actix = "0.13.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,51 @@
 version = 3
 
 [[package]]
+name = "actix"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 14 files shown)


messages/Cargo.toml
```diff
@@ -33,3 +33,4 @@ strum = "0.16.0"
 strum_macros = "0.16.0"
 derive_builder = "0.10.2"
 tokio = { version = "1.15.0" }
+actix = "0.13.0"
```
Patrik-Stas, line 36: Was wondering if we could drop this from `messages` and turns out yes
https://github.com/hyperledger/aries-vcx/pull/680

Patrik-Stas:
> As you can see, it can be used, but the benefits outside of perhaps saving a few lines per integration test are unclear. On the other hand, there are glaring drawbacks:

I think this PR is within the space of primary goal to discover the best approach for writing non-mediated agent2agent tests we've been waiting for for so long. Both this, and the test `async-channel` are very exciting to see :-) There can probably be some other approaches (opening a TCP port and listening for message there comes to mind - though I am not sure if this would have any benefits over the 2 demonstrated approaches)

I am not sure what's the best approach but we should pick something and then start eliminating mediation from tests. Let's leave some time to think about it, in the meantime let's keep this PR open.

PR #676, opened 2022-12-01
Handle did:sov service omitting routing_keys attribute

guijd3p:
when performing a test with acapy a routing_keys problem was discovered, it generates an empty ledger.


Patrik-Stas:
Awesome, great to hear you've tested with acapy :-) 

Yeah in general `.unwrap()` ideally shouldn't exist in the codebase (expect for tests), I've missed this in the review. 

While we are at it, if you want to earn some bonus points, it'd be cool if you'd add test which test which would have catched the problematic scenario (eg `routing_keys` missing).

Patrik-Stas (approved):

Patrik-Stas (approved):

PR #684, opened 2022-12-02
Replace failure crate with thiserror

mirgee:
Replaces the dependency on outdated `failure` crate with the dependency on `thiserror` in `messages`, `aries-vcx-agent`, `aries-vcx` and `libvcx`. Although our entire approach to logging needs to be reconsidered, removing a dependency on an unmaintained crate is likely not a step in the wrong direction.

It seems that we have to temporarily give up backtrace in libvcx errors stored in `CURRENT_ERROR_C_JSON` until `error_generic_member_access` is [stabilized](https://github.com/rust-lang/rust/issues/99301) since `thiserror` makes use of [unstable](https://doc.rust-lang.org/std/error/trait.Error.html#method.provide) `provide()` method to the error type in order to capture and share `std::backtrace::Backtrace`.


Patrik-Stas:
@mirgee awesome! But I please rebase on https://github.com/hyperledger/aries-vcx/pull/648 rather soon then later ;-) It's going to be the next going to be the next merge to `main` followed by `0.5.0.0` release right after

Patrik-Stas (approved):
It's also great that `indy-vdr` and `indy-shared-rs` are both using `thiserror`

gmulhearn-anonyome (approved):

Issue #694, opened 2022-12-12
Use `bs58` instead of `rust-base58`


Patrik-Stas:
https://crates.io/crates/bs58/0.4.0

>  Another Rust [Base58](https://en.wikipedia.org/wiki/Base58) codec implementation.
Compared to [base58](https://github.com/debris/base58) this is significantly faster at decoding (about 2.4x as fast when decoding 32 bytes), almost the same speed for encoding (about 3% slower when encoding 32 bytes), doesn't have the 128 byte limitation and supports a configurable alphabet.
Compared to [rust-base58](https://github.com/nham/rust-base58) this is massively faster (over ten times as fast when decoding 32 bytes, almost 40 times as fast when encoding 32 bytes), has no external dependencies and supports a configurable alphabet.

Main motivation is to simplify dependency tree as much as possible, this looks like low hanging fruit. Also 40x speedup for base58 de/encoding as cherry on top

PR #695, opened 2022-12-16
Replace rust-base58 by bs58

Patrik-Stas:
Address https://github.com/hyperledger/aries-vcx/issues/694

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,12 +89,12 @@ version = "0.50.0"
 dependencies = [
  "android_logger",
  "async-trait",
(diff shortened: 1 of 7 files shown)


Patrik-Stas:
It'd be also nice to address the de-duplication (and there's more than what seen in diff) Perhaps we could create 1 more crate for things like this

gmulhearn (commented):
~~Looks good. There's still a rust-base58 dependency in the messages crate, is that ok?~~ nvm sorry, bad diff

gmulhearn (approved):

PR #703, opened 2022-12-21
Code formatting

Patrik-Stas:
- Fail CI if code is not correctly formatted
- Adjusted `rustfmt.toml` to bit more liberal setup from its default `max_width=100` to `max_width=120` - however given `use_small_heuristics` https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#use_small_heuristics most of the lines will be well below 120 characters.
- Applied reformat


gmulhearn (approved):
lgtm!

PR #702, opened 2022-12-20
Refactor error handling

Patrik-Stas:
Is addressing https://github.com/hyperledger/aries-vcx/issues/582

changes:
- Removed dead code
- Previously there was `VcxError` defined in `aries-vcx`, this was used by both `aries-vcx` and `libvcx`. Because `libvcx` maps needs to map error types to u32, encoding them for FFI interface, this was part of `aries-vcx` - but obivously shouldn't. So to fix this:
   - `aries-vcx` still defined `VcxError`, but stripped of to minimum needed - mainly removed `u32` mapping etc.
   -  in `libvcx` was created module `errors` which defines `LibvcxError` which includes `u32` mappings
   - to bridge `VcxError` with `LibvcxError`, there's mapping `impl From<VcxErrorKind> for LibvcxErrorKind` in  `mapping_from_ariesvcx.rs` 
- In general, the `errors.rs` files has also been broken down into smaller pieces, typically into files such as `errors/mapping_from_<crate>` which maps errors from `<crate>` to domain errors of the given crate (such as `mapping_from_ariesvcx.rs` mentioned earlier)
- All our crates now consistently contain module `src/errors` 
- Some methods in `api_c` contained more logic than they should - the logic has been moved to `api_handle` layer instead (for example the entire file `api_handle/wallet.rs` has been extracted from code in `api_c` and shifted layer down, to `wallet.rs`). This is good step to separate `api_handle` and `api_c` into separate crates https://github.com/hyperledger/aries-vcx/issues/700
- libvcx: change the way u32 error codes and `LibvcxErrorKind` are associated - there's static vector of `(LibvcxErrorKind, u32)` tuples. This way, adding new error code is minimal effort, as this would be the only single place where mapping is handled - as opposed to previously defined `Kind->u32` and `u32->Kind` mapping declarations, which was quite error prone.
- Additionally renamed all error and error kinds, so there is:
   - libvcx: `ErrorLibvcx`, `ErrorLibvcx`
   - aries-vcx: `ErrorAriesVcx`, `ErrorKindAriesVcx`
   - messages: `ErrorMessages`, `ErrorKindMessages`
   - agency client: `ErrorAgencyClient`, `ErrorKindAgencyClient`

Issue #697, opened 2022-12-17
Simplify `messages` crate


Patrik-Stas:
The crate should be purely data model of aries messages and its construction. The crate should have minimal number of dependencies.
Some of the obvious action steps:
- move DID DOC stuff away (this could well be a separate crate)
- `ServiceResolvable` should be moved to `aries-vcx`
- review crate dependencies

PR #704, opened 2022-12-21
Refactor diddoc, extract as crate, create crate vcx_shared

Patrik-Stas:
Note: The commits in these PR are quite well separated, might we worthy to review commit-by-commit approach

This PR is further addressing https://github.com/hyperledger/aries-vcx/issues/697

Changes:
- Aries/w3c diddoc divergence
   - Added `W3cDidDoc` - this reflects https://www.w3.org/TR/did-core/#example-30-did-document-with-1-verification-method-type but is not currently used anywhere as there's a slight divergence due aries `service` extensions.
   - Renamed original `DidDoc` to `AriesDidDoc` 
   - The difference between the 2 is most of all `service` attribute - the rendered diddoc resolved by `did:sov` has an extra attributes like `recipientKeys`, `routingKeys` etc. (see `Resolver DID Document Format` @ https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html)
   - Long term we should use `W3cDidDoc` or find better abstraction to fit in the aries stuff
   
- Renamed `ServiceResolvable` - which sounds like and interface but is enum, to `ServiceOob`. This represents either `AriesService` or a DiD, for which AriesService can be resolved from ledger per OOB spec https://github.com/hyperledger/aries-rfcs/blob/main/features/0434-outofband/README.md#messages
   > services - an array of union types that the receiver uses when responding to the message. Each item is either a DIDComm service object (as per [RFC0067](https://github.com/hyperledger/aries-rfcs/blob/main/features/0067-didcomm-diddoc-conventions/README.md#service-conventions)) or a DID (as per [Decentralized Identifiers v1.0](https://w3c.github.io/did-core/#did-syntax)). Additional details below.
   
- Moved `EndpointDidSov` under `aries-vcx/common/ledger` - it has nothing to do with Aries messages, eventually should end up in some DID DOC resolver module
- Added optional field `types` which was missing in `EndpointDidSov` (as specified in `DID Service Endpoint` @ https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html#crud-operation-definitions )
- Extracted all diddoc related code to separate crate - as by itself is more general concept and has no relation to `aries`
- As result, I also ended up creating `vcx_shared` crate to avoid yet another duplication of methods `validate_did` and `validate_verkey`

Built on top of https://github.com/hyperledger/aries-vcx/pull/702


aries_vcx/src/common/ledger/service_didsov.rs
```diff
@@ -1,14 +1,30 @@
...
-// Note that is divergence from w3c spec https://w3c.github.io/did-core/#service-properties
-#[derive(Debug, Deserialize, Serialize, Clone,PartialEq)]
+// https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html
+#[derive(Debug, Deserialize, Serialize, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 pub struct EndpointDidSov {
     pub endpoint: String,
     #[serde(default)]
-    #[serde(rename = "routingKeys")]
+    #[serde(rename = "camelCase")]
```
gmulhearn, line 11: will this behave correctly? i.e. will it be literally `"camelCase"`

gmulhearn (changes requested):
looks good, just the one comment i wasn't sure about

aries_vcx/src/common/ledger/service_didsov.rs
Patrik-Stas: uhh this was great catch, signifying importance of de/serialization tests - these has now been added

gmulhearn (dismissed):

Patrik-Stas:
gmulhearn thanks for the review! 

Rebased and merging

Issue #699, opened 2022-12-17
Implement issuer/verifier profiles traits


Patrik-Stas:
- Implement issuer's side for interfaces created in https://github.com/hyperledger/aries-vcx/pull/648
- The ultimate goal is to completely migrate away from `vdr-tools` in favor of `indy-vdr` and `indy-credx`. 
- It might make sense however to extract out implementation of sqlite+mysql wallets from `vdr-tools` and enable using that going forward.

PR #708, opened 2022-12-23
CredxAnoncreds verifier functionality support

gmulhearn:
#699 Implements credx verifier method and tests

* Add new error `ProofRejected` to map to credx and vdrtools errors
* minor fix to logging in setupprofile
* changed tests as possible to use a dynamic profile (not locked to indy) for verifier functionality
* ^ this meant using `holder_setup` profile, after the work for credx issuer APIs are complete, ALL tests should be able to use `SetupProfile::run` rather than `SetupProfile::run_indy`


libvcx/src/errors/mapping_from_ariesvcx.rs
```diff
@@ -79,6 +79,7 @@ impl From<AriesVcxErrorKind> for LibvcxErrorKind {
             AriesVcxErrorKind::VdrToolsError(num) => LibvcxErrorKind::LibndyError(num),
             AriesVcxErrorKind::NoAgentInformation => LibvcxErrorKind::NoAgentInformation,
             AriesVcxErrorKind::InvalidMessageFormat => LibvcxErrorKind::InvalidMessageFormat,
+            AriesVcxErrorKind::ProofRejected => LibvcxErrorKind::LibndyError(405), // AnoncredsProofRejected Vdrtools error code
```
Patrik-Stas, line 82: Let's also create new `LibvcxErrorKind::ProofRejected`

aries_vcx/src/common/proofs/verifier/verifier.rs
```diff
@@ -195,17 +193,17 @@ pub mod unit_tests {
                 .await
                 .unwrap();
             assert_eq!(
-                validate_indy_proof(&setup.profile, &prover_proof_json, &proof_req_json)
+                validate_indy_proof(&holder_setup.profile, &prover_proof_json, &proof_req_json)
                     .await
                     .unwrap_err()
                     .kind(),
-                AriesVcxErrorKind::VdrToolsError(405)
+                AriesVcxErrorKind::ProofRejected
```
Patrik-Stas, line 200: 👍  just run into this as well while improving err handling a bit in https://github.com/hyperledger/aries-vcx/pull/721 and though you might have as well by what you mentioned on the last call - and yeah here it is :-)

Patrik-Stas: After this PR and #721, there won't be really explicit usage of `VdrToolsError` in aries-vcx or vdrtools - just will be left as variant for vdrtools errors which don't really need to be mapped to something (bunch of which should probably be removed)

gmulhearn: Thanks! reviewing 721 now, let's get that in first and I'll rebase and finish up this MR

Patrik-Stas (approved):

mirgee (approved):

Issue #700, opened 2022-12-17
Split libvcx crate


Patrik-Stas:
Split libvcx in 2 crates
- `libvcx_core` - this would be today's "handle" layer of `libvcx`
- `libvcx` - this would be today's `api_c` later of `libvcx`, eg built on top of `libvcx_core` crate

This will enable building different FFI approaches on shared base layer, ruling out a chance of dealing with some c symbol conflicts. For example we can build new `libvcx_napi` on top of `libvcx_core` as initiated in this PR https://github.com/hyperledger/aries-vcx/pull/665 
Similarly per @dinbtechit question https://github.com/hyperledger/aries-vcx/discussions/669 this would enable writing new wrappers, like flutter, with relative ease.

Implementation of this should have no breaking changes and the split should be invisible to the currently built `libvcx` dynamic library

Patrik-Stas:
libvcx was not split, but separated in 2 submodules `api_c`, `api_vcx`, see https://github.com/hyperledger/aries-vcx/pull/710

PR #710, opened 2022-12-27
Libvcx/refactor

Patrik-Stas:
- Divides `libvcx` in 2 main `api_c` and `api_vcx` modules which could be potentially split out as separate crates https://github.com/hyperledger/aries-vcx/issues/700
- `api_c` depends on `api_vcx` and contain libvcx c interface
- `api_vcx` doesn't know about `api_c` and contains the internal logic
- This should make implementation on https://github.com/hyperledger/aries-vcx/pull/665 smoother
- `errors` module was moved under `src/errors` to be consistent with other crates (previously under `src/api_lib/errors`)
- Defined build feature `libvcx_c` - if specified, `api_c` is included in the build


gmulhearn (approved):
lgtm

PR #711, opened 2022-12-29
Libvcx/minimize api c

Patrik-Stas:
Improvements in libvcx triggered working on https://github.com/hyperledger/aries-vcx/pull/665

- make feature `libvcx_c` default
- minimize responsibility of `api_c` - move logic of reading `CONFIG_INSTITUTION_DID` from `api_c` to `api_vcx`
- change FFI signature of `vcx_credentialdef_create_v2` by removing its `issuer_did: *const c_char` argument - this is inconsistent with all the other APIs which load the did from setting `CONFIG_INSTITUTION_DID` - technically a breaking change, but holistically the argument didn't have usecase, nobody should be impacted.
- shorten test `test_release_all` 
- do not return unnecessary `error::SUCCESS_ERR_CODE` codes from `api_vcx` layer - this was a leftover of past where today's `api_c` and `api_vcx` were tied much closer together
- all `update_state*` methods in `api_vcx` return new state on success
- reuse state value returned by `update_state` in `update_state*` methods in `api_c`


libvcx/src/api_vcx/api_handle/mediated_connection.rs
```diff
@@ -220,48 +221,39 @@ pub async fn handle_message(handle: u32, message: &str) -> LibvcxResult<u32> {
...
-    let res = if connection.is_in_final_state() {
+    if connection.is_in_final_state() {
         info!(
             "connection::update_state >> connection {} is in final state, trying to respond to messages",
             handle
         );
         let profile = get_main_profile_optional_pool(); // do not throw if pool is not open
-        match connection
+        connection
             .find_and_handle_message(&profile, &get_main_agency_client().unwrap())
```
gmulhearn, line 236: unrelated to this PR, but is there any issues with these `unwrap`s? i.e. is panicking dangerous for the iOS/android/node consumers?

wrappers/ios/vcx/VcxAPI.m
```diff
@@ -835,22 +835,19 @@ - (int)vcxRevocationRegistryRelease:(NSNumber *)revRegHandle {
 
 - (void)vcxCredentialDefinitionCreateV2:(NSString *)sourceId
                                schemaId:(NSString *)schemaId
-                              issuerDid:(NSString *)issuerDid
```
gmulhearn, line 838: just double checking - is there any updates needed to the android/java wrapper APIs?

gmulhearn (approved):
Looks good, just had one comment and one unrelated question

libvcx/src/api_vcx/api_handle/mediated_connection.rs
Patrik-Stas: Good catch - there should be no `unwraps` anywhere in the code (with exception of test code). The unwrap() on the line (and few similar I found now in `api_handle` module) are unjustified. 

I am not sure if perhaps different wrappers react differently on panic. Afaik in the old (current) node wrapper it results in exception on nodejs. But it can have more consequences, if you happen to own a mutes/rwlock lock while panicking, it can cause issues with the subsequent calls as the lock become poisoned. 

So definitely we want to avoid any panic, generally the application should shut down on panic, similarly to `Error` in Java

wrappers/ios/vcx/VcxAPI.m
Patrik-Stas: as far as I am aware, all places in wrappers has been updated respectively

Patrik-Stas:
@gmulhearn thanks for review! Please go ahead and merge (Squash and merge) if you are okay with my responses above :-)

gmulhearn:
Hey @Patrik-Stas , looks good. I would merge however github won't let me: 
image

haven't seen this before... is it a glitch?

Patrik-Stas:
@gmulhearn ahh I see. I changed laptop and haven't setup gpg commit signing which is set as required on the repo

Patrik-Stas:
@gmulhearn signed all commits, should work now

gmulhearn (approved):

PR #713, opened 2022-12-30
Clippy do not allow unwrap used

gmulhearn:
Denies clippy rule defined here: https://rust-lang.github.io/rust-clippy/master/#unwrap_used

ideally we should be not panicking at all (unless testing). however in some places where it's obvious an error will not occur (e.g. some `lazy_static` variable blocks), we will use `.expect(...)` instead - knowing that the `expect` should never occur.

Note that also in some places, where we know it is only used for testing (e.g. mocks) `expect`s are used - this may be changed in the future if desired.


gmulhearn:
@Patrik-Stas  let me know what you think about this potential change. It might be overkill, but it at least ensures we don't have any potentially dangerous `unwrap`s slip thru. 

By default, clippy allows this lint to pass thru without warnings, so I guess they are making a statement that it _may_ be overkill to have this check. See here; https://rust-lang.github.io/rust-clippy/master/#unwrap_used

But up to you - I'm personally a fan of this enforcement

Patrik-Stas (approved):

Patrik-Stas:
@gmulhearn this is great improvement! And I absolutely agree with `deny` lint

Merging

PR #715, opened 2023-01-03
Remove public agent

mirgee:
Removes any and all usage of public agents in the codebase. In Rust integration tests where connection bootstrapping was previously performed via a public agent, an alternative approach was temporarily (i.e. until mediated connection is removed) chosen: the connection request message is sent / received via a channel instead of a DidComm endpoint to avoid overhead associated with managing an open TCP socket per test in Rust. However, vcx-agent-core integration tests do exchange connection requests via DidComm endpoint.


Patrik-Stas (approved):
Though still in draft, preliminary approved from my side, looks great! 

Just list of changes to the PR, mainly changes to `aries-vcx` APIs

Patrik-Stas (approved):

PR #717, opened 2023-01-04
Nonmediated connection (de)serialization

mirgee:
Exposes (de)serialization methods on nonmediated connection handler.


aries_vcx/src/utils/mockdata/mockdata_connection.rs
```diff
@@ -1,409 +1,543 @@
...
-    ],
-    "serviceEndpoint": "http://localhost:8080/agency/msg"
-}"#;
-
-// Alice created and serialized connection created from received invitation
 pub const CONNECTION_SM_INVITEE_INVITED: &str = r#"
 {
-    "version": "1.0",
-    "source_id": "alice-670c6360-5c0e-4495-bd25-2ee58c39fc7e",
-    "thread_id": "b5517062-303f-4267-9a29-09bc89497c06",
```
Patrik-Stas, line 22: Although I see `thread_id`and `source_id` defined `SmConnectionInvitee`, this information seems to be not included in serialized Connection. I think we should, especially the `thread_id`.
I am not fan of the existence of `source_id`, but given it's everywhere else, probably it should be included equally here as well. 
We could perhaps remove `source_id` across the board in separate PR in the future.

aries_vcx/src/protocols/connection/invitee/state_machine.rs
```diff
@@ -23,7 +23,7 @@ use messages::protocols::connection::request::Request;
 use messages::protocols::connection::response::SignedResponse;
 use messages::protocols::discovery::disclose::{Disclose, ProtocolDescriptor};
 
-#[derive(Clone)]
+#[derive(Clone, Serialize, Deserialize)]
 pub struct SmConnectionInvitee {
     source_id: String,
     thread_id: String,
```
Patrik-Stas, line 29: as per my other comment, it seems these fields are not being included in serialized form of the Connection

Patrik-Stas (changes requested):
left some comments

Patrik-Stas (approved):

PR #734, opened 2023-01-12
Remove generic UrsaError err mapping, handle possible UrsaErrors explicitly

Patrik-Stas:
- Remove `impl From<UrsaCryptoError> for AriesVcxError` mapping - it seemed confusing, because it was not clear which of the error conditions can actually / in what circurmstance happen / where are ursa error thrown. In fact, most of the UrsaErrorKinds were never thrown. It turned out there's only 3 places where we deal with Ursa directly, and those errors are always about dealing with `BigNumber` conversions.
- So instead of the generic handling covering cases which don't happen, I've handled ursa errors directly on spot and converted them to new error kind `UrsaError`


aries_vcx/src/plugins/anoncreds/credx_anoncreds.rs
```diff
@@ -56,7 +56,12 @@ impl IndyCredxAnonCreds {
 
         let ms_value = (&record).try_get("value")?;
         let ms_decimal = ms_value.try_as_str()?;
-        let ms_bn: BigNumber = BigNumber::from_dec(ms_decimal)?;
+        let ms_bn: BigNumber = BigNumber::from_dec(ms_decimal).map_err(|err| {
+            AriesVcxError::from_msg(
+                AriesVcxErrorKind::UrsaError,
+                format!("Failed to create BigNumber from value: {}, error: {}", ms_decimal, err),
```
gmulhearn, line 62: We're dealing with link secrets here, which are pretty much private keys, we probably want to be careful to not leak any of these values in error messages. in retrospect, i'd probably be cautious about leaking the stringified error too (incase that error string leaks the value). So maybe all of these error messages should not contain any variables, and just be constant strings. What do you think?

Patrik-Stas: good points, let me minimize content of these message

gmulhearn (approved):
lgtm!

mirgee (approved):

PR #738, opened 2023-01-19
Expose nonmediated connection in node wrapper

mirgee:


wrappers/node/src/api/connection.ts
```diff
@@ -0,0 +1,174 @@
...
+export type INonmediatedConnectionInvite = string;
+
+export type INonmeditatedConnectionData = string;
+
+export interface IEndpointInfo {
+  serviceEndpoint: string,
+  routingKeys: string[],
+}
+
+export class NonmediatedConnection extends VcxBaseWithState<INonmeditatedConnectionData, ConnectionStateType> {
```
Patrik-Stas, line 16: I guess going forward we can remove existing `Connection` and rename `NonmediatedConnection -> Connection`, agree?

wrappers/vcx-napi-rs/Cargo.toml
```diff
@@ -1,9 +1,11 @@
 [package]
 name = "vcx-napi-rs"
-version.workspace = true
-authors.workspace = true
-license.workspace = true
-edition.workspace = true
+version = "1.0.0"
+edition = "2021"
+# version.workspace = true
```
Patrik-Stas, line 5: not sure about the commented config here

wrappers/vcx-napi-rs/dist/index.d.ts
```diff
@@ -0,0 +1,179 @@
+export const updateWebhookUrl: any;
```
Patrik-Stas, line 1: commited by mistake?

wrappers/vcx-napi-rs/index.d.ts
```diff
@@ -101,6 +120,7 @@ export function outOfBandReceiverCreate(msg: string): number
 export function outOfBandReceiverExtractMessage(handle: number): string
 export function outOfBandReceiverConnectionExists(handle: number, connHandles: Array<number>): Promise<number>
 export function outOfBandReceiverBuildConnection(handle: number): Promise<string>
+export function outOfBandReceiverBuildNonmediatedConnection(handle: number, serviceEndpoint: string, routingKeys: Array<string>): Promise<string>
```
Patrik-Stas, line 123: I understand this is mirroring the status quo, so no need to address it here, but: 

I think we should strive to invert this API - instead of calling something like this, the user should call just `Connection.createWithInvitation(oob)`, and in fact that should work even today I suppose. Perhaps this API is redundant to begin with.

wrappers/vcx-napi-rs/index.js
```diff
@@ -11,7 +11,8 @@ function isMusl() {
   // For Node 10
   if (!process.report || typeof process.report.getReport !== 'function') {
     try {
-      return readFileSync('/usr/bin/ldd', 'utf8').includes('musl')
+      const lddPath = require('child_process').execSync('which ldd').toString().trim();
```
Patrik-Stas, line 14: wondering, what's this?

agents/node/vcxagent-core/src/agent.js
```diff
@@ -34,8 +35,9 @@ async function createVcxAgent ({ agentName, genesisPath, agencyUrl, seed, wallet
   const agentProvision = await storageService.loadAgentProvision()
   const issuerDid = agentProvision.issuerConfig.institution_did
 
-  async function agentInitVcx () {
+  async function agentInitVcx() {
```
Patrik-Stas, line 38: Just dropping here as note, we should create issue to replace usage of `serviceConnections`  by `serviceNonmediatedConnections` everywhere. Then we can also drop agency client init code, starting up agency in nodejs tests

Patrik-Stas (dismissed):

wrappers/vcx-napi-rs/dist/index.d.ts
mirgee: Yes, `dist` file should be gitignored

wrappers/vcx-napi-rs/Cargo.toml
mirgee: This should not have been committed of course, made some changes to appease LSP

wrappers/node/src/api/connection.ts
mirgee: Yes - I decided to go for fairly consistent `NomediatedConnection` naming convention from to wrapper layer for purposes of backwards-compatibility, but also to make it easy to rename in the future once the old, mediated connection handler is gone.

wrappers/vcx-napi-rs/index.js
mirgee: This code was auto-generated by napi-rs

wrappers/vcx-napi-rs/index.d.ts
mirgee: We are in absolute agreement on this point.

agents/node/vcxagent-core/src/agent.js
mirgee: By "everywhere", I assume you mean the `vcxagent-core` integration tests (the current Alice and Faber agents)? Although, since I think that you would agree that mediation should remain optional for `vcxagent-core` clients for a while, we should perhaps also leave at least some coverage for the mediated connection?

Issue #706, opened 2022-12-21
Implement DDO Resolver


Patrik-Stas:
- Design trait for DDO Resolver
- Native implementation for trait for `did:sov` did method
- Expect native implementation for for `did:indy` did method
- Consider adding non-native implementation for other methods via api https://dev.uniresolver.io/ 

Note: also need to consider how current divergence in `service` plays into this. The resolver itself should return DDO as per w3c spec, which specifies `service` certain way, illustrated by example https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html#crud-operation-definitions
```
{
  "service": [{
    "id":"did:example:123#linked-domain",
    "type": "LinkedDomains", 
    "serviceEndpoint": "https://bar.example.com"
  }]
}
```
On the other hand, working in aries domain, there's seems to be somewhat of an extension of DDO, where `service` object can have rendered extra fields such as `routing_keys`, `recipient_keys` etc.  as https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html#crud-operation-definitions illustrates on example 
```
"service": [
    {
      "type": "endpoint",
      "serviceEndpoint": "https://example.com/endpoint"
    },
    {
      "id": "did:sov:HR6vs6GEZ8rHaVgjg2WodM#did-communication",
      "type": "did-communication",
      "priority": 0,
      "recipientKeys": [
        "did:sov:HR6vs6GEZ8rHaVgjg2WodM#key-agreement-1"
      ],
      "routingKeys": [],
      "accept": [
        "didcomm/aip2;env=rfc19"
      ],
      "serviceEndpoint": "https://example.com/endpoint"
    }
  ]
  ```

PR #163, opened 2020-10-23
Revert rename of decrypted_payload to decrypted_msg in Message

Patrik-Stas:
This reverts breaking change introduced in https://github.com/hyperledger/aries-vcx/pull/134

Specifically the unintended breaking changes were:
- rename of `decrypted_payload` to `decrypted_msg`
- change in content format of `decrypted_payload`/`decrypted_msg` . Originally these were json strings such:
`{"@msg": <StringifiedAriesJsonMessage>}`, but the PR 134 changed this to simply `<StringifiedAriesJsonMessage>`

Signed-off-by: Patrik Stas <patrik.stas@absa.africa>

PR #739, opened 2023-01-24
Refactor/typestate connections

bobozaur:
Very basic draft implementation for a typestate Connection type. 

Note that the inner methods modifications are minimal for now, as the first step was to transpose the nested enums to typestate.


Patrik-Stas:
Drafted myself a small diagram while reviewing this, so I'll just drop it here

My thoughts:

- The new implementation is nice in terms of structuring. In yesterday private discussion we it was suggested that state type pattern would remove the need for the "handlers" layer but it seems to be that `Connection<I, T, S>` would become equivalent of the current handlers layer, eg  the current`impl Connection {` (`src/handlers/connection/connection.rs`). 

- My main problem with the new implementation is following:
   - With existing implementation, the presence of states is hidden internally behind `impl Connection` handler/facade. Altough this enables consumers of `Connection` to call wrong methods (and be punished with error :-)), it on the other hand makes the client code simpler - and especially in terms of de/serialization. Lets' take following example:
   ```
   let connection_serialized = database.load_connection_by_id("alice");
   let connection = Connection::from_string(connection_serialized);
   connection.some_operation()?; // we might get error if `some_operation` did not 
                                 // make sense in whatever state `connection` was at
   ```
   However, with the new approach, assuming `Connection<I, T, S>` would present the public API for aries-vcx consumers, it gets trickier, but let me try anyway: (note I just ignore the `transport / T` generic arg)
   ```
   let connection_serialized = database.load_connection_by_id("alice");
   let connection: Connection <Inviter, T, CompletedState> = Connection::from_string(connection_serialized);
   connection.some_operation(); // this is safer to call now, but we had to know the state ahead of time
   ```
   well that works, but the developer needs to already know if the serialized connection was inviter or invitee and what state is that connection at, as it's now required at compile time. Which seems kinda impractical. 
   
   Perhaps the state pattern can be still adopted internally, but `impl Connection<I, T, S> { }` would be internal model, and this would be wrapped by "handler" like the `impl Connection` in `/handlers` today.
   
- More of a side note: I like that this PR is addition of alternative implementation instead of mutating and disrupting existing code. The new `Connection<I, T, S>` structure would serialize obviously different and though possible to retrofit on old serialization format, I am not even sure if that would be worthy the effort. If this will move forward, I'd suggest continue in this fashion (perhaps even more structurally separate old/new approach), deprecate old style, let people migrate, delete old code. 


statepattern

bobozaur:
@Patrik-Stas That's a very good observation and yes, that's the disadvantage described by me as well.
However, the serialization/deserialization scenario can still be worked around by introducing helper structs and perhaps even a trait for making things easier for users.

An oversimplified example is:

```
// User code:
enum MyStoredConnection {
    Http(StorableConnection<HttpClient>),
    Websocket(StorableConnection<Websocket>)
}

// Library code:
pub enum StorableConnection<T> {
    Invitee(StorableInviteeConnection<T>),
    Inviter(StorableInviterConnection<T>)
}

pub enum StorableInviteeConnection<T> {
    Initial(CommonConnectionData<T>),
    Invited(CommonConnectionData<T>),
    Requested(CommonConnectionData<T>),
    Responded(CommonConnectionData<T>),
    Complete(CommonConnectionData<T>)
}

pub enum StorableInviterConnection<T> {
    Initial(CommonConnectionData<T>),
    Invited(CommonConnectionData<T>),
    Requested(CommonConnectionData<T>),
    Responded(CommonConnectionData<T>),
    Complete(CommonConnectionData<T>)
}

pub struct CommonConnectionData<T> {
    marker: PhantomData<T>
}

pub trait StoredConnections {
    type StorableConnection;

    fn new() -> Self;

    fn store_inviter_initial<T>(&mut self, con: Connection<Inviter, T, InitialState>) {}

    fn store_inviter_invited<T>(&mut self, con: Connection<Inviter, T, InvitedState>) {}

    fn store_inviter_requested<T>(&mut self, con: Connection<Inviter, T, RequestedState>) {}

    fn store_inviter_responded<T>(&mut self, con: Connection<Inviter, T, RespondedState>) {}

    fn store_inviter_complete<T>(&mut self, con: Connection<Inviter, T, CompleteState>) {}

    fn store_invitee_initial<T>(&mut self, con: Connection<Invitee, T, InitialState>) {}

    fn store_invitee_invited<T>(&mut self, con: Connection<Invitee, T, InvitedState>) {}

    fn store_invitee_requested<T>(&mut self, con: Connection<Invitee, T, RequestedState>) {}

    fn store_invitee_responded<T>(&mut self, con: Connection<Invitee, T, RespondedState>) {}

    fn store_invitee_complete<T>(&mut self, con: Connection<Invitee, T, CompleteState>) {}

    fn store_inviter<T>(&mut self, con: StorableInviterConnection<T>) {
        match con {
            Initial(con_data) => self.store_inviter_initial(con_data.into()),
            Invited(con_data) => self.store_inviter_invited(con_data.into()),
            Requested(con_data) => self.store_inviter_requested(con_data.into()),
            Responded(con_data) => self.store_inviter_responded(con_data.into()),
            Complete(con_data) => self.store_inviter_complete(con_data.into()),
        }
    }

    fn store_invitee<T>(&mut self, con: StorableInviteeConnection<T>) {
        match con {
            Initial(con_data) => self.store_invitee_initial(con_data.into()),
            Invited(con_data) => self.store_invitee_invited(con_data.into()),
            Requested(con_data) => self.store_invitee_requested(con_data.into()),
            Responded(con_data) => self.store_invitee_responded(con_data.into()),
            Complete(con_data) => self.store_invitee_complete(con_data.into()),
        }
    }

    fn store_connection<T>(&mut self, con: StorableConnection<T>) {
        match con {
            Invitee(invitee_con) => self.store_invitee(invitee_con),
            Inviter(inviter_con) => self.store_inviter(inviter_con)
        }
    }

    fn handle_storable_connection(&mut self, con: Self::StorableConnection);
}
```

The idea is that the `Connection<I, T, S>` would serialize to `StorableConnection` (can be done under the hood by serde) and users would just wrap it to provide transport distinction (not mandatory if only one transport is used throughout the agent - `StorableConnection<Transport>` can be used in this case). Deserialization would happen to the user defined type (or `StorableConnection<Transport>` and could be uniformly handled through a `StoredConnection` trait impl.

An implementor of `StoredConnection` can decide which connections it wants to store (as all default to no-op) and how to store them and use them throughout the rest of the code (could use some type-based `HashMap`, `Box<dyn Any>`, multiple `HashMap` or `Vec`, channels for sending the connections to other parts of the code, etc.). The only "mandatory" method to implement would be `handle_storable_connection` where users implement the transport distinguishing logic.

NOTE: An idea might be to return errors from the trait methods, but this was merely written as an example.

This is indeed more advanced but definitely not extreme.

aries_vcx/src/protocols/connection/invitee/state_machine.rs
```diff
@@ -242,17 +242,9 @@ impl SmConnectionInvitee {
...
-                    state.clone(),
-                    invitation,
-                    state.did_doc.ok_or(AriesVcxError::from_msg(
-                        AriesVcxErrorKind::InvalidState,
-                        "Expected none None state.did_doc result given current state",
-                    ))?,
-                )
-                    .into(),
-            ),
+            InviteeFullState::Initial(state) => {
```
Patrik-Stas, line 245: Can you please extract the changes related to cleaning up the initial/invited state and related APIs a small separate PR? 
For sake of better changelog, plus it's something that can be merged probably very soon

aries_vcx/src/protocols/typestate_con/invitee/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+            source_id,
+            state: InitialState,
+            pairwise_info,
+            initiation_type: Invitee,
+        }
+    }
+}
+
+impl InviteeConnection<InitialState> {
+    // This should take an Invitation, but that also implies a DDO resolver
```
Patrik-Stas, line 46: good observation :-) Once we have a DDO Resolve trait https://github.com/hyperledger/aries-vcx/issues/706 I think that could improve the function signature here to. I wouldn't like state machines to depend on particular ddo resolving code, so passing some DdoResolver trait along with the invitation could perhaps be the ultimate solution

Patrik-Stas (changes requested):
I like the further separation of old/new code! As you know, in the next step I'd mostly like to see usage of this in libvcx. I'd suggest there to keep similar separation between the old/new approach so we don't break libvcx wrappers.

aries_vcx/src/protocols/connection/invitee/state_machine.rs
bobozaur: In fact, I'm separating the code completely so that I don't interfere with current code. I'll revert the changes on these files, as initially I was trying to use the same state structs.

aries_vcx/src/protocols/typestate_con/invitee/mod.rs
bobozaur: That would indeed be the proper way to do it :-).

Patrik-Stas:
Hi @gmulhearn have a look at this, I am interested to hear your thought on this. Consider waiting until today's video recording is uploaded on https://wiki.hyperledger.org/display/ARIES/2023-01-26+Aries-vcx+community+call where @bobozaur goes through the approach taken here.

gmulhearn:
Hey @Patrik-Stas  sorry for delays, just got around to catching up on the recording and this PR now.

This is a very exciting POC 👌. 

@bobozaur  I share similar concerns around what ser/de in agents which store connections persistently. In the recording, I believe you mention how depending on the situation, you could assume that dev would know at compile time what the expected state is. But ocourse that's not always the case. E.g. imagine an ACA-py-like agent, where the agent persistently stores connections in permutations of different states, ACA-py fetches those connection records from DB without knowing at all what state permutation they are in - this is the most common pattern in DI agents from what I've been exposed to.

I know you've covered the ser/de concerns several times. However, maybe a good way for you to make the solution clearer is if you try to build onto the rust agent (within `agents/rust/aries-vcx-agent`) to store these new connection structs persistently (or perhaps even storing the string serealized connections to further drive home the solution).

gmulhearn:
Thinking out loud, but...

IMO, the main style of DI agents is where; 

1. an agent takes a step in a protocol, 2. stores the protocol state/handler in some persistent DB, 3. then waits for a new message 4. handles that message (which may involve fetching a protocol state/handler from the DB), 5. repeat.

so perhaps when you're at step 4 of the above, the type of message that you receive is enough to infer the exact type of the handler. e.g. if you receive a `ConnectionRequest` message, when you fetch the appropriate `Connection` from the DB, your 'connection request' handler code could infer that the type _should be_ `Connection<Inviter, InvitedState>`. So maybe it's generally not a problem to expect codebases to **know** the expected type of their serialized handler.

I do however share the concern about this increasing complexity for consumers. But I'm not sure how much of a concern that should be.. Rust is generally more complex, so you could argue that it's good we are taking advantage of rust features. Further, this concern can be mitigated with very good docco and some good code examples.

bobozaur:
@gmulhearn There's still work to be done in this, but the serialization/deserialization part would essentially happen like this:
 * Concrete connection type (e.g: `Connection<Inviter, Requested>`) -> Some intermediate type (used for alignment between serialization/deserialization) -> String: This is all under-the-hood. The user has access to call `Connection<Inviter, Requested>::serialize(serializer)` as you would with any type and serialize directly.
 * String -> `VagueConnection` (type that can deserialize any connection, exposed by the API) -> Concrete connection type (e.g: `Connection<Invitee, Complete>`): There's `TryFrom` implementations for all connection states to attempt conversion from `VagueConnection` if the state is the expected one. This would allow users that know they expect a certain connection state to deserialize into that directly.
 * String -> `VagueConnection` -> Runtime pattern matching -> Handle concrete connection type: Since the `VagueConnection` is exposed, users can take advantage of the type for things like storing all connections in a `Vec` or any sort of thing. They can then either use `TryFrom` or pattern matching on the state to handle concrete connection types based on their state.

Is this more complex than it used to be? Definitely. Is it worth it? I wholeheartedly believe it is. As you pointed out, Rust is more complex. Regardless of how user-friendly an API we expose, it will never be easier for people to develop a Rust agent than, for instance, a Python one (assuming a similar level of knowledge of the two languages).

Don't get me wrong, we should definitely make the library as user-friendly as possible, and you also pointed out other important aspects such as examples and documentation, but on the other hand nobody writes Rust because of how easy it is to code in. So I'd say that we should focus on the language's strengths, such as performance, safety and robustness at the cost of being somewhat more verbose and complex.

Another important idea that I also mentioned in the community call was that while this kind of feels like forcing users to handle and deal with certain aspects of the library, I'd say that we do that because they *should* do that anyway. We enforce the idea of being aware of the state a connection is in and handle it appropriately because... they should be aware.

We *make* them handle connection states separately because, again, they *should*. I'd argue that we're in fact helping them a bit by pointing out *"Hey, these connections do different things because of the different states. Handle them separately".*

@Patrik-Stas mentioned the same assumption that you pointed out regarding the `ConnectionRequest` example. You get a `ConnectionRequest` and you assume the connection will be in the invited state, but there's no *guarantee*. So with the current API you could try to handle the request, and progress the state machine if the state is the expected one or return an error if it isn't. Or, you check the state before even trying, and handle that instead.

With the proposed API you always *know* the state you should be in to take a specific action, and simply try to get into that state and handle whether that succeeded or not. After that it's all compile-time guarantees and a lot less pattern matching.

Regarding the example agent, we could probably do that too after I finish the `libvcx` implementation. You could look at that for an example of how you can take advantage of the `VagueConnection` in the general case and then attempt to convert a connection to the expected state in order to take a specific action.

gmulhearn:
Hey @bobozaur could you please let me know as well when this is in a final review-ready state (if not already). Would love to review these changes

bobozaur:
@gmulhearn Almost there. I still need to integrate some stuff from #163 for backwards compatibility with agents built on top of the wrapper. I'll update the PR when it is done.

aries_vcx/src/protocols/connection/generic/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+/// ``` ignore
+/// // Assume the `con` variable stores a `GenericConnection`:
+///
+/// let initial_connections = Vec::new();
+/// let completed_connections = Vec::new();
+///
+/// // Unwrapping after the match is sound
+/// // because we can guarantee the conversion will work
+/// match con.state() {
+///     State::Invitee(Stage::Initial) => initial_connections.push(con.try_into().unwrap()),
```
gmulhearn, line 58: Just a general comment to make sure you circle back to this docco - i.e. it should be `ThinState::Invitee(State::Initial)` I believe. other parts of this docco make need to be double checked as well

bobozaur: Good catch. It took a bit to settle on some naming 😅 .

Fixed.

bobozaur:
Necessary changes from #741 to get the wrapper working were done. I'd say this PR is in it's final state, so please feel free to review it.

aries_vcx/src/protocols/connection/mod.rs
```diff
@@ -1,3 +1,188 @@
...
+#[serde(bound = "(I, S): TryFrom<GenericState, Error = AriesVcxError>")]
+pub struct Connection<I, S> {
+    source_id: String,
+    pairwise_info: PairwiseInfo,
+    initiation_type: I,
+    state: S,
+}
+
+impl<I, S> Connection<I, S> {
+    pub(crate) fn from_parts(source_id: String, pairwise_info: PairwiseInfo, initiation_type: I, state: S) -> Self {
```
gmulhearn, line 46: Just leaving a comment here in-case we decide in tomorrow's meeting that we should make the method `pub`.

aries_vcx/src/protocols/connection/inviter/mod.rs
```diff
@@ -1,2 +1,235 @@
...
+
+        Ok(Connection {
+            source_id: self.source_id,
+            pairwise_info: new_pairwise_info,
+            initiation_type: self.initiation_type,
+            state,
+        })
+    }
+
+    pub fn get_invitation(&self) -> &Invitation {
```
gmulhearn, line 176: pulling at straws here; but it might be worth adding a doc that if `into_invited` was used to reach this state, and then `get_invitation` is called, the `Invitation` that is returned is meaningless and not representative of anything real

aries_vcx/src/protocols/connection/invitee/mod.rs
```diff
@@ -1,2 +1,228 @@
...
+            state,
+            source_id: self.source_id,
+            pairwise_info: self.pairwise_info,
+            initiation_type: Invitee,
+        })
+    }
+}
+
+impl InviteeConnection<Responded> {
+    /// Sends an acknolwedgement message to the inviter and transitions to [`InviteeConnection<Complete>`].
```
gmulhearn, line 185: typo; `acknowledgement`

aries_vcx/src/protocols/connection/inviter/mod.rs
```diff
@@ -1,2 +1,235 @@
...
+        &self.state.invitation
+    }
+}
+
+impl InviterConnection<Requested> {
+    /// Sends a [`Response`] to the invitee and transitions to [`InviterConnection<Responded>`].
+    ///
+    /// # Errors
+    ///
+    /// Will return an error if sending the repsonse fails.
```
gmulhearn, line 186: typo; `response`

libvcx/src/api_vcx/api_handle/connection.rs
```diff
@@ -42,62 +118,143 @@ where
...
+    add_connection(con)
 }
 
 // ----------------------------- GETTERS ------------------------------------
 pub fn get_thread_id(handle: u32) -> LibvcxResult<String> {
     trace!("get_thread_id >>> handle: {}", handle);
-    CONNECTION_MAP.get(handle, |connection| Ok(connection.get_thread_id()))
+
+    let lock = CONNECTION_MAP.read()?;
+    let con = lock.get(&handle).ok_or_else(|| {
```
gmulhearn, line 145: there's a bit of code duplication here. Could it be moved to a function similar to `get_cloned_connection`? something like:

```
fn get_borrowed_connection_from_lock<'a>(lock: &'a RwLockReadGuard<HashMap<u32, GenericConnection>>, handle: u32) -> LibvcxResult<&'a GenericConnection>
```

libvcx/src/api_vcx/api_handle/out_of_band.rs
```diff
@@ -175,6 +176,28 @@ pub async fn connection_exists(handle: u32, conn_handles: &Vec<u32>) -> LibvcxRe
     }
 }
 
+// todo: remove this
```
gmulhearn, line 179: ? is this comment still relevant

gmulhearn:
Just finished a pass thru and couldn't really find anything other than a few small things. Looks really good. I'm not too versed with the `napi` stuff, so I have no comments there.

libvcx/src/api_vcx/api_handle/connection.rs
bobozaur: Not sure what the benefit would really be. 

You'd still have to call `CONNECTION_MAP.read()` to get the lock guard, which must live for the entire duration of the operation, so we essentially only get rid of the `lock.get()` and replace it with `get_borrowed_connection_from_lock()`.

I don't really like the duplication either but in terms of ownership it has to behave like that. 
I'll look into using a function that takes a closure, or maybe a macro taking a token tree.

aries_vcx/src/protocols/connection/inviter/mod.rs
bobozaur: Good idea. Unfortunately `into_invited` is a necessary evil for backwards compatibility.

The way I see that this would ideally behave is that you generate an `Inviter`, call `create_invitation`, publish the invitation, and whenever you get a `ConnectionRequest` for that invitation you clone the `Inviter` and continue the protocol from there.

The way the previous state machine was designed was that an `Inviter` would respond to a `ConnectionRequest` in both the `Invited` and `Initial` states. So to kind-of make it consistent across the FFI and agent implementation, there's the `into_invited` method to transition the inviter into `Invited` from `Initial`, so it can then process the `ConnectionRequest`.

I'll also document that using `into_invited()` is discouraged and you should clone the `Connection<Inviter, Invited>` instead.

libvcx/src/api_vcx/api_handle/connection.rs
gmulhearn: kk. just to be clear, was referring to the `lock.get` and the mapping to the same error type, as it's duplicated 6 times;

```
fn get_borrowed_connection_from_lock<'a>(lock: &'a RwLockReadGuard<HashMap<u32, GenericConnection>>, handle: u32) -> LibvcxResult<&'a GenericConnection> {
    lock.get(&handle).ok_or_else(|| LibvcxError::from_msg(
        LibvcxErrorKind::ObjectAccessError,
        format!("No connection found for handle: {}", handle),
    ))
}
```

libvcx/src/api_vcx/api_handle/out_of_band.rs
bobozaur: I honestly have no idea. It was needed for backwards compatibility, but I don't really know why they would be removed in the future.

I would assume though that the rationale would be to allow the connection matching in the client rather than doing it in the library or even the FFI. The OOB invitation matching with existent connections was already too intertwined by requiring a `Vec<&GenericConnection>`, so switched to needing a generic iterator instead. 

But I still think they might be a bit too intertwined still.

gmulhearn: Oh sry, i see, it's just copied from the other mirrored API - `connection_exists`

aries_vcx/src/protocols/connection/inviter/mod.rs
bobozaur: Added more comments.

aries_vcx/src/protocols/connection/invitee/mod.rs
bobozaur: Fixed

aries_vcx/src/protocols/connection/inviter/mod.rs
bobozaur: Fixed

libvcx/src/api_vcx/api_handle/connection.rs
bobozaur: Ah, fair. I was speaking from memory and forgot that there's also the common error being returned if a connection is not found for a handle, apart from the attribute specific error returned if the connection is in an invalid state and does not contain the attribute.

Nevertheless, I added `get_con_attribute_with_closure` to factor out the duplicate code.

gmulhearn-anonyome:
I may have noticed a potential bug. After the invitee calls `send_request`, if the invitation was pairwise, the `thread_id` for the invitee `Requested` state is not set as the request message's ID. Shouldn't it be set as that, such that when the `verify_thread_id` is called in `handle_response` it can verify that the Response message thread_id matches the request ID that we sent? 

My reading of RFC 0160 is that the invitee's request ID should == the inviter's response thid.

I'm a little confused about how this works, could you guys shed some light on if I'm misunderstanding? a test could be nice. 

Also two other things;
 it would be nice if the Request message's ID is set as a UUID rather than using a default (`"testid"`) - like how the generate_invitation does now. 
is it necessary to set the `thread_id` of the request message? rfc 0160 does not seem to require or imply this.

bobozaur:
@gmulhearn I replicated the old behavior, so I'm not 100% certain of the implementation details, but from my limited understanding a pairwise invitation is for a pair of agents. You wouldn't expect a second invitee to accept it. So the thread ID can remain the invitation ID, as it leads to contiguous context tracking.

For the Request ID, it's out of scope for this PR. It's the old behavior and while I agree it should change, probably many things could/should 😅 .

Patrik-Stas (approved):

Patrik-Stas:
The current failure on nodejs serialization in CI is fixed in this PR

https://github.com/hyperledger/aries-vcx/pull/749

Patrik-Stas:
@gmulhearn is `testid ` actually ever used as value in non-test context?
@gmulhearn About the logic of setting `@id` as response to pairwise invitation, I agree with @bobozaur in scope of this PR I'd just stick to original behaviour, for starter.
We can discuss this separately, but I'll briefly share my opinion - I think this is actually right behaviour, as in case of connection protocol kicked-off by inviter sending pairwise invitation, that's esentially the first agent2agent (altough out of band) message kicking off the protocol with certain `@id`. Then the `connection-request` just continues in that thread (dictated by invite `@id`).

gmulhearn (approved):
Lgtm!

gmulhearn:
Kk no worries, I might open an issue where further discussion can happen. But ye agreed, not in scope here

mirgee (approved):

Issue #742, opened 2023-02-02
Handlers are missing a alternative forms of construction (`from_parts`)


gmulhearn-anonyome:
The `MediatedConnection` handler has a public `from_parts` API, which allows consumers to reconstruct their `MediatedConnection` objects with typed data. 

This is useful to consumers who have data representing a 'mediatedconnection' in another form (e.g. they are coming from another lib, or they are internally storing 'connections' in their own format).

`from_parts` allows consumers to not depend on the exact format/result of serializing/deserializing the handler (which may change over time, especially with the recently proposed typestate patterns).

However other handlers - Connection, Prover, Verifier, Holder, Issuer don't have such a function, and the only way to reconstruct them is by serializing and deserializing. So if you want to reconstruct them, and you don't have a serializing version of the handler, your only option is to hackily recreate the serializing format with your own data and try deserialising it... 

So i'd propose that, like `MediatedConnection`, handlers should have an alternative (re-)constructor method such as `from_parts`. In making this method, we'd need to ensure that consumers can publicly access ways to construct the 'parts' they need. 
E.g. for `Connection::from_parts(..., state: SmConnectionState)`, we'd need to ensure that consumers can construct the `SmConnectionState` object virtually from scratch - which I believe is already the case, but just pointing it out.

Alternatively, we could make the fields of handlers all public, but IMO `from_parts` constructors would be cleaner.

gmulhearn-anonyome:
Let me know if you agree or disagree with this proposal - was there a reason that all other handlers are missing the `from_parts` constructor? I will start working on an PR for `Connection` handler in the meantime

gmulhearn-anonyome:
Update; from the meeting last week, we agreed to include `from_parts` and `to_parts` operators in handlers. These were added to `Connection` with the typestate MR #740 . However other handlers are still missing them

PR #743, opened 2023-02-02
Connection Handler from_parts constructor

gmulhearn-anonyome:
Part of #742 - pending validation of issue from vcx team.

### DIFF
--- a/aries_vcx/src/handlers/connection/connection.rs
+++ b/aries_vcx/src/handlers/connection/connection.rs
@@ -61,6 +61,24 @@ impl Connection {
         })
     }
 
+    pub fn from_parts(
+        source_id: String,
+        thread_id: String,
+        pairwise_info: PairwiseInfo,
+        state: SmConnectionState,
+    ) -> Self {
+        let connection_sm = match state {
+            SmConnectionState::Inviter(state) => {
+                SmConnection::Inviter(SmConnectionInviter::from(source_id, thread_id, pairwise_info, state))
+            }
+            SmConnectionState::Invitee(state) => {
+                SmConnection::Invitee(SmConnectionInvitee::from(source_id, thread_id, pairwise_info, state))
+            }
+        };
+
+        Self { connection_sm }
+    }
+
     // ----------------------------- GETTERS ------------------------------------
     pub fn get_thread_id(&self) -> String {
         match &self.connection_sm {


gmulhearn-anonyome:
Hmm. It may not actually be this simple... As a consumer cannot fully construct a `SmConnectionState` themselves without having a serialized copy of it...

This is because the `states` submodule of `protocols::connection::[invitee | inviter]` is private. Meaning that consumers cannot construct the states within `InviteeFullState` and `InviterFullState`.

e.g. `SmConnectionState::Invitee(InviteeFullState::Invited( InvitedState { ... } ))` cannot be constructed by consumers, since `InvitedState` is a private struct...

@Patrik-Stas I'm unsure of the best solution to this? ^, we could maybe make that `states` submodule public, are there any consequences to this? - this is all assuming that you agree with me in believing that handlers should have an alternative mechanism for construction other than serialization, which you may not agree with. Keen to hear your opinions 👌

Patrik-Stas:
Hi @gmulhearn-anonyome I am afraid I didn't properly communicate the Connection refactor PR https://github.com/hyperledger/aries-vcx/pull/739 as under the current plan, this PR #743 wouldn't apply anymore, as new `Connection` implementation is quite different. 
Although https://github.com/hyperledger/aries-vcx/pull/739 was originally set-up as addition only, given the short history of non-mediated `Connection`, we've decided to apply the refactor directly on it (instead of maintaining 2 implementations of `Connection`)

If this comes as too sudden breaking change, we can keep `Connection` (just rename it perhaps), but long term #739 seems superior to the current approach for state machines.

Let's first clarify this and then we can address the actual proposal / problem you are trying to solve.

gmulhearn:
Oh nice. Thought that might be the case! Ok no worries, I'll need to check in that MR if my issue in connections has been fixed. It ultimately comes down to whether or not the states can be manually reconstructed.

Regardless the issue I created still applies for other handlers in vcx. (I.e. they're missing constructors). So let me know if you agree that's an issue

Patrik-Stas:
@gmulhearn 
Just to make sure to get it 100% straight, let's look at aries-vcx serialized Verifier 
```
{
    "verifier_sm": {
      "source_id": "1",
      "thread_id": "0a794e24-c6b4-46bc-93b6-642b6dc98c98",
      "state": {
        "PresentationRequestSent": {
          "connection_handle": 3330318696,
          "presentation_request": {
            "@id": "testid",
            "comment": "default wants you to share Optional",
            "request_presentations~attach": [
              {
                "mime-type": "application/json",
                "@id": "libindy-request-presentation-0",
                "data": {
                  "base64": "eyJuYW1l....IjEuMCJ9"
                }
              }
            ]
          }
        }
      }
    }
```
I suspect in this case, the 3 parts you would like to get is `(source_id, thread_id, state)`,  and serialize it on your own, presumably in simpler format skipping the redundant top level "verifier_sm" node, and have it serialize into something like:
```
{
      "source_id": "1",
      "thread_id": "0a794e24-c6b4-46bc-93b6-642b6dc98c98",
      "state": {
        "PresentationRequestSent": {
          "connection_handle": 3330318696,
          "presentation_request": {
            "@id": "testid",
            "comment": "default wants you to share Optional",
            "request_presentations~attach": [
              {
                "mime-type": "application/json",
                "@id": "libindy-request-presentation-0",
                "data": {
                  "base64": "eyJuYW1l....IjEuMCJ9"
                }
              }
            ]
          }
        }
      }
 ```
Is that right? 

I think your proposal makes totally sense given the current state of art. There's few issues few current default serialization formats so I can see why you would want to serialize on your own into a different one. 
I think we can go forward adding the `from_parts`, `into_parts` kind of API to other protocol handlers in their current form.

Keep in mind that in relatively soon future (I think in terms of weeks) we would like to deprecate current state machines in favor of their new reimplementation using #739 approach - but this can be separate discussion as how this can be done in a sensible manner to prevent causing too much disruption. 

It's actually great you bring this up, cause I find de/serialization as one of the top weaknesses of `aries-vcx` and along with the #739 topic of de/serialization has been brought up quite a bit and your proposal this PR/linked issue confirms that further. We have a great opportunity to get that fixed while building new #739 state machines. I'll create issue to address current issues with serialization. I think if we make the serialization format better, you perhaps wouldn't need to come up with custom de/serialization in a first place.

**Comment by Patrik-Stas** [comment]
I've created https://github.com/hyperledger/aries-vcx/issues/745 as proposal for new version of serialization formats. Do you think the proposed format would remove the need to create your own?

Also, I'd like to summon @mirgee @bobozaur what do you guys think?

**Comment by gmulhearn** [comment]
@bobozaur changes somewhat solve this problem in connection as well, as all the state objects now have a publicly accessible `::new(..)` methods so that consumers can manually construct them.

here's two examples on bobozaur's branch of how a consumer can construct handler's using primitive data which they should be storing anyway:
(please note that i did have to change bobozaur's `from_parts` construct from `pub(crate)` to `pub` for this example.
```
    let conn = Connection::<Invitee, Invited>::from_parts(
        String::new(),
        PairwiseInfo {
            pw_did: my_did,
            pw_vk: my_verkey,
        },
        Invitee,
        Invited::new(did_doc, invitation),
    );

    let conn = Connection::<Inviter, inviter::states::requested::Requested>::from_parts(
        String::new(),
        PairwiseInfo {
            pw_did: my_did,
            pw_vk: my_verkey,
        },
        Inviter,
        inviter::states::requested::Requested::new(signed_response, did_doc),
    );
```

So in this example a consumer is able to re-construct their Connection handler using the base SSI primitive data which they would be likely to storing and managing themselves in their own agent.

gmulhearn:
If this thread is getting too much, would be good to continue the discussion in this week's call if possible 👌

bobozaur:
I think a discussion in the weekly call is the best way to go about it, but I'll also provide my two cents here:

I believe we should not guarantee a serialization format of the connection and even discourage people from directly accessing the serialized format. Things can still change in terms of the format and it would be best not to make a promise like this unless we are completely sure we won't ever break it. But even then, direct access should not be encouraged and rather working with `getters`.

I'm quite torn between whether we should provide the additional constructor or not. @gmulhearn makes a good point that people can modify the serialized connection and deserialize to get a different state if they really want to and maybe migrating a connection from a different agent does make sense. However, the ability of constructing an arbitrary state sounds like looking for trouble.

But it might be just an irrational thought. After all, even if a connection is constructed improperly, it doesn't really do much without the counterpart connection of the other agent being in an improper state. So we might as well make it easy and less-error prone by adding `from_parts` and `into_parts` methods instead of having people implement some sort of connection replicas just for serialization/deserialization into different states.

gmulhearn:
Yeah exactly ^. Without going into the nitty gritty right now, we have a particular use case where a handler needs to be restored via protocol data being stored in our own format, and the only way to do that currently is via the connection replica + serialization/deserialization hacky strategy mentioned by Bogdan ^. 

Of course my particular use case should not inform aries-vcx design, however I imagine this use case will not be that uncommon to anyone building an agent with persistent storage onto of aries-vcx. And IMO since the goal can be accomplished in a hacky/dangerous way currently, we may as well just make it possible in a less hacky and type-safe way via providing constructors. 

anyway, keen to discuss this further in the call

gmulhearn:
handled in typestate PR

PR #746, opened 2023-02-08
Fix the signature VcxAPI.credentialGetOffers

raphaelguye:
Fixing the signature of the implementation to match the one from the header to avoid a crash when calling the function.

[VcxAPI.h](https://github.com/hyperledger/aries-vcx/blob/66ee076fb4abd479dd6bbd05eb42270bdca1643b/wrappers/ios/vcx/VcxAPI.h#L444):

[VcxAPI.m](https://github.com/hyperledger/aries-vcx/blob/66ee076fb4abd479dd6bbd05eb42270bdca1643b/wrappers/ios/vcx/VcxAPI.m#L1769):


raphaelguye:
Close because of no "Signed-off-by"
Use #748 instead.

PR #749, opened 2023-02-10
Simplify connection serialization from nodejs

Patrik-Stas:
- NodeJS wrapper was previously trying to keep value of `source_id` as part of constructed JS objects / TS classes, see changes in class `VcxBase`
```
protected _sourceId: string;
```
- Consequently deserialization constructors requiring `source_id`. That led me to generalizing signature for its static method ` _deserialize` such the serialized data to be deserialized are simply a JSON object (expressed as `Record<string, unknown>` in TS) instead of previous more strict definition `objData: ISerializedData<{ source_id: string }>`  which couldn't acommodate different serialization format of `Connection` state machine.
- Additionally removed some dead code
- Note: More code can be purged and cleaned up, for example given the decision to treat serialized data as opaque, it makes no sense to define `ISerializedData`, as this is expression of certain assumption as to how that serialized data looks like. The only thing which should be held by TS classes is the numeric rust handle
- Simplifies `Connection` serialization such that in serialized to (removing extra `data` level in hierarchy, removes duplicated field `source_id`, removes `version`)
```
{
    "source_id": "",
    "pairwise_info": {
        "pw_did": "FhrSrYtQcw3p9xwf7NYemf",
        "pw_vk": "91qMFrZjXDoi2Vc8Mm14Ys112tEZdDegBZZoembFEATE"
    },
    "state": {
        "Inviter": {
            "Initial": null
        }
    }
}
```

### DIFF
--- a/wrappers/node/src/api/common.ts
+++ b/wrappers/node/src/api/common.ts
@@ -165,10 +165,6 @@ export enum VerifierStateType {
   Failed = 5
 }
 
-export interface IInitVCXOptions {
-  libVCXPath?: string;
-}
-
 export interface ISerializedData<T> {
   version: string;
   data: T;
--- a/.github/actions/setup-testing-nodejs/action.yml
+++ b/.github/actions/setup-testing-nodejs/action.yml
@@ -13,10 +13,14 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Use Node.js ${{ inputs.node-version }}
+    - name: "Use Node.js ${{ inputs.node-version }}"
       uses: actions/setup-node@v3
       with:
         node-version: ${{ inputs.node-version }}
+    - name: "Install npm 8.19.3"
+      shell: bash
+      run: |
+        npm install -g npm@8.19.3
     - uses: actions-rs/toolchain@v1
       with:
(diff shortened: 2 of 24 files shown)


bobozaur (changes requested):
Not related to the code itself, but rather to the base branch.
I think it would be best to merge #739 and merge this into main afterwards.

The current state of #739 behaves exactly the way main currently does. This PR changes that, which is why I think it deserves it's separate spot for code history's reasons.

bobozaur (approved):
Looks good!

wrappers/node/src/api/common.ts
```diff
@@ -165,10 +165,6 @@ export enum VerifierStateType {
   Failed = 5
 }
 
-export interface IInitVCXOptions {
-  libVCXPath?: string;
-}
-
 export interface ISerializedData<T> {
```
mirgee, line 168: Note: we agreed to remove the assumptions this interface makes about the serialization format (i.e. containing `source_id`, `data`, `version` or any other fields, but the assumption of it being a JSON will stay for now).

mirgee (approved):

PR #755, opened 2023-02-15
Add tests on top of messages_crate PR

Patrik-Stas:


messages2/src/message_type/message_family/connection.rs
```diff
@@ -16,13 +16,11 @@ pub enum ConnectionV1 {
...
 
 #[derive(Copy, Clone, Debug, AsRefStr, EnumString, PartialEq)]
-#[strum(serialize_all = "kebab-case")]
+#[strum(serialize_all = "snake_case")]
 pub enum ConnectionV1_0 {
     Invitation,
     Request,
     Response,
-    // Note that the underscore is on purpose!
-    #[strum(serialize = "problem_report")]
```
bobozaur, line 25: Just an observation: When I see it, I guess it makes more sense to use `snake_case` formatting here, as these messages won't change. It's just odd how across the ecosystem both `kebab-case` and `snake_case` seem to be used, with `kebab-case` being predominant, so I resorted to that.

bobozaur (approved):
I agree with merging this.

PR #759, opened 2023-02-21
Split libvcx in 2 crates, update diagrams

Patrik-Stas:
Split `libvcx` in 2 crates:
- `libvcx_core` - the handle layer of `libvcx` extracted as separate crate
- `libvcx` - this represents the same thing as `libvcx` used to be. Built on top of `libvcx_core`. This crate is now deprecated.


libvcx_core/src/api_vcx/api_handle/credential.rs
```diff
@@ -338,6 +338,11 @@ pub async fn decline_offer(handle: u32, connection_handle: u32, comment: Option<
     HANDLE_MAP.insert(handle, credential)
 }
 
+#[cfg(feature = "test_utils")]
+pub mod tests_utils {
+    pub const BAD_CREDENTIAL_OFFER: &str = r#"{"version": "0.1","to_did": "LtMgSjtFcyPwenK9SHCyb8","from_did": "LtMgSjtFcyPwenK9SHCyb8","claim": {"account_num": ["8BEaoLf8TBmK4BUyX8WWnA"],"name_on_account": ["Alice"]},"schema_seq_no": 48,"issuer_did": "Pd4fnFtRBcMKRVC2go5w3j","claim_name": "Account Certificate","claim_id": "3675417066","msg_ref_id": "ymy5nth"}"#;
+}
+
```
bobozaur, line 345: Probably not in the scope of this PR, as this spans across most the crates in the repository, but we should clear out these test-only stuff from the actual code.

Instead, we could provide something like `common.rs` (or `common` as a dir) in the integration `tests` folder where we can declare all the tests related stuff. Then `mod common` and `use common::*` would allow us to use all of these in integration tests without having to intertwine these things with the code.

Alternatively, if these need to be available for unit/doc tests, we could have separate crates for them directly. It would make the implementation crates cleaner and smaller.

bobozaur (dismissed):
Apart from the tests thing (which would probably be handled separately), LGTM.

libvcx/src/api_c/ledger/schema.rs
```diff
@@ -608,9 +608,9 @@ mod tests {
     fn test_vcx_create_schema_success() {
         let _setup = SetupMocks::init();
 
-        let (_, schema_name, schema_version, data) = prepare_schema_data();
-        let handle = vcx_schema_create_c_closure(&schema_name, &schema_version, &data).unwrap();
-        assert!(handle > 0)
+        let foo = prepare_schema_data();
+        // let handle = vcx_schema_create_c_closure(&schema_name, &schema_version, &data).unwrap();
+        // assert!(handle > 0)
```
mirgee, line 613: Can this assertion be re-enabled?

Patrik-Stas: good catch, reverted changes

libvcx_core/src/api_vcx/api_handle/credential.rs
Patrik-Stas: Yes, the `test_utils` under the `/src` are used for unit tests - though I am actually leaning to keep this as is, I am not strongly opinionated. I find it nice that the unit test utils for some domain are in respective files and right next to the tests which are using them the most.

We can discuss on other channels further

mirgee (approved):

PR #762, opened 2023-02-28
OOB invite attachment field conditional serialization, default to empty if not present

mirgee:
Currently:

* deserialization of OOB invitation fails if `requests~attach` field is not present, and
* when no messages were attached to OOB invitation, the attachments field serializes to an empty vec unnecessarily.

This change:

* sets default value for the field to fix the first issue, and
* skips serialization if the attachments are empty to fix the second issue.

This is a breaking change: prior versions of aries-vcx are not forward compatible with this change and fail to deserialize OOB invitations without `requests~attach` field. However, after this change, aries-vcx is still backwards compatible with OOB invitations produced by earlier versions of aries-vcx.

### DIFF
--- a/messages/src/protocols/out_of_band/invitation.rs
+++ b/messages/src/protocols/out_of_band/invitation.rs
@@ -23,7 +23,11 @@ pub struct OutOfBandInvitation {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub handshake_protocols: Option<Vec<MessageType>>, // TODO: Make a separate type
     pub services: Vec<ServiceOob>,
-    #[serde(rename = "requests~attach")]
+    #[serde(
+        default = "Attachments::new",
+        skip_serializing_if = "Attachments::is_empty",
+        rename = "requests~attach"
+    )]
     pub requests_attach: Attachments,
     #[serde(rename = "~timing")]
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


messages/src/protocols/out_of_band/invitation.rs
```diff
@@ -23,7 +23,11 @@ pub struct OutOfBandInvitation {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub handshake_protocols: Option<Vec<MessageType>>, // TODO: Make a separate type
     pub services: Vec<ServiceOob>,
-    #[serde(rename = "requests~attach")]
+    #[serde(
+        default = "Attachments::new",
+        skip_serializing_if = "Attachments::is_empty",
```
Patrik-Stas, line 28: Should it be solved by rather by wrapping `Option` around `requests_attach`? Cause currently OOB sender will produce correct OOB if no attachments are present, but OOB receiver won't be able to parse it, since `requests_attach` is required.
And I suppose we are missing test-case for that

Patrik-Stas (dismissed):
lgtm

We should also put notice into pr description / with tags, that this is breaking change in setup where 
- revision based on this PR produces OOB (with no attachemnts)
- an older revision of `aries-vcx` receives the OOB.
As the older revision on receiver side would break due missing required `requests~attach` field.

Not aware of anyone using `aries-vcx` on server side atm, so shouldn't impact anyone in practice. 
And obviously no breaking change against acapy, rather a fix.

messages/src/protocols/out_of_band/invitation.rs
```diff
@@ -23,7 +23,11 @@ pub struct OutOfBandInvitation {
...
+        rename = "requests~attach"
+    )]
```
bobozaur, line 30: I think you can say just `#[serde(defaul)]`, as that calls the `Default::default()` of the type, so `Attachments::default()`.

bobozaur (changes requested):

bobozaur (approved):
LGTM

Patrik-Stas (approved):

PR #763, opened 2023-03-01
Experiment - Vdrtools as a feature flag

gmulhearn:
This is mostly just an experimental proof of concept to put all aries-vcx usages of libvdrtools under a feature flag (which aries-vcx has as a default flag).

The idea is that consumers who do not want to use vdrtools dependencies (e.g. using modular deps or their even own deps instead), can override the default flag and use their own `Profile`. This way the consumer does not have to pull in the heavy-weight deps of libvdrtools

I imagine once the modular deps (i.e. issuer&verifer & askar-wallet) implementations are fleshed out, we would have a feature flag for those implementations as well, and then importing consumers can toggle between what they wish to use.

### DIFF
--- a/aries_vcx/Cargo.toml
+++ b/aries_vcx/Cargo.toml
@@ -11,6 +11,8 @@ path = "src/lib.rs"
 doctest = false
 
 [features]
+default = ["vdrtools"]
+vdrtools = ["dep:libvdrtools"]
 test_utils = [ "messages/test_utils" ]
 pool_tests = [ "test_utils" ]
 agency_pool_tests = [ "test_utils" ]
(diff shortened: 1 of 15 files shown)


gmulhearn:
@Patrik-Stas requested for review a bit early bcus i'm too impatient for wait for the pipeline! I will double check that it has passed later... but it should be good

aries_vcx/Cargo.toml
Patrik-Stas, line 1: nice job with feature documentation

Patrik-Stas (approved):

Patrik-Stas:
@mirgee @bobozaur one of you please review and merge if you good with this

Patrik-Stas:
@gmulhearn another great incremental step forward!

bobozaur (approved):
I'm okay with the changes considering the goal of the PR.

I think that in the future, though, we should strive towards refactoring all this test feature flag and mocks mess outside of aries-vcx (at least it's `src`).

Since aries-vcx does not depend on `libvdrtoola`, then it shouldn't even be used. We could maybe provide a vdrtools-based wallet et al in a separate crate that can be used by consumers.

Patrik-Stas:
@bobozaur just some clarification
> Since aries-vcx does not depend on libvdrtoola, then it shouldn't even be used. 
aries-vcx does depend on libvdrtools, but the goal is to get rid of it - the feature flags are way to opt-in either `vdrtools` or `indy-vdr+credex` as implementation under the hood.

The testing feature flags are orthogonal to the changes in this pr, though I don't disagree there can be better way to test

PR #768, opened 2023-03-08
Draft: Apply Typestate pattern to Holder

gmulhearn:
TODO...

To determine:
* Approach for legacy `Holder` support? - should this replace the existing or be a new construct?
* How to handle state transitions which previous result in two different state?.. e.g. `Holder<OfferReceived>::send_request` could progress into `Failed` or `RequestSent` according to the old API


Patrik-Stas:
Thanks for this work! Amazing to see the upgrade on other protocols.

1. I think it depends, we can keep your current approach having this as completely new code. If we by the end of this work find out there's not breaking changes in data structures for important states (`Finished` for holder, `CredentialSent` and `Finished` for issuer), it could be fairly smooth to replace old code by new one. 
But if we find out we want to actually take the new implementation to make bunch of changes as well, then we can just deprecate old code and endorse the new state type pattern implementation). Personally I encourage you to fix any weirdness you run into and not hold yourself back too much for sake of compatibility. 

2. Not sure about this. I guess that's a case we didn't run into with connection protocol. First thought which comes to my mind is that we'll just have to sometimes return some enum of 2 state and then the user would have to match the variant.
Perhaps smthing like:
```
enum SendRequestResult { 
   Failed(FailedState),
   RequestSent(RequestSentState)
}
```
and then have `Holder<SendRequestResult>::send_request`?

Alternatively it can be worth looking from completely different angle, could we somehow linearize the state machine so that every method would call would either fail or progress (and return) new state? 
I feel like this is actually the right way. Either pass to next state, or return error, and let the wrapping code - either retry or manually move the state machine to failed state. That would mean that transitions like `send_request` would no longer try to send problem report automatically.

Patrik-Stas:
Just to make sure i am coming clear with point `2b`, simple picture to further elaborate on the suggestion

image

image

bobozaur:
I second the idea of striving to deprecate the old implementation in favor of the typestate pattern. Ideally, the new impl would replace the old one wherever it was used.

Regarding error handling, I'll mention the ideas proposed in the community call:

1) Error states such as `Holder<Failed>` could implement `Error` and the result from an operation would be `Result<Holder<NextState>, Holder<Failed>>`. This would be ideal because a state machine ending in that state is practically unusable and should only be used for logging/error handling purposes, not reused in some way.
2) I haven't done this in the `Connection` state machine (**yet**), but the way I've seen it being used makes me believe we could return things like `Result<Connection<NextState>,(Connection<OldState>, Error>)`. Essentially, this would allow us to re-use the state machine in case of an error unrelated to the protocol progress (network issue, wallet issue, etc.). This would allow us to stop cloning a state machine whenever we wanna use it and instead be able to re-use it in case of a non-protocol failure.

gmulhearn:
Thanks for the feedback @Patrik-Stas &  @bobozaur.

Something I've been struggling with finding a solution for, is the question around; when is it aries-vcx's job to send a problem report and consider a protocol as 'failed'? Should it ever be aries-vcx's job to do that automatically, or should we just offer the consumer an easy way for them to enter this state themselves?

Consider `send_request`; in the original handler, if it fails to `_make_credential_request`, it automatically sends a problem report and transitions into the terminal `failed` state.

But this could fail for numerous reasons;
* a network anomaly occurs when communicating with the ledger
* the wrong `profile` is used by the consumer - e.g. wrong wallet, or wrong genesis file being used for ledger
* the wallet isn't opened
* etc

A lot of these cases are not "fatal" failures, and they could potentially be recovered by the consumer in a way that lets them continue with the protocol.

So I wonder if instead, all aries-vcx should do, is attempt to progress in the protocol, and if it fails, we return that error back to the consumer without transitioning to 'fatal' or sending a problem report on their behalf. And then, if the consumer can process the error (with whatever application logic they have), and if they decide it is recoverable, then they can try again. And if they decide it is a fatal error, then they can call on some `send_problem_report` API which sends the report and transitions them into the `Holder<Failed>` state.

gmulhearn:
Another benefit of this approach, is that it allows consumers to control what the problem_report message payload is. Currently, the handlers just convert the VcxError to a string and send that as the problem report payload.

The consumer may not desire that behavior, and it may even potentially leak details they did not wish to share with the peer they are connected with.

gmulhearn:
I do agree with @bobozaur that since these type state APIs consume the current object, it would be nice if (in the case of an error result) we return back the original object in some way. As this could be more efficient than requiring them to clone the entire `Holder` before they partake in any consuming APIs.

bobozaur:
@gmulhearn I agree with you that we should be as minimal as possible in terms of what we do. Things like sending a `ProblemReport` should be the user's decision. We perhaps only have the responsibility of generating it (when appropriate).

In fact, I do want to explore a *sans IO* approach to the protocols, but it would be quite the research. The idea would essentially be to take all IO out of our protocol implementations and leave it to the user. 

A lot of errors that would occur because of this would simply not be our responsibility, and while that sounds like we're avoiding dealing with them, we in fact let users deal with them however they want, without sacrificing the state machines and stuff like that.

For instance, I want to explore the idea of having something like a `Guard` ZST bound to the state machine in such a way that if you don't take the appropriate (possibly external) action, you don't get an instance of the `Guard`  back that you would use to progress the state machine.

A very raw draft would be something like this:
```rust
pub struct Guard {
    users_cant_build_me: () // or PhantomData, or whatever
}

fn main() {
    let holder: Holder<OfferReceived> = gimme_holder();
    let guard: Guard = holder.send_request().unwrap() // non-consuming
    let holder: Holder<RequestSent> = holder.progress(guard);
}
```

This might require additional states in the state machine, beside the ones provided in the protocol (e.g: OfferReceived -> RequestCreated -> RequestSent). I personally don't see this as a problem, as I think the ones specified in the protocols are basically *flow-related* and not absolute. Also, the benefits of something like this are, in my opinion, huge.

The end goal of an approach like this would be to make all state machines operations either *non-consuming* or *infallible*.

Patrik-Stas:
hi @gmulhearn how do you see this coming along? I understand that you've mostly been waiting for moving parts around the `messages` to settle. I find it quite fortunate as in the meantime, we had extra time to come up with the implementation guidelines https://github.com/hyperledger/aries-vcx/blob/main/aries_vcx/docs/guidelines.md

Perhaps before you restore the work on this, how about you sketch out design for the states and transitions - we can get that approved here so there won't be any surprises upon the review

gmulhearn:
hey @Patrik-Stas , we can discuss this further tomorrow, but here's what I'm thinking so far:

# Draft of new APIs:

```
impl Holder<ProposalPrepared> {
    fn create_from_proposal(proposal_data) -> Self
    fn get_proposal_message(&self) -> CredentialProposalMessage

    // can only fail if invalid thread_id
    fn receive_offer(&self, offer: CredOfferMsg) -> Result<Holder<OfferReceived>, AriesVcxError>
}

impl Holder<OfferReceived> {
    fn create_from_offer(offer) -> Self
    
    fn prepare_proposal(self, proposal_data) -> Holder<ProposalPrepared>
    
    // can fail if: offer attachment could not be parsed, create-cred-request fails (ledger or anoncreds failure)
    fn prepare_request(&self, ...) -> Result<Holder<RequestPrepared>, AriesVcxError>

    fn decline_offer(self) -> Failed
}

impl Holder<RequestPrepared> {
    fn get_request_message(&self) -> CredReqMessage
    
    // can fail if: cred attachment could not be parsed, processing/storing cred fails (ledger or anoncreds failure)
    fn receive_credential(&self, cred: CredMsg) -> Result<Holder<Finished>, AriesVcxError>
}

impl Holder<Finished> {
    fn get_ack_message(&self) -> Option<AckMessage>
}

impl Holder<Failed> {
    fn get_problem_report(&self) -> ProblemReport
}

// impl for all states
impl Holder {
    // could rename this to just `fail`?
    fn transition_to_failed(self) -> Holder<Failed>

    // ....
}
```

_Note that I'm just focusing on the new transition based APIs, there will also be some utility APIs around getting data at different states, but that should be fairly straight forward_

# Things I'm unsure of:
## Error Transitions
When transitions have the possibility of throwing an Error (e.g. `prepare_request`) I'm having the APIs take in `&self` as the param rather than `self`. This is one approach to satisfying this statement in the guidelines:
> Transition should either fail and leave the current state unmodified, or transition to predetermined new state.

If the transition results in an error, the consumer will still have access to `self`. I know we discussed fancier approaches, (such as `(self,..) -> Result<Holder<NewState>, (Holder<CurrentState>,AriesVcxError)>`), however I don't see the benefit over just taking `&self` - which is a simpler and easier approach. Maybe someone could shed some light of why the fancier approach is better?

## Transition in Failure
I have it such that transition into `Failed` are manually chosen by the consumer. This is done by calling the `transition_to_failed`. Is this what we intended.

## ACK handling in the Finished state...
`Holder<Finished>::get_ack_message` returns optional `AckMessage` bcus the current implementation of the Holder handler ONLY sends the ack message if the credential message has the `please_ack` flag enabled. So the idea is that if the `please_ack` flag was true, this method would return Some, else None. But I'm not 100% sure. IMO it's always good to ACK.

And further, should the ACK maybe be an intermediate step before `Finished`??

gmulhearn:
@bobozaur @Patrik-Stas 
keeping everyone in the loop; A few questions that were raised in the meeting today;

# Result types:
should our result types be;
* `Result<Holder<NewState>, AriesVcxError>`, or
* `Result<Holder<NewState>, Holder<Failed>>`

# Consumption of self;
for APIs which transition states AND have the chance of failing (error result), we want the consumer to not lose the original Holder state in the event that an error result occurs. So should the signature be;
* take self as a reference: `fn xyz(&self,..) -> Result<Holder<NewState>, Error>`
* consume self, but error type contains a tuple of the OG state: `fn xyz(self,..) -> Result<Holder<NewState>, (Holder<CurrentState>, Error)>`

IMO option 1 is much simpler for consumers BUT allows the consumers to _easily_ have the old and new state simultaneously.

Option 2 is more complicated for consumers,  BUT makes the APIs consistent and makes it harder for consumers to have the old and new state simultaneously.

# How to handle Ack...
The 0036 RFC has me believe that sending an ACK is an optional step for Holders:
> If the issuer wants an acknowledgement that the issued credential was accepted, this message must be decorated with ~please-ack, and it is then best practice for the new Holder to respond with an explicit ack message as described in [0317: Please ACK Decorator](https://github.com/hyperledger/aries-rfcs/blob/main/features/0317-please-ack/README.md).

Given this, do you think it is ok to have the ACK message be apart of the `Finished` state? i.e. 
```
impl Holder<Finished> {
    fn get_ack_message(&self) -> Option<AckMessage>
}
```
where it is `None` if please_ack was not decorated.

bobozaur:
> # Result types:
> 
> should our result types be;
> 
>     * `Result<Holder<NewState>, AriesVcxError>`, or
> 
>     * `Result<Holder<NewState>, Holder<Failed>>`

2nd option.

> # Consumption of self;
> 
> for APIs which transition states AND have the chance of failing (error result), we want the consumer to not lose the original Holder state in the event that an error result occurs. So should the signature be;
> 
>     * take self as a reference: `fn xyz(&self,..) -> Result<Holder<NewState>, Error>`
> 
>     * consume self, but error type contains a tuple of the OG state: `fn xyz(self,..) -> Result<Holder<NewState>, (Holder<CurrentState>, Error)>`
> 
> 
> IMO option 1 is much simpler for consumers BUT allows the consumers to _easily_ have the old and new state simultaneously.
> 
> Option 2 is more complicated for consumers, BUT makes the APIs consistent and makes it harder for consumers to have the old and new state simultaneously.

This is also a matter of performance. Taking `self`by reference means you have to clone it to transition, since you need to create a new type when you change the state.

Option 2 is the way to go, in case the transition can indeed fail.

> # How to handle Ack...
> 
> The 0036 RFC has me believe that sending an ACK is an optional step for Holders:
> 
> > If the issuer wants an acknowledgement that the issued credential was accepted, this message must be decorated with ~please-ack, and it is then best practice for the new Holder to respond with an explicit ack message as described in [0317: Please ACK Decorator](https://github.com/hyperledger/aries-rfcs/blob/main/features/0317-please-ack/README.md).
> 
> Given this, do you think it is ok to have the ACK message be apart of the `Finished` state? i.e.
> 
> ```
> impl Holder<Finished> {
>     fn get_ack_message(&self) -> Option<AckMessage>
> }
> ```
> 
> where it is `None` if please_ack was not decorated.

I think that the transition to `Finished` should give the consumer all they need to proceed, which meams either providing the message directly *or* embedding it into the state. 

Embedding it follows the "message generated state" philosophy we want to adopt, so I'd go with that.

Patrik-Stas:
> > George: 
> > # Result types:
> > should our result types be;
> > ```
> > * `Result<Holder<NewState>, AriesVcxError>`, or
> > 
> > * `Result<Holder<NewState>, Holder<Failed>>`
> > ```
> Bogdan: 
> 2nd option.

I was considering 1st option, as that would let the consumer "retry", but assuming we are minimizing IO (at least not sending external messages, although there might be IO with wallet), then consumers shouldn't retry to yield different results and there would be little value doing so.
So I agree with 2nd option as well. That means we'll need to somewhat update this paragraph
https://github.com/hyperledger/aries-vcx/blob/main/aries_vcx/docs/guidelines.md#deterministic-transitions
While still deterministic, the paragraph stated that we either move to NEXT state or keep the old state - eg. assumed the option 1 `Result<Holder<NewState>, AriesVcxError>`

> 
> > George:
> > # Consumption of self;
> > for APIs which transition states AND have the chance of failing (error result), we want the consumer to not lose the original Holder state in the event that an error result occurs. So should the signature be;
> > ```
> > * take self as a reference: `fn xyz(&self,..) -> Result<Holder<NewState>, Error>`
> > 
> > * consume self, but error type contains a tuple of the OG state: `fn xyz(self,..) -> Result<Holder<NewState>, (Holder<CurrentState>, Error)>`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > IMO option 1 is much simpler for consumers BUT allows the consumers to _easily_ have the old and new state simultaneously.
> > Option 2 is more complicated for consumers, BUT makes the APIs consistent and makes it harder for consumers to have the old and new state simultaneously.
> Bogdan:
> This is also a matter of performance. Taking `self`by reference means you have to clone it to transition, since you need to create a new type when you change the state.
> 
> Option 2 is the way to go, in case the transition can indeed fail.
I suppose you could potentially be consistent with both approaches, but with approach 1 you always end up with 2 state machines. With approach 2, you by default only have 1 - making maintaining pre-transition version opt-in by coping it prior to executing the transaction. So I'd go with 
```
fn xyz(self,..) -> Result<Holder<NewState>, (Holder<CurrentState>, Error)>
```
or even just
```
fn xyz(self,..) -> Result<Holder<NewState>, Holder<Failed>>
```

> > George:
> > # How to handle Ack...
> > The 0036 RFC has me believe that sending an ACK is an optional step for Holders:
> > > If the issuer wants an acknowledgement that the issued credential was accepted, this message must be decorated with ~please-ack, and it is then best practice for the new Holder to respond with an explicit ack message as described in [0317: Please ACK Decorator](https://github.com/hyperledger/aries-rfcs/blob/main/features/0317-please-ack/README.md).
> > 
> > 
> > Given this, do you think it is ok to have the ACK message be apart of the `Finished` state? i.e.
> > ```
> > impl Holder<Finished> {
> >     fn get_ack_message(&self) -> Option<AckMessage>
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where it is `None` if please_ack was not decorated.
> Bogdan: 
>
> I think that the transition to `Finished` should give the consumer all they need to proceed, which meams either providing the message directly _or_ embedding it into the state.
> 
> Embedding it follows the "message generated state" philosophy we want to adopt, so I'd go with that.

Makes me think that there can be more state outcomes than just "SuccessState" and "FailState" for some transition. 

As I was thinking here, at first I drafted this:
```
enum HolderFinishedState { 
   Finished(Finished)
   FinishedNotAcked(FinishedNotAcked)
}

impl Holder<RequestPrepared> {
    fn receive_credential(&self, cred: CredMsg) -> Result<Holder<HolderFinishedState>, Holder<Failed>>
}

impl Holder<FinishedNotAcked> {
    fn mark_acked(&self, cred: CredMsg) -> Holder<Finished>
}
```
So `receive_credential` would returns new state conditionally based on content of the received message. If the credential message contains `please-ack` request, then we would return `Holder<FinishedNotAcked>`. If not, we would return `Holder<Finished>`.

But this is kind of complicating things as now  the `T` in `Holder<T>` is some enum, which wraps 2 states - so why not just flatten in out to something like "HolderResultReceiveCredential" aka our "context specific" custom Result type.
```
impl Holder<RequestPrepared> {
    fn receive_credential(self, cred: CredMsg) -> HolderReceiveCredentialResult<Holder<Finished>, Holder<FinishedNotAcked>, Holder<Failed>>
}
```
So then the consuming code would have to much all of the N variants
```
let new_sm: Holder<Finished> = match sm.receive_credential(cred_msg) {
  HolderReceiveCredentialResult::Finished(sm_finished) -> { 
     info("nice, we didn't even need to try to ack!");
     sm_finished
   },
  HolderReceiveCredentialResult::FinishedNotAcked(sm_not_acked) -> {  
    let ack_msg = sm_not_acked.get_ack_message()
    if let Err(error) = send_msg(ack_msg) {
       error!("Failed to ack and we are giving up ¯\_(ツ)_/¯")
    } 
   let sm_finished: Holder<Finished> = sm_not_acked.mark_acked(&self);
   sm_finished
  },
  HolderReceiveCredentialResult::Failed(sm) -> {  
    unwrap!("omg!")
  } 
}
```
Obviously the disadvantage is that we lose ability to use `?` to propagate up the error result variant wrapping `Holder<Failed>`
So this is actually broader suggestion about how we could express varios states a transition can lead to. Not sure if this is good idea, just brainstorming :-) 
Also this is even more in violation of forementioned guideline we agreed on https://github.com/hyperledger/aries-vcx/blob/main/aries_vcx/docs/guidelines.md#deterministic-transitions but given the user would be nudged to handle all of the possible conditions, maybe it's not bad?

gmulhearn-anonyome:
Hmm. Yeah, it's quite awkward to deal with this final ACK state..

Just brainstorming and building on Patrik's suggestions;

what about this:

```
enum HolderReceiveCredentialNextState {
     Finished(Holder<Finished>),
     AckPrepared(Holder<AckPrepared>)
}

impl Holder<RequestPrepared> {
     fn receive_credential(self, cred) -> Result<HolderReceiveCredentialNextState, (Holder<CurrentState>, Holder<Failed>)>
}

impl Holder<AckPrepared> {
    fn get_ack_message(&self) -> AckMessage

    // consumer manually calls this after they are confident they have transmitted the ack message? - a bit awkward.
    fn transition_to_finished(self) -> Holder<Finished>
}
```

bobozaur:
> E.g. the flow might look as followings:
> 
> ```
> // API signature
> async fn xyz(self,..) -> Result<Holder<NewState>, (Holder<CurrentState>, Holder<FailedState>)>
> 
> // consumer calls:
> let res = holder.xyz(..).await;
> 
> if let Err((holder, failed_holder)) = res {
>     if failed_holder.get_error() == AriesVcxError::LedgerTimeout {
>          // consumer considers ledger timeouts as a soft error and wants to retry
>          holder.xyz(...).await;
>          // ... etc
>     }
> }
> ```

Why would this ever happen? Maybe I'm missing something but I don't understand why `Holder<NewState>` ,`Holder<CurrentState>` and `Holder<Failed>` would all be possible in the same method. If the operation is meant to be retried, then don't return `Holder<Failed>` but rather some `Error`.

`Holder<Failed>` is meant to be an unrecoverable final state. If something returns it, it's not meant to be retried.

gmulhearn:
Right, my bad, I misunderstood your original push for using Holder<Failed> instead of AriesVcxError. I understand the motive now.

In that case, all the error results that can happen in this Holder typestate pattern (prepare request, and receive credential) are not HARD errors.

As they both could be a result of a ledger timeout or network error when fetching the cred def or rev reg states. Which could be considered a soft error in some use cases.

So they should effectively be;

`fn xyz(self,..) -> Result<Holder<NewState>, (Holder<currentState>, AriesVcxError)>`

bobozaur:
@gmulhearn Did you merge `main` into this branch? Or did you rebase? I'm asking because if you merged that messes up history and it would be best to rebase instead.

I had a very hard time rebasing a while ago after `aries_vcx_core` crate was split from `aries_vcx` and you did that regression fix. Idk if you merged back then too, but it would make sense as the problems I had were coming from the commits being different (because of the merge). Ignore this comment if it's unrelated 😅

gmulhearn:
@bobozaur i generally merge main into branches rather than rebase. But I don't see how that can cause issues since the PR should squash the branch when merging into master (unless we aren't squashing PRs?)

I generally prefer merge main into branch rather than rebase, as I'm not the biggest fan on re-writing history of a branch. But I know that's a bit of a political debate...

If it is an issue to merge instead of rebase on PR branches, could you explain why pls

bobozaur:
Yeah, after a closer look it seems to be completely unrelated. I don't really care that much, honestly, tho I personally prefer rebasing. My main concern was whether that caused the issue I had then, but it doesn't look like it.

I remember I rebased from the branch in #798 as I was integrating the `messages` crate and once that got merged into `main` and I wanted to rebase again all hell broke loose. Just noticed that you merged main in here and wondered if this was the issue if that's what you usually do.

But, like you said, you didn't even rebase/merge then. And as I'm writing this I realized what the issue was: I rebased from that PR branch, then the commits got squashed when merged into main, so the rebase was painful because of that. Should've reset to `main` and cherry pick, but oh well... The more you know 😅 .

In conclusion, do whatever you're comfortable with and ignore these comments 😄 .

gmulhearn:
Ahh yep, thanks for explaining, makes sense. Another personal reason I don't rebase is because I never seem to get it right... haha. Maybe I should learn how to do it properly.

PR #770, opened 2023-03-13
Remove Verifier's Proof method get_presentation_status(...)

Patrik-Stas:


aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs
```diff
@@ -1,15 +1,32 @@
...
+use crate::protocols::proof_presentation::verifier::verification_status::PresentationVerificationStatus;
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct FinishedState {
     pub presentation_request: Option<PresentationRequest>,
     pub presentation: Option<Presentation>,
     pub status: Status,
-    pub revocation_status: Option<RevocationStatus>,
+    #[serde(
+        default = "PresentationVerificationStatus::Unavailable",
```
mirgee, line 16: I assume this is why you gave the `Unavailable` variant parenthesis? How about we rather implement `Default` for `PresentationVerificationStatus` and use 
```
#[serde(default, ...)]
```
here?

Patrik-Stas: correct, that's reason for parenthesis

I'd prefer to keep this explicit , I think `default` should be used where it really somewhat conceptually make sense to make certain value default. But in this case, it's only temporary migration/compatibility reason.

Patrik-Stas: I removed empty tuple parameter on `Unavailable` because it was causing issue with de/serialization - it ended up being serialized as 
```
{ ... "verification_status": { "Unavailable": []} }
```
and even failed to be deserialized - not sure but looked like serde bug.

So anyway, I changed it but exactly as you suggested, instead I refer to function `verification_status_unavailable` as default builder in this case. Since it's only local case where we need this default, I think it's not appropriate global default for the enum.

aries_vcx/src/protocols/proof_presentation/verifier/verification_status.rs
```diff
@@ -0,0 +1,68 @@
...
+    Unavailable,
+}
+
+// todo: to be removed in 0.54.0, this supports legacy serialization when the enum had values "Revoked" and "NotRevoked"
+impl<'de> Deserialize<'de> for PresentationVerificationStatus {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s: String = Deserialize::deserialize(deserializer)?;
```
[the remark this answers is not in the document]
bobozaur: Also, you can alias variants directly, like: `#[serde(alias = "NonRevoked")]`. This will attempt deserialization from the variant name as well as the alias/es. Would be shorter than writing the `Deserialize` impl.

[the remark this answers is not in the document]
Patrik-Stas: Are you sure about `#[serde(alias = "NonRevoked")]`? Also, how would it know it should map `NonRevoked` to `Invalid` variant and `Revoked` as `Valid` variant?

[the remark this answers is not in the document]
Patrik-Stas: I've address the `String/&str` comment though

aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs
bobozaur: I don't think it's a serde bug. Why would the unit type be present in the `Unavailable` variant?

aries_vcx/src/protocols/proof_presentation/verifier/verification_status.rs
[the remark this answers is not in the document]
bobozaur: The attribute is added to each necessary variant, not on the enum itself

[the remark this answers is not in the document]
Patrik-Stas: Ah right, obivously :-) Thanks, updated

aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs
Patrik-Stas: Well I would expect that if a structure is serialized, it would be successfully deserialized back. Perhaps it was related to the custom deserialization?
Anyway, it doesn't matter anymore.

aries_vcx/src/protocols/proof_presentation/verifier/verification_status.rs
```diff
@@ -0,0 +1,64 @@
+use serde::{Deserialize, Deserializer};
+
+#[derive(Serialize, Debug, Clone, PartialEq, Eq)]
+pub enum PresentationVerificationStatus {
+    Valid,
+    Invalid,
+    Unavailable(),
```
bobozaur: I know this is removed, but I wanted to look it up. I didn't even know this is possible.
It's actually very interesting that you can declare variants like that, getting a constructor closure for free.

Nevertheless, in this case it is redundant. And yeah, the serialization -> deserialization roundtrip fails because the output changes. Apparently `serde` serializes this to `"Unavailable": []`, probably because in proc_macros tuple variants get treated as arrays of fields.

So apart from single field tuples, which are handled transparently, you end up with an array. Interestingly, I would expect the empty tuple variant to be set to `null` instead.

aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs
```diff
@@ -1,15 +1,37 @@
...
+use crate::protocols::proof_presentation::verifier::verification_status::PresentationVerificationStatus;
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct FinishedState {
     pub presentation_request: Option<PresentationRequest>,
     pub presentation: Option<Presentation>,
     pub status: Status,
-    pub revocation_status: Option<RevocationStatus>,
+    #[serde(
+        default = "verification_status_unavailable", // todo: to be removed in 0.54.0, this supports legacy serialization when the field was 'null'
```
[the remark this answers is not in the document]
bobozaur: I guess that makes sense. But given the narrow scope and short life of this, I'd use a narrower name for the sake of readability. It's declared in the local module anyway, so just naming the function `unavailable` would do, in my opinion.

aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs
```diff
@@ -1,15 +1,37 @@
...
+        deserialize_with = "null_to_unavailable" // todo: to be removed in 0.54.0, this supports legacy serialization when the field was 'undefined'
```
bobozaur, line 17: This is unnecessary. `serde` will use the `default` impl if the value is not found. Or, more accurately, it will never resort to `default` because the field's absence is treated in the custom deserialize function.

bobozaur (changes requested):

aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs
Patrik-Stas: It seems to be necessary, the behaviour is different when value is `null` and `undefined`. 

The `default` is used if is the field is undefined completely in the json.
But if `verification_status` is present and is `null`, without `#[serde(deserialize_with = "null_to_unavailable")]` it will fail tests
```
 thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected value", line: 1, column: 89)', aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs:170:80
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
FAILED
test protocols::proof_presentation::verifier::states::finished::unit_tests::test_verifier_state_finished_deser_legacy_values_verification_status ... thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected value", line: 1, column: 91)', aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs:144:80
FAILED
```

To save you look ups, line 170:
```
let serialized =
                r#"{"presentation":null,"presentation_request":null,"status":"Success","revocation_status":null}"#;
            let deserialized: FinishedState = serde_json::from_str(serialized).unwrap();
```
line 144
```
            let serialized =
                r#"{"presentation":null,"presentation_request":null,"status":"Success","verification_status":null}"#;
            let deserialized: FinishedState = serde_json::from_str(serialized).unwrap();
```

bobozaur: Fair point. But I believe that still makes the `default` redundant, as deserialization will just resort to `None` in the `deserialize` function and be handled accordingly.

Patrik-Stas: fair enough, removed `default`

bobozaur (dismissed):
LGTM

Patrik-Stas:
@bobozaur 
```
---- protocols::proof_presentation::verifier::states::finished::unit_tests::test_verifier_state_finished_deser_legacy_values_verification_status stdout ----
thread 'protocols::proof_presentation::verifier::states::finished::unit_tests::test_verifier_state_finished_deser_legacy_values_verification_status' panicked at 'called `Result::unwrap()` on an `Err` value: Error("missing field `verification_status`", line: 1, column: 68)', aries_vcx/src/protocols/proof_presentation/verifier/states/finished.rs:128:80
```
so nope, `#[serde(default = "verification_status_unavailable")]` must also stay, otherwise deserialization fails if neither`verification_status` nor `revocation_status` is specified.
And `#[serde(deserialize_with = "null_to_unavailable")]` must stay for the case when it is specified, but value is `null`

PR #778, opened 2023-03-18
Refactor `encode_attributes` function

### LINKED ISSUE #776: Refactor function `encode_attributes`
In file `encoding.rs` function  `pub fn encode_attributes(attributes: &str) -> VcxResult<String>` takes `&str`  as argument - it then tries to parse it to `Hashmap<String, serde_json::Value>`.

Task:
- change the function input and output types to be less generic than just `&str` and `String` - for example, it could look like 
```
pub fn encode_attributes(attributes: CredentialAttributes) -> VcxResult<CredentialAttributesEncoded>
```
and hence making the function declaration communicate more about the input it requires and output it's producing.

Obivously, this will have impact on all of the code calling the function, so you will have to update that appropriately. The main place where `encode_attributes` is being called from. You could transitively continue applying this typing improvements on calling functions.

styvane:
This PR closes #776

### DIFF
--- a/aries_vcx/src/common/credentials/encoding.rs
+++ b/aries_vcx/src/common/credentials/encoding.rs
@@ -1,99 +1,140 @@
-use crate::errors::error::{AriesVcxError, AriesVcxErrorKind, VcxResult};
-use crate::utils::openssl::encode;
 use std::collections::HashMap;
 
-pub fn encode_attributes(attributes: &str) -> VcxResult<String> {
-    let mut dictionary = HashMap::new();
-    match serde_json::from_str::<HashMap<String, serde_json::Value>>(attributes) {
-        Ok(mut attributes) => {
-            for (attr, attr_data) in attributes.iter_mut() {
-                let first_attr = match attr_data {
-                    // new style input such as {"address2":"101 Wilson Lane"}
-                    serde_json::Value::String(str_type) => str_type,
+use serde_json::Value;
 
-                    // old style input such as {"address2":["101 Wilson Lane"]}
-                    serde_json::Value::Array(array_type) => {
-                        let attrib_value: &str = match array_type.get(0).and_then(serde_json::Value::as_str) {
-                            Some(x) => x,
-                            None => {
-                                return Err(AriesVcxError::from_msg(
-                                    AriesVcxErrorKind::InvalidAttributesStructure,
-                                    "Attribute value not found",
-                                ));
-                            }
-                        };
+use crate::errors::error::{AriesVcxError, AriesVcxErrorKind, VcxResult};
+use crate::utils::openssl::encode;
 
-                        warn!("Old attribute format detected. See vcx_issuer_create_credential api for additional information.");
-                        attrib_value
-                    }
-                    _ => {
-                        return Err(AriesVcxError::from_msg(
-                            AriesVcxErrorKind::InvalidJson,
-                            "Invalid Json for Attribute data",
-                        ));
-                    }
-                };
+/// `CredentialAttributeData` represents contains all the attributes data
+/// in a credential.
+#[derive(Debug, Clone)]
+pub struct CredentialAttributeData<D> {
+    /// A reference to the attributes data in the credentials.
+    data: D,
+}
 
-                let encoded = encode(first_attr)?;
-                let attrib_values = json!({
-                    "raw": first_attr,
-                    "encoded": encoded
-                });
+/// `EncodedCredentialAttributes` contains the encoded credential attributes data.
+#[derive(Debug, Clone)]
+pub struct EncodedCredentialAttributes {
+    /// Encoded data text.
+    pub data: String,
+}
 
-                dictionary.insert(attr.to_string(), attrib_values);
-            }
-            serde_json::to_string_pretty(&dictionary).map_err(|err| {
-                warn!("Invalid Json for Attribute data");
-                AriesVcxError::from_msg(
-                    AriesVcxErrorKind::InvalidJson,
-                    format!("Invalid Json for Attribute data: {}", err),
-                )
-            })
-        }
-        Err(_err) => {
-            // TODO: Check error type
-            match serde_json::from_str::<Vec<serde_json::Value>>(attributes) {
-                Ok(mut attributes) => {
-                    for cred_value in attributes.iter_mut() {
-                        let name = cred_value.get("name").ok_or(AriesVcxError::from_msg(
-                            AriesVcxErrorKind::InvalidAttributesStructure,
-                            format!("No 'name' field in cred_value: {:?}", cred_value),
-                        ))?;
-                        let value = cred_value.get("value").ok_or(AriesVcxError::from_msg(
-                            AriesVcxErrorKind::InvalidAttributesStructure,
-                            format!("No 'value' field in cred_value: {:?}", cred_value),
-                        ))?;
-                        let encoded = encode(value.as_str().ok_or(AriesVcxError::from_msg(
-                            AriesVcxErrorKind::InvalidAttributesStructure,
-                            format!("Failed to convert value {:?} to string", value),
-                        ))?)?;
-                        let attrib_values = json!({
-                            "raw": value,
-                            "encoded": encoded
-                        });
-                        let name = name
-                            .as_str()
-                            .ok_or(AriesVcxError::from_msg(
-                                AriesVcxErrorKind::InvalidAttributesStructure,
-                                format!("Failed to convert attribute name {:?} to string", cred_value),
-                            ))?
-                            .to_string();
-                        dictionary.insert(name, attrib_values);
-                    }
-                    serde_json::to_string_pretty(&dictionary).map_err(|err| {
-                        warn!("Invalid Json for Attribute data");
-                        AriesVcxError::from_msg(
-                            AriesVcxErrorKind::InvalidJson,
-                            format!("Invalid Json for Attribute data: {}", err),
-                        )
-                    })
+impl<D> CredentialAttributeData<D>
+where
+    D: AsRef<str>,
+{
+    /// Create new attributes data.
+    pub fn new(data: D) -> Self {
+        Self { data }
+    }
+
+    /// Encodes the attributes in a credential.
+    pub fn encode(&self) -> VcxResult<EncodedCredentialAttributes> {
+        let mut dictionary = HashMap::new();
+        match serde_json::from_str::<HashMap<String, serde_json::Value>>(self.data.as_ref()) {
+            Ok(attributes) => Self::encode_table_attribute(attributes, &mut dictionary),
+            Err(_err) => {
+                // TODO: Check error type
+                match serde_json::from_str::<Vec<serde_json::Value>>(self.data.as_ref()) {
+                    Ok(attributes) => Self::encode_attribute_list(attributes, &mut dictionary),
+                    Err(err) => Err(AriesVcxError::from_msg(
+                        AriesVcxErrorKind::InvalidAttributesStructure,
+                        format!("Attribute value not found: {:?}", err),
+                    )),
                 }
-                Err(err) => Err(AriesVcxError::from_msg(
-                    AriesVcxErrorKind::InvalidAttributesStructure,
-                    format!("Attribute value not found: {:?}", err),
-                )),
             }
         }
+    }
+
+    /// Encodes attributes in a hashmap.
+    fn encode_table_attribute(
+        attributes: HashMap<String, Value>,
+        dictionary: &mut HashMap<String, Value>,
+    ) -> VcxResult<EncodedCredentialAttributes> {
+        for (attr, attr_data) in attributes {
+            let first_attr = match &attr_data {
+                // new style input such as {"address2":"101 Wilson Lane"}
+                serde_json::Value::String(str_type) => str_type,
+
+                // old style input such as {"address2":["101 Wilson Lane"]}
+                serde_json::Value::Array(array_type) => {
+                    warn!("Old attribute format detected. See vcx_issuer_create_credential api for additional information.");
+                    array_type.get(0).and_then(serde_json::Value::as_str).ok_or_else(|| {
+                        AriesVcxError::from_msg(
+                            AriesVcxErrorKind::InvalidAttributesStructure,
+                            "Attribute value not found",
+                        )
+                    })?
+                }
+                _ => {
+                    return Err(AriesVcxError::from_msg(
+                        AriesVcxErrorKind::InvalidJson,
+                        "Invalid Json for Attribute data",
+                    ));
+                }
+            };
+
+            let attrib_values = json!({
+                "raw": first_attr,
+                "encoded": encode(first_attr)?
+            });
+
+            dictionary.insert(attr, attrib_values);
+        }
+        let data = serde_json::to_string_pretty(&dictionary).map_err(|err| {
+            warn!("Invalid Json for Attribute data");
+            AriesVcxError::from_msg(
+                AriesVcxErrorKind::InvalidJson,
+                format!("Invalid Json for Attribute data: {}", err),
+            )
+        })?;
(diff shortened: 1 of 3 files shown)


aries_vcx/src/common/credentials/encoding.rs
```diff
@@ -1,99 +1,140 @@
...
+    /// Create new attributes data.
+    pub fn new(data: D) -> Self {
+        Self { data }
+    }
+
+    /// Encodes the attributes in a credential.
+    pub fn encode(&self) -> VcxResult<EncodedCredentialAttributes> {
+        let mut dictionary = HashMap::new();
+        match serde_json::from_str::<HashMap<String, serde_json::Value>>(self.data.as_ref()) {
+            Ok(attributes) => Self::encode_table_attribute(attributes, &mut dictionary),
```
Patrik-Stas, line 36: Rather than passing mutable dictionary in, you can allocate it in the encoding functions and return it, it's cleaner

aries_vcx/src/common/credentials/encoding.rs
```diff
@@ -1,99 +1,140 @@
...
-                let attrib_values = json!({
-                    "raw": first_attr,
-                    "encoded": encoded
-                });
-
-                dictionary.insert(attr.to_string(), attrib_values);
+/// `CredentialAttributeData` represents contains all the attributes data
+/// in a credential.
+#[derive(Debug, Clone)]
+pub struct CredentialAttributeData<D> {
```
Patrik-Stas, line 11: This is currently as well pretty much just wrapping something what can be turned into `&str`, so lets either. I suggest you keep `encode` input signature as was, simply taking `&str`. 

Nevertheless, it would be actually amazing if the input argument for `encode_attributes` can be  dealt with properly. We can make that separate issue, separate PR if you'd like. Here's bit more detailed suggestion how to go about it:

The main issue is that `encode_attributes` supports 3 different input formats are. To make the `encode` function signature nicer (and implementation too) we need to fix the problem from its root. 
The 3 different formats currently supported:
```
{"address2":"101 Wilson Lane", "surname": "Smith"}
```
```
{"address2":["101 Wilson Lane"], "surname": ["Smith"]}
```
```
[{"name": "address2", "value": "101 Wilson Lane"}, {"name": "surname", "value": "Smith"}]
```
and apparently, the current implementation just does not care up until `encode` is called. Not only this makes it hard to create input type for `encode_attributes` function to accept, it also doesn't follow the "fail early" principle. 

So here's idea:
- find original place where this ambiguous string data is originating from - hint: it's when issuer is building credential offer msg `Issuer::build_credential_offer_msg`; `issuer.rs`) and parse the data there
- let's select the most sensible format and internally stick with that - probably `{"address2":"101 Wilson Lane", "surname": "Smith"}` makes sense the most. And therefore we can also type it less ambiguously than just `String` - we can now work with `HashMap<String, String>` instead.
- we can then therefore update `OfferSentState`, `RequestReceivedState` to work with that instead of `String`.
- the `encode_attributes` can then take `HashMap<String, String>` instead of `String` as well, and the implementation will be much simpler as we have compile time guarantee about the data we'll be processing.
Unless we implement some deserialization, this change will be breaking change for state machines in `OfferSentState`, `RequestReceivedState` states, but arguably it's low impact and acceptable.

Patrik-Stas (changes requested):
Thanks for this @styvane much appreciated! I left comments, lets yet improve typing on output data in this PR. I also left comment about the input data typing - if you'd like to work on it, you can choose to do it in this PR or separate :-)

aries_vcx/src/common/credentials/encoding.rs
styvane: If we enumerate all the currently supported formats, we could limit the number of breaking changes in the state machines.

styvane:
> Thanks for this @styvane much appreciated! I left comments, lets yet improve typing on output data in this PR. I also left comment about the input data typing - if you'd like to work on it, you can choose to do it in this PR or separate :-)

Thank you for the review @Patrik-Stas. I intend to continue working on this issue and contribute more to the project. We can tackle all these points in this PR.

aries_vcx/src/common/credentials/encoding.rs
```diff
@@ -1,99 +1,140 @@
...
+    /// A reference to the attributes data in the credentials.
+    data: D,
+}
+
+/// `EncodedCredentialAttributes` contains the encoded credential attributes data.
+#[derive(Debug, Clone)]
+pub struct EncodedCredentialAttributes {
```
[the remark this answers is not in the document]
styvane: Hum, having raw data in an encoded type is a bit confusing IMO. What do you think about this?

```rust
pub CredentialAttribute<State> {
    encoded: Option<String>,
    raw: String,
   _marker: PhantomData<State>,
   
}

pub struct  EncodedAttribute;

pub struct RawAttribute;

impl CredentialAttribute<RawAttribute> {
    fn encode(...) -> ...
}

impl CredentialAttribute<Encoded> {
    fn encoded(...) -> ...
}
```

aries_vcx/src/common/credentials/encoding.rs
Patrik-Stas: That's true, but the implementation will be a lot simpler if we drop it, the impact won't be significant. Speaking for Absa, we don't mind and I am not aware of other parties currently using Issuer state machine (aries-vcx users seem to be typically on mobile side building native apps)

Quoting myself
> but arguably it's low impact and acceptable.
to elaborate on this, the state machines would rarely be long-term stored in `OfferSentState`, `RequestReceivedState ` long term. Usually entire credential exchange happens within a seconds.

aries_vcx/src/common/credentials/encoding.rs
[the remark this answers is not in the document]
styvane: We need the `PhantomData` here to workaround unused generic type error since we don't have a field with type `State` in `CredentialAttribute`, but we still want to say that `CredentialAttribute` has a state.

Patrik-Stas:
Hi @styvane just a small update/request.

1. Let's keep the scope of this PR as small as possible - let's address the return value first. 
2. Fixing the input of the function has longer roots, as we've dicussed, so let's keep that for separate issue. We are planning big rewrite of state machines, so we'll need to consider the timing for these changes. But I would like to avoid situation where you've done useful rework around the output value, but the changes to state machines would conflict with an other ongoing PR. So let's do the 1. (the output)  first, then let's consider addressing 2. (the input)

My apologies I underestimated scope of the work when writing issue #776. I only realized after your submission that proper solution so solving the issue with the input types is streching across multiple files of issuer state machine.

aries_vcx/src/common/credentials/encoding.rs
```diff
@@ -1,99 +1,157 @@
...
+}
+
+/// `RawAttributeValue` is a credential attribute raw state.
+pub struct RawAttributeValue;
+
+/// `RawAttributeValue` is a credential attribute in encoded state.
+pub struct EncodedAttributeValue;
+
+impl<'raw> CredentialAttribute<'raw, RawAttributeValue> {
+    /// Create new credential attribute.
```
Patrik-Stas, line 28: The function name is self explanatory, I think we can drop these comments

aries_vcx/src/common/credentials/encoding.rs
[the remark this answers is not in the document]
Patrik-Stas: Sorry but I don't quite follow. What state does it have? What kind of programmer error are you trying to prevent in compile time by using the `State` generics arg?

aries_vcx/src/common/credentials/encoding.rs
```diff
@@ -1,99 +1,157 @@
...
-                        warn!("Old attribute format detected. See vcx_issuer_create_credential api for additional information.");
-                        attrib_value
+/// `CredentialAttribute` contains credential attributes data in a given state.
+#[derive(Debug, Clone)]
+pub struct CredentialAttribute<'r, State> {
+    /// Raw credential attribute data.
+    raw: &'r str,
+
+    /// Encoded credential attribute data.
+    encoded: Option<String>,
```
Patrik-Stas, line 16: Let's get rid of `Option` here. It's `None` only temporarily when calling `new` function. But since it's now `Option`, we don't have quarantee it's not `None` when we return `CredentialAttribute`.

aries_vcx/src/common/credentials/encoding.rs
```diff
@@ -299,35 +365,37 @@ pub mod unit_tests {
             }
         });
 
-        static TEST_CREDENTIAL_DATA: &str = r#"{"address2":["101 Wilson Lane"],
+        let attr_data = CredentialAttribute::new(
```
Patrik-Stas, line 368: This doesn't seem to make quite sense, `CredentialAttribute` as I understand currently refers to 1 attribute (with raw and encoded value), but here you are passing entire json of attributes and their respective raw values.

Patrik-Stas (changes requested):
Hi @styvane, I've updated review. Sorry for somewhat slower feedback time

aries_vcx/src/common/credentials/encoding.rs
styvane: Yes, I was not sure of the option as well because the you can only access the encoded value when you have `AttributeValue<EncodedAttributeValue>`.

aries_vcx/src/common/credentials/encoding.rs
styvane: You're right, and my apologies on this one. I mixed things here. I will get that fixed.

aries_vcx/src/common/credentials/encoding.rs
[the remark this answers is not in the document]
styvane: This is to prevent an API misused. One cannot get encoded data unless they are in an encoded state.

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
```diff
@@ -557,9 +557,15 @@ async fn _create_credential(
         ));
     };
     let request = &request.requests_attach.content()?;
-    let cred_data = encode_attributes(cred_data)?;
+    let cred_attrs = CredentialAttribute::new(cred_data).encode()?;
```
Patrik-Stas, line 560: Note similar to my comment around tests, this also seems to be bit of type mismatch here, as `CredentialAttribute` would represent just a single attribute

aries_vcx/src/common/credentials/encoding.rs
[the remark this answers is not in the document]
Patrik-Stas: Ah I see I get it now - only 
```
CredentialAttribute<RawAttributeValue>
```
implements `encode` functions, while only 
```
CredentialAttribute<EncodedAttributeValue>
```
implements 
```
pub fn encoded(&self) -> VcxResult<&str>
```
So with `CredentialAttribute<EncodedAttributeValue>` we essentially do get the compile time guarantee that `encoded: Option<String>` is `Some`.
It's smart, I haven't used `PhantomData` much in past. Nice!

However I still believe it would be easier to use if the `Option` could be removed - consider if there's a reasonable way to do it, nevertheless I think we can keep the current phantom data approach as well.

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
styvane: I pushed another commit to address my mistake. I added a `CredentialAttributes::encode_all` API to encode all attributes in a credential. 

`CredentialAttributes` stores the attributes in an `attributes` table. However, using a vector will be better IMHO, but changing this breaks the tests, so I left it as it's now.

aries_vcx/src/common/credentials/encoding.rs
bobozaur, line 1: What is the point of having a generic type? If there are only two states with no common code, why not separate the types entirely?

bobozaur (changes requested):

aries_vcx/src/common/credentials/encoding.rs
styvane: The generic type is not for code de-duplication. It's to provide better API.

Because the credential attribute does not change, what change is the state. We can provide different APIs when a credential attribute is in some concrete state.  Having two separate types gives the impression that we have two different kinds of attributes.

bobozaur: But from what I see the states themselves hold no additional data. Having them as generic parameters to a parent structure seems to unnecessarily complicate the API with no real benefit.

Why not have something like `EncodedCredAttr` and `RawCredAttr` as two separate structures altogether? This gets rid of the `PhantomData` as there's no need to bind a type anymore.

I would see the benefit of the typestate pattern being used here if this was some sort of container, the states brought by a different set of data or there was common functionalities between multiple types.

Unrelated, but as a bonus, in situations where you would do something like this it's best to choose something else as a type parameter for `PhantomData`. `T` is generally used when the structure does in fact *own* the data, where as in other circumstances other covariants are preferred.

I generally use `fn() -> T` because it's still a covariant, yet without `dropck` enforcement, and it implements `Send` and `Sync`. 

These resources explain this in a fair amount of detail:
 - <https://doc.rust-lang.org/std/marker/struct.PhantomData.html>
 - <https://doc.rust-lang.org/nomicon/phantom-data.html>
 - <https://doc.rust-lang.org/nomicon/dropck.html>

bobozaur (changes requested):

aries_vcx/src/common/credentials/encoding.rs
styvane: I'm encoding the attribute's state into a state machine, so I think we shouldn't separate this into two types. However, I will update the PR with your suggestion if that is how we should do it.

I'm aware of these docs, and I know that carrying around `PhantomData` is sometimes annoying.

bobozaur: It's still two types, they're just wrapped. Generally I'd do that if I wanted to take advantage of monomorphization or the state carried some data. But in this case using two "entirely distinct" structs to encode the state in the state machine would be no different.

I'm not trying to nitpick or anything, so please don't feel discouraged :). I would be glad if you provided an argument over why your approach is better and changed my mind. I just don't really see it myself.

Patrik-Stas:
Hi @styvane, I know this has been quite a long journey so I hope we didn't get you too exhausted, your contribution is already valuable! Yet I need to ask you, are you planning to further iterate on @bobozaur's review? If yes, it's all good and we can wait - this is not in hurry. If not we'll discuss this PR on Thursday's call and decide the way forward.

styvane:
> Hi @styvane, I know this has been quite a long journey so I hope we didn't get you too exhausted, your contribution is already valuable! Yet I need to ask you, are you planning to further iterate on @bobozaur's review? If yes, it's all good and we can wait - this is not in hurry. If not we'll discuss this PR on Thursday's call and decide the way forward.

Hi @Patrik-Stas, I'm planning to address the changes @bobozaur suggested, and the last points you mentioned, but I'm currently away from computers and will do it in the first week of May.

Patrik-Stas:
Cool! Thanks for update 👍

styvane:
> Cool! Thanks for update 👍

Hi @Patrik-Stas, unfortunately, I will not continue working on this PR. Thank you for all the reviews.

Patrik-Stas:
hi @styvane all good, may I ask you to grant me rights to push into your fork? I would still like to get this PR in, but it requires at least resolving conflicts and perhaps addressing some of Bogdan's concerns.

styvane:
> hi @styvane all good, may I ask you to grant me rights to push into your fork? I would still like to get this PR in, but it requires at least resolving conflicts and perhaps addressing some of Bogdan's concerns.

I've added you to the fork, thanks

Patrik-Stas:
The problems with encoding (supporting multiple formats, string ambuguity) solved as part of implementation of v2 issuance protocol https://github.com/hyperledger/aries-vcx/pull/987/files

Closing this, the implementation will stay local for v1 issuance protocol as is, with all its flaws.

PR #781, opened 2023-03-23
Add state machines implementation guidelines

Patrik-Stas:
Per discussions on the last 2 community calls

https://wiki.hyperledger.org/display/ARIES/2023-03-23+Aries-vcx+community+call
https://wiki.hyperledger.org/display/ARIES/2023-03-16+Aries-vcx+community+call

It will be important to stay aligned on the approach as we are preparing to rewrite state machines with the state pattern.

### DIFF
--- a/aries_vcx/docs/guidelines.md
+++ b/aries_vcx/docs/guidelines.md
@@ -1,0 +1,71 @@
+# Guidelines for implementing protocol state machines
+
+### No 'Sent' states
+Do not create states to track whether a reply has been sent to counterparty. It is responsibility of
+`aries-vcx` consumers to keep track of whether response message has been sent to counterparty. From state
+machine perspective, creation of a protocol message implies a reply to that message can be processed.
+
+Example: Do not create states such as `PresentationSent` (in context of `presentation-protocol` in role of prover).
+Instead, only create state `PresentationPrepared` which enables `aries-vcx` consumers to get the presentation message
+and send it to the counterparty. Subsequently, it should be possible to further drive state machine in  `PresentationPrepared` 
+using `presentation-ack` message received from verifier. 
+
+### Separate final states
+While Aries RFCs commonly refer to single final state to convey nothing more can happen after exchange
+of certain messages - for example in diagram [here](https://github.com/hyperledger/aries-rfcs/blob/main/features/0453-issue-credential-v2/README.md)
+- on implementation level we want to distinguish different internal states signalling different circumstances (which 
+simply happen to share the property of being final).
+
+Example: instead of having `Final` state which would maintain `status: Success | Failed | Rejected`, we
+should have 3 separate states `Success`, `Failed`, `Rejected`.
+
+### Receives Aries messages always move state
+Processing a message from counterparty must move state. Even if we are dealing just with acknowledgement type of message.
+
+Example: In `issue-credential` protocol, do not enter `Finished` simply after building verifiable credential.
+You might rather have `CredentialBuilt` and only enter `Finished` state after receiving `ack` message, which
+is part of `issue-credential` protocol - as opposed of simply tracking acknowledgement within the `Finished` state.
+
+### Use intermediate states when needed
+Aries RFCs describe the protocol state from perspective of external viewer. While we need to adhere the protocol
+in terms of what messages does the counterparty accept at a given moment of the protocol lifecycle, as library
+developers, we are also concerned about the APIs we expose and its properties. It proved to be useful to create
+intermediate states that do not exhibit a strict bijection with states described by aries RFCs.
+
+Example: In [`present-proof 1.0`](https://github.com/hyperledger/aries-rfcs/blob/main/features/0037-present-proof/README.md)
+protocol, there's following states described:
+```
+request-received
+proposal-sent
+presentation-sent
+reject-sent
+done
+```
+However, in practice it's useful to have not only state `request-received` but also `presentation-prepared`. It's
+also useful to have multiple final states than just `done`, for example `Verified`, `NotVerified`, `Failed`.
+
+On the flip side, as per first point in this list `"No 'Sent' states"` - it's better to avoid creating "sent" states like
+the suggested byt the RFCs
+```
+proposal-sent
+presentation-sent
+reject-sent
+```
+
+However, note that in with our custom states, it must be always possible to determine where we are within the protocol
+as described by the RFC - in other words, what messages we accept in given moment. For example, our 
+state `presentation-prepared` would reflect RFC state `presentation-sent` where we are ready 
+to receive `presentation-ack` from verifier. 
+Equally, our custom states `Verified`, `NotVerified`, `Failed` map to RFC state `done`.
+
+### Deterministic transitions
+Transition should either fail and leave the current state unmodified, or transition to predetermined new state.
+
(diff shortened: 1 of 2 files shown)


bobozaur:
Probably the most notable takeaways would be:
- **Failure states** -> Failed states could be represented as `Err(StateMachine<FailedState>)` variants of a `Result`. We can then treat the failed state types as errors and handle them however it seems appropriate. The success variant would thus look something like `Ok(StateMachine<SuccessState>)`.
- **Reduced I/O** -> Most state machines have states of the sort `MessageSent`. However, when thinking about it a bit, simply saying we sent something doesn't really mean that much since we cannot guarantee it was also received, especially in a transport abstract manner. Therefore, we could model the state machines so that we work on `MessageGenerated` states, which we can guarantee, and then the consumer code can retrieve the message and send it (as many times as they want). Then the state machine would advance when the reply to the sent message is received (if one is expected).

aries_vcx/docs/guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+Transition should either fail and leave the current state unmodified, or transition to predetermined new state.
+
+Example: in our legacy implementations, some transitions lead to multiple states, and the actual
+new state was determined in runtime. The typical case was that transition either succeeded and moved to
+the "happy-case" next state; or failed and moved to "failed" state. 
+
+## Problem reports
+State machines should not automatically submit problem-reports. If a problem occurs upon execution of a transition,
+it should fail with error (per previous point) and should be left up to state machine user to decide: retry,
+do nothing, move to failed state, move to failed state and send problem report.
```
bobozaur, line 71: It would probably be desirable to create the `ProblemReport` ourselves and pass it to consumer code. They could then send it if they want to.

bobozaur (commented):
Just comments, really. Overall it looks good!

aries_vcx/docs/guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+While Aries RFCs commonly refer to single final state to convey nothing more can happen after exchange
+of certain messages - for example in diagram [here](https://github.com/hyperledger/aries-rfcs/blob/main/features/0453-issue-credential-v2/README.md)
+- on implementation level we want to distinguish different internal states signalling different circumstances (which 
+simply happen to share the property of being final).
+
+Example: instead of having `Final` state which would maintain `status: Success | Failed | Rejected`, we
+should have 3 separate states `Success`, `Failed`, `Rejected`.
+
+## Receives Aries messages always move state
+Processing a message from counterparty must move state. Even if we are dealing just with acknowledgement type of message.
```
mirgee, line 23: IMO the claim in the heading is too strong (confusingly so):
* just receiving a message by itself shouldn't move state, but processing it should,
* this obviously doesn't apply to final states (so "always" is too much).
Perhaps an alternative (although still imperfect) would be something like: "Protocol-based expectation of message receipt implies non-final state".

aries_vcx/docs/guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+`aries-vcx` consumers to keep track of whether response message has been sent to counterparty. From state
+machine perspective, creation of a protocol message implies a reply to that message can be processed.
+
+Example: Do not create states such as `PresentationSent` (in context of `presentation-protocol` in role of prover).
+Instead, only create state `PresentationPrepared` which enables `aries-vcx` consumers to get the presentation message
+and send it to the counterparty. Subsequently, it should be possible to further drive state machine in  `PresentationPrepared` 
+using `presentation-ack` message received from verifier. 
+
+## Separate final states
+While Aries RFCs commonly refer to single final state to convey nothing more can happen after exchange
```
mirgee, line 14: I would say "sometimes" instead of "commonly", which may be too strong. For example. did exchange protocol has 2 terminal states.

aries_vcx/docs/guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+proposal-sent
+presentation-sent
+reject-sent
+done
+```
+However, in practice it's useful to have not only state `request-received` but also `presentation-prepared`. It's
+also useful to have multiple final states than just `done`, for example `Verified`, `NotVerified`, `Failed`.
+
+On the flip side, as per first point in this list `"No 'Sent' states"` - it's better to avoid creating "sent" states like
+the suggested byt the RFCs
```
mirgee, line 48: "by the"

aries_vcx/docs/guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+```
+proposal-sent
+presentation-sent
+reject-sent
+```
+
+However, note that in with our custom states, it must be always possible to determine where we are within the protocol
```
mirgee, line 55: I would prefer the terms "internal" and "external" states established earlier (talking about custom states might sound like we are violating the protocols to some).

aries_vcx/docs/guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+
+Example: Do not create states such as `PresentationSent` (in context of `presentation-protocol` in role of prover).
+Instead, only create state `PresentationPrepared` which enables `aries-vcx` consumers to get the presentation message
+and send it to the counterparty. Subsequently, it should be possible to further drive state machine in  `PresentationPrepared` 
+using `presentation-ack` message received from verifier. 
+
+## Separate final states
+While Aries RFCs commonly refer to single final state to convey nothing more can happen after exchange
+of certain messages - for example in diagram [here](https://github.com/hyperledger/aries-rfcs/blob/main/features/0453-issue-credential-v2/README.md)
+- on implementation level we want to distinguish different internal states signalling different circumstances (which
```
mirgee, line 16: Does this apply only to the final states?

aries_vcx/docs/guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+
+## No 'Sent' states
+Do not create states to track whether a reply has been sent to counterparty. It is responsibility of
+`aries-vcx` consumers to keep track of whether response message has been sent to counterparty. From state
+machine perspective, creation of a protocol message implies a reply to that message can be processed.
+
+Example: Do not create states such as `PresentationSent` (in context of `presentation-protocol` in role of prover).
+Instead, only create state `PresentationPrepared` which enables `aries-vcx` consumers to get the presentation message
+and send it to the counterparty. Subsequently, it should be possible to further drive state machine in  `PresentationPrepared` 
+using `presentation-ack` message received from verifier.
```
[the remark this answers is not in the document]
Patrik-Stas: Thanks! added additional note

aries_vcx/docs/guidelines.md
Patrik-Stas: While I addressed your comment, I've reached for simpler wording - let me know if you think thins works

aries_vcx/docs/guidelines.md
Patrik-Stas: Not necessarily, this is now covered by section `Creating new states`

aries_vcx/docs/guidelines.md
Patrik-Stas: It's what I meant but the sentence was a bit ambiguous. Hopefully it's clearer now

mirgee (approved):

PR #785, opened 2023-03-28
Profiles refactoring

Patrik-Stas:
- renamed `modular_wallet_profile` to `modular_libs_profile` - it seem confusing to mention "wallet" while the profile encapsulates also credex, ledger client

- Modified `IndyVdrLedger` to store require and store minimal amount of information needed - it doesn't need entire profile, it doesn't need to know about exists of Profiles, it only needs `dyn BaseWallet` for transaction signing
- Similarly modified vdr-tools trait implementations to be unaware of `Profile`, but inject `wallet_handle`, `pool_handle` in them instead

### DIFF
--- a/agents/rust/aries-vcx-agent/src/agent/init.rs
+++ b/agents/rust/aries-vcx-agent/src/agent/init.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use aries_vcx::{
     agency_client::{agency_client::AgencyClient, configuration::AgentProvisionConfig},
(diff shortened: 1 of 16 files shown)


gmulhearn (approved):
Nice refactor, will make profiles cleaner to use!
Looks good.

If you wanna take it another step, you could now change the signature of the Profile trait params to &self rather than Arc<Self>. And then consequently, you could remove ALOT of Arc::clones of the Profile around the entire codebase.

mirgee (approved):

PR #787, opened 2023-03-30
Refactor/replacing messages

bobozaur:
Draft for the new `messages` crate integration.

Status:
    - [x] Integrated in `aries_vcx` - compiles
    - [x] Integrated in `libvcx` - compiles
    - [x] Integrated in the Rust agent - compiles.
    - [x] Integrated in the UniFFI wrapper - does not compile.
    - [x] Integrated in the `napi-rs` wrapper - compiles. 
    - [x] Test backwards compatibility when `aries_vcx` is used for a server.
    - [x] Test backwards compatibility when `aries_vcx` is used for a client.
    - [x] Removed feature flags used for testing from `aries_vcx`. 
    - [x] Accommodate `aries_vcx` integration integration tests. 

#### UniFFI
~The UniFFI wrapper doesn't compile because of a data type change in some parts of the code where instead of a `String` a `Url` is expected now. I tried to implement the `Url` in the UDL file, but it seems like adding an external type requires it's entire definition, as well as the definition of it's subtypes. And at some point, the `Url` depends on `std::net::Ipv4Addr` and `std::net::Ipv6Addr` which are wrappers over static arrays, which don't seem to be supported in UniFFI.~

This has been fixed by accepting a `String` instead of an `AriesDidDoc`

#### NOTE: 
This PR is simply to get rid of the old messages and implement the new ones. It's not performant, readable, or anything like that. I cut as many corners as I could because the real work will come with the `aries_vcx` and its state machines refactor. The goal here is to have things working as before, nothing more.


wrappers/node/package-lock.json
Patrik-Stas, line 1: should not be removed

wrappers/ios_legacy/ci/build.sh
```diff
@@ -14,7 +14,7 @@ OUTPUT_DIR=/tmp/artifacts
 setup() {
     echo "ios_legacy/ci/build.sh: running setup()"
     echo "Setup rustup"
-    rustup default 1.64.0
+    rustup default 1.65.0
```
Patrik-Stas, line 17: we should try to make small PR with this change in isolation to check if this is the cause of failing ios CI - this is main suspect probably

wrappers/node/package-lock.json
bobozaur: Does it matter? Doesn't it get recreated?

wrappers/ios_legacy/ci/build.sh
Patrik-Stas: Looks innocent https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#splitting-linux-debuginfo but probably main suspect here, as it's the only change dealing with linker

bobozaur: I looked at the release notes. That's for linux. It was already enabled for macOS.

bobozaur: This is now (finally) fixed.

wrappers/node/package-lock.json
bobozaur: Added back.

aries_vcx/src/common/proofs/prover/prover_internal.rs
```diff
@@ -386,30 +387,32 @@ pub mod unit_tests {
...
-                .unwrap_err()
-                .kind();
-            assert_eq!(err_kind, AriesVcxErrorKind::InvalidProofCredentialData);
-        })
-        .await;
-    }
+    // Idk why this is failing and at this point I'm too afraid to ask
+    //
+    // #[tokio::test]
+    // async fn test_find_credential_def_fails() {
```
Patrik-Stas, line 393: seem to be passing consistently now

aries_vcx/src/protocols/connection/inviter/states/requested.rs
```diff
@@ -26,10 +27,6 @@ impl TheirDidDoc for Requested {
 impl ThreadId for Requested {
     //TODO: This should land in the threadlike macro.
```
Patrik-Stas, line 28: perhaps not valid todo anymore

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
```diff
@@ -236,39 +271,42 @@ impl IssuerSM {
...
+                        return Some((uid, message));
                     }
                 }
-                IssuerFullState::OfferSent(_) => match message {
-                    A2AMessage::CredentialRequest(credential) => {
-                        if credential.from_thread(&self.thread_id) {
-                            return Some((uid, A2AMessage::CredentialRequest(credential)));
+                IssuerFullState::OfferSent(_) => match &message {
+                    AriesMessage::CredentialIssuance(CredentialIssuance::RequestCredential(msg)) => {
+                        if matches_opt_thread_id!(msg, self.thread_id.as_str()) {
```
Patrik-Stas, line 280: This is slightly downgrade, as at least `thread` decorator is known to be always required to present on a message. The protocol can not start from `credential-request` message and is always reply to `credential-offer`. So I think it can be safely said it should always have a thread.

What do you think? Just to clarify, I am not suggesting to address it here.

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
```diff
@@ -441,9 +479,11 @@ impl IssuerSM {
                 )
                 .await
                 {
-                    Ok((credential_msg, cred_rev_id)) => {
-                        let credential_msg = credential_msg.set_thread_id(&self.thread_id).ask_for_ack(); // TODO: Make configurable
-                        send_message(credential_msg.to_a2a_message()).await?;
+                    Ok((mut credential_msg, cred_rev_id)) => {
+                        credential_msg.decorators.thread.thid = self.thread_id.clone();
+                        credential_msg.decorators.please_ack = Some(PleaseAck::new(vec![])); // ask_for_ack sets this to an empty vec
```
Patrik-Stas, line 484: Weird, that would definitely deserve separate issue/fix.

aries_vcx/src/protocols/proof_presentation/verifier/states/presentation_proposal_received.rs
```diff
@@ -1,14 +1,31 @@
...
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct PresentationProposalReceivedState {
-    pub presentation_proposal: PresentationProposal,
-    pub presentation_request: Option<PresentationRequest>,
+    pub presentation_proposal: ProposePresentation,
+    pub presentation_request: Option<RequestPresentation>,
+}
+
+impl Default for PresentationProposalReceivedState {
```
Patrik-Stas, line 13: It seems we could avoid this Default if on line 31 we set `presentation_request: None` instead of `..Self::default()`

aries_vcx/src/protocols/proof_presentation/verifier/states/presentation_request_set.rs
```diff
@@ -1,13 +1,29 @@
...
-use messages::protocols::proof_presentation::presentation_request::PresentationRequest;
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct PresentationRequestSetState {
-    pub presentation_request: PresentationRequest,
+    pub presentation_request: RequestPresentation,
+}
+
+impl Default for PresentationRequestSetState {
```
Patrik-Stas, line 13: compiles without `Default`

aries_vcx/src/utils/author_agreement.rs
```diff
@@ -107,10 +107,10 @@ mod unit_tests {
         assert_eq!(expected_meta, meta);
     }
 
-    #[test]
-    fn get_txn_author_agreement_works_for_not_set() {
-        let _setup = SetupDefaults::init();
+    // #[test]
+    // fn get_txn_author_agreement_works_for_not_set() {
```
Patrik-Stas, line 111: passing now

uniffi_aries_vcx/core/src/vcx.udl
```diff
@@ -37,32 +37,6 @@ dictionary PairwiseInfo {
     string pw_vk;
 };
 
-dictionary AriesDidDoc {
```
Patrik-Stas, line 40: just checking, I guess this is what you were discussing with George on Discord, right?

Patrik-Stas (approved):
left minor comments but 🟢 from me

- This PR has been tested thoroughly outside of this repo as well
- Many unit tests is commented out but many can definitely be brought back. Can actually be good-first-issues to restore the individual suites

uniffi_aries_vcx/core/src/vcx.udl
gmulhearn-anonyome: lgtm, matches my interpretation of the conversation. Moving back to using `string` for aries_did_doc due to uniffi typing issues that are outside the scope of the PR to fix.

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
bobozaur: It was optional before, so I left it like that.
Per the RFC, it is accurate though. 
https://github.com/hyperledger/aries-rfcs/blob/main/features/0036-issue-credential/README.md#messages

uniffi_aries_vcx/core/src/vcx.udl
bobozaur: It is, yes.

aries_vcx/src/protocols/connection/inviter/states/requested.rs
bobozaur: Mother of all nitpicks, Patrik 😅 .

The whole state machine would get a pass through during the upcoming refactor (even if it had the type state pattern implemented already). 
It's not valid anymore, but if we start discussing invalid/obsolete/bad/inefficient/illegible stuff in the code this PR will never end.

aries_vcx/src/protocols/proof_presentation/prover/states/presentation_proposal_sent.rs
```diff
@@ -1,12 +1,28 @@
...
+};
+use uuid::Uuid;
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct PresentationProposalSent {
-    proposal: PresentationProposal,
+    proposal: ProposePresentation,
+}
+
+impl Default for PresentationProposalSent {
```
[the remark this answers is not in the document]
bobozaur: It implemented Default before, so I implemented it as well.
The purpose of this PR is the match the previous impl. There's a lot of stuff that can and should be refactored/removed.
I suggest to handle this in a separate PR, when the present proof protocol's state machines are refactored.

aries_vcx/src/protocols/proof_presentation/verifier/states/presentation_proposal_received.rs
bobozaur: Same as the other comment, I'd say it's outside of the PR.

aries_vcx/src/protocols/proof_presentation/verifier/states/presentation_request_set.rs
bobozaur: Same as the other comment, I'd say it's outside of the PR.

wrappers/node/test/suite1/ariesvcx-mediated-connection.test.ts
```diff
@@ -134,10 +134,12 @@ describe('Connection:', () => {
       assert.equal(await connection.getState(), ConnectionStateType.Finished);
     });
 
-    it('should not fail on attempt to handle unknown message type', async () => {
-      const connection = await createConnectionInviterFinished();
-      await connection.handleMessage(JSON.stringify(ARIES_UNKNOWN_TYPE));
-    });
+    // it should, actually.
```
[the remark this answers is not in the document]
bobozaur: Removed.

aries_vcx/src/utils/author_agreement.rs
bobozaur: Uncommented.

aries_vcx/Cargo.toml
```diff
@@ -15,31 +15,13 @@ default = ["vdrtools"]
...
-warnlog_fetched_messages = []
+
 
 [dependencies]
 agency_client = { path = "../agency_client" }
 messages = { path  = "../messages" }
-# default-features as false, since we transitively choose features via `vdrtools` and `modular_libs` features
-aries_vcx_core = { path  = "../aries_vcx_core", default-features = false }
+diddoc = { path = "../diddoc" }
+aries_vcx_core = { path  = "../aries_vcx_core" }
```
gmulhearn, line 24: this is a regression on a regression fix 😅.. removing `default-features = false` from the `aries_vcx_core` dependency forces all `aries_vcx` consumers to have a dependency on `vdrtools`, even when they have specified they don't want that dependency.

For some (including myself), this is a fatal regression, as `vdrtools` is full of old dependencies that can cause tremendous dependency conflicts (e.g. with aries_askar and other more modern crates).

Trying not to nitpick, but I consider this a fatal regression - was something about `default-features = false` causing you issues?? if not, would be good if you could add it back

bobozaur: It probably was if I removed it, but I'll look into it. Good catch

gmulhearn: thnx. Lets see how the CI reacts to that...

gmulhearn: Hmm. I believe a quick fix for the issue in the CI: [check-aries-vcx-feature-variants](https://github.com/hyperledger/aries-vcx/actions/runs/4806298131/jobs/8553648204#logs), is to add the `vdrtools` feature flag above the `test_utils` and `devsetup` modules:

```
#[cfg(feature = "vdrtools")]
pub mod devsetup;
```

```
#[cfg(feature = "vdrtools")]
pub mod test_utils;
```

aries_vcx/src/utils/devsetup.rs
```diff
@@ -365,20 +369,27 @@ impl SetupProfile {
         cfg!(feature = "modular_libs_tests")
```
gmulhearn, line 369: this function will always return false now since we aren't using this flag anymore. 

This function is used by `init_holder_setup_in_indy_context` to decide we should run the holder POV using vdrtools or modular_libs. So always returning false means that a lot of tests with modular_libs are no longer being ran in CI, which is dangerous IMO.

I think changing this to `"modular_libs"` instead should work..

aries_vcx/Cargo.toml
gmulhearn: @bobozaur  your recent fix may cause issues in libvcx, since it depends on test_utils and/or devsetup i believe... (which is not the best IMO)..

Adding `[cfg(feature = 'vdrtools')]` may fix the problem with the CI step you're trying to fix.. unsure if it will break tests tho..

aries_vcx/src/utils/devsetup.rs
bobozaur: Great catch again. It was a leftover. Fixed. Hopefully the CI will behave.

aries_vcx/Cargo.toml
gmulhearn: hahhahah yep, it's a bit tangled. I think removing the testing flags is a good first step though.

bobozaur: A good first step to hell

aries_vcx/src/utils/devsetup.rs
bobozaur: Fixed.

aries_vcx/Cargo.toml
bobozaur: Went at this differently, because of the hundreds of errors this generated.
Removed the `default` from `aries_vcx_core` instead. This means that neither `vdrtools` nor `modular_libs` are selected implicitly, so are controlled instead from `aries_vcx`.

The benefit is that it does not require changes in `aries_vcx` where every line change causes an avalanche of errors. We really need to put effort, a lot of it, into getting this into a better shape. 

I've had more fun getting punched in the face. Hopefully the CI will run okay.

gmulhearn: yup, that approach makes sense. 

It's close, but compiling aries_vcx with `--features modular_libs --no-default-features` and `--no-default-features` won't work bcus test_utils and devsetup assumes vdrtools is on. 

I'm thinking that the quick fix is to add the `[cfg(feature = 'vdrtools')]` for test_utils and devsetup

gmulhearn: it's very unfortunate that `tests/` and `libvcx` depend on functionality within `test_utils` and `devsetup` :(

bobozaur: There are things in `devsetup` that depend on `modular_libs`. We can't just add `vdrtools` to it...

gmulhearn: Not sure if that matters; currently tests must be ran with either; just the `vdrtools` flag (default), or both the `vdrtools` and `modular_libs` flag (default  + modular_libs). Never is it just `modular_libs`.

`devsetup` and `test_utils` are only used by tests, and those tests are only ran if atleast `vdrtools` is enabled.

It's a bit sketchy, but it's a quick fix to get you thru it I believe.

With this fix locally, I've been able to run a good amount of integration tests (both vdrtools and modular_libs), and all the `cargo check` variants pass (since these cargo checks are not checking any `cfg(test)` code)

bobozaur: Hm... Let me try.

bobozaur: It looks like it works. Let's see the CI. I don't understand why they're not exclusive and at this point I am too afraid to ask.

gmulhearn: The main reason is that a lot of `Profile` functionality (issuer and wallet) has not been implemented with `modular_libs` yet. So having pure 'modular_libs' end-to-end tests aren't possible, which means there's a bit of weird hybrid tests and all the end to end tests still depend on vdrtools a little bit... 

Once we reach pure `modular_libs` implementation, they can be exclusive, and a lot of the conditional mess can be cleaned up

bobozaur: All ran now!

Patrik-Stas (approved):

mirgee (approved):

gmulhearn (approved):
Mostly looked at aries_vcx and aries_vcx_core for my review. but lgtm! very epic PR

PR #789, opened 2023-04-03
Fix iOS wrapper, connectionHandle

pomfar:
Fix in the iOS wrapper, connectionHandle parameter as Number* instead of VcxHandle


Patrik-Stas (approved):

Patrik-Stas:
thank you @pomfar

PR #793, opened 2023-04-06
Add get_revocation_id for IssuerCredential

Patrik-Stas:


aries_vcx/src/handlers/issuance/issuer.rs
```diff
@@ -209,6 +209,20 @@ impl Issuer {
...
     }
 
+    pub fn get_revocation_id(&self) -> VcxResult<String> {
+        self.issuer_sm
+            .get_revocation_info()
+            .ok_or(AriesVcxError::from_msg(
+                AriesVcxErrorKind::InvalidState,
+                "Credential is has not yet been created",
+            ))?
+            .cred_rev_id
```
mirgee, line 219: Tangential to the PR, but `cred_rev_id` probably shouldn't be a string?

aries_vcx/src/handlers/issuance/issuer.rs
```diff
@@ -209,6 +209,20 @@ impl Issuer {
         self.issuer_sm.find_message_to_handle(messages)
     }
 
+    pub fn get_revocation_id(&self) -> VcxResult<String> {
+        self.issuer_sm
+            .get_revocation_info()
+            .ok_or(AriesVcxError::from_msg(
+                AriesVcxErrorKind::InvalidState,
+                "Credential is has not yet been created",
```
mirgee, line 217: Nitpick, but perhaps one of "Credential exchange protocol has not yet finished" / "Credential exchange protocol has not yet completed" / "Credential exchange protocol has not yet concluded" / "Credential exchange protocol incomplete" / "Revocation info not present" might be slightly more descriptive as an error message.

aries_vcx/src/handlers/issuance/issuer.rs
Patrik-Stas: yeah, but that issue roots from vdrtools

aries_vcx/src/handlers/issuance/issuer.rs
Patrik-Stas: I don't quite agree, as that would indicate this error will be returned whenever the protocol is not completed. But that's not true, because `CredentialState` is not final state and `ack` is expected, yet `get_revocation_info()` will return data.

Patrik-Stas: But this, along with other error handlings due to usage of `Option` in our data structures will go away with state pattern

Patrik-Stas: fixed minor typo in the message though

aries_vcx/src/handlers/issuance/issuer.rs
mirgee: OK, I apologize for the comment, then.

mirgee (approved):

PR #812, opened 2023-04-24
Add new crates for DID parsing, DID document building, and DID resolution

mirgee:
This PR introduces new crates related to DID parsing, DID document building, and DID resolution, thus partially addressing issue #706 .

Some features in the added crates have no immediate use inside of AriesVCX, and we may consider moving some of these crates to a separate project, especially as the number of implemented resolvers begins to grow. However, their inclusion in the PR is due to the immediate need for a Sovrin DID method resolver, while striving to keep in line with following design goals:

* suitability for use in AriesVCX, and
* reusability in other projects, such as a driver for [universal-resolver](https://github.com/decentralized-identity/universal-resolver),
* compliance with the relevant standards and specifications:
  * [DID Core specification](https://www.w3.org/TR/did-core),
  * [DID Resolution draft](https://w3c-ccg.github.io/did-resolution),
  * [Sovrin DID method specification](https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html).

For example DID resolver HTTP(S) binding implementation using these crates, see [here](https://github.com/AbsaOSS/driver-did-sov).

 The following crates have been added:

#### did_parser
This crate is responsible for parsing DID URLs and DIDs into easily queriable components, as well as DID (URL) validation. Includes `ParsedDID` and `ParsedDIDUrl` structs, which are used heavily throughout the remaining crates.

#### did_doc_builder
The `did_doc_builder` crate allows its users to construct, serialize and deserialize DID documents and their components that comply with the DID Core specification (https://www.w3.org/TR/did-core/).

#### did_resolver
This crate defines traits that should be implemented by DID resolvers of specific DID methods. For now, the traits are `DIDResolvable` and `DIDDereferenceable`.

#### did_resolver_registry
The `did_resolver_registry` is a simple crate that enables the registration of multiple DID resolvers for different DID methods at runtime and subsequent resolution of DIDs for methods with a registered resolver, thus enabling support for variable number of DID methods.

#### did_resolver_sov
This crate provides a DID resolver for the Sovrin DID method (https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html), and hence is the main candidate for possible use in AriesVCX.

Please review and let me know if any further changes are required.


nicobao:
Really nice!
I wonder if there is any work planned to support non-ledger methods such as `did:web`?
The `did-web` rust crate written by Spruce could be reused.

Patrik-Stas:
@nicobao most definitely we'll be adding support for more methods. Thanks for heads up about the spruce repo - I wasn't aware of it and it actually looks amazing 👍 💯  - so many did methods ready to hook tap into at the fingertips https://github.com/spruceid/ssi

The main concern imo would be all the dependencies and the code it would bring it, many of which would likely be unnecessary. I think we could help spruceid to feature flag the dependencies, so it would be possible to bring in only the deps for desired set of did methods.

@mirgee @bobozaur @gmulhearn check it out guys :-)

did_parser/src/utils/parse.rs
```diff
@@ -0,0 +1,128 @@
...
+    // assumed: method-specific-id = 1*idchar
+    // actual : method-specific-id = *( *idchar ":" ) 1*idchar
+    let id_start = method_end + 1;
+    let id_end = did_url[id_start..]
+        .find(|c: char| c == ';' || c == '/' || c == '?' || c == '#' || c == '&')
+        .map_or(did_url.len(), |i| i + id_start);
+
+    let did = 0..id_end;
+    let method = method_start + 1..method_end;
+    let id = id_start..id_end;
```
Patrik-Stas, line 56: we should verify `id` doesn't contain disallowed characters

did_parser/src/utils/parse.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    Ok((key_start, value_start, next_pos))
+}
+
+// TODO: Support tunnel methods
+pub fn parse_did_method_id(did_url: &str) -> Result<(DIDRange, DIDRange, DIDRange), ParseError> {
+    // DID = "did:" method ":" method-specific-id
+    let method_start = did_url
+        .find(':')
+        .ok_or(ParseError::InvalidInput(did_url.to_string()))?;
```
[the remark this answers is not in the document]
Patrik-Stas: but now I see why you probably didn't do it - you reuse `parse_did_method_id` in `ParsedDIDUrl::parse` which totally makes sense, but in that context you expect further characters to come (which are not valid DID characters) hence the weak constrains here.

Suggested way out of it:
- let `parse_did_method_id` return `(did, method, id, rest)`, eg adding the `rest` which would be string containing whatever followed after first invalid `method-specific-id` character (inclusive). Then in `ParsedDIDUrl::parse` you could proceed to attempt parse the `rest` as the subsequent components of DID URL (eg. `path-abempty [ "?" query ] [ "#" fragment ]`
 )

aries_vcx_core/src/ledger/base_ledger.rs
```diff
@@ -4,6 +4,7 @@ use crate::errors::error::VcxCoreResult;
 
 /// Trait defining standard 'ledger' related functionality.
 #[async_trait]
+#[cfg_attr(any(test, feature = "mock"), mockall::automock)]
```
Patrik-Stas, line 7: automock, sweet :-) 
@gmulhearn

aries_vcx_core/Cargo.toml
```diff
@@ -12,6 +12,7 @@ modular_libs = ["dep:indy-vdr", "dep:indy-credx"]
 ########################## TEST FLAGS ###############################
 # Feature flag for base requirments for running tests underneath the `test_utils` flag (most tests).
 # Both `vdrtools` and `modular_libs` are transitively enabled, so they can be used for tests (including interop tests).
+mock = ["dep:mockall"]
```
Patrik-Stas, line 15: I'd suggest different feature name to better distinguish it / make it easier to look up. Also short description what code does the future enables would be nice

.github/workflows/main.yml
```diff
@@ -432,6 +432,19 @@ jobs:
         run: |
           RUST_TEST_THREADS=1 cargo test --manifest-path="libvcx/Cargo.toml" -F "pool_tests";
 
+  test-integration-resolver:
```
Patrik-Stas, line 435: Following up the testing re-approach, integration tests are now typically have `ignore` annotation and we typically run unit and integration tests separately. While the resolves resides under aries-vcx, we should probably stay consistent

Patrik-Stas, line 435: The name is not quite accurate right now, the command also currently run unit tests too. Either update now, or tweak the cargo cmd

.github/workflows/main.yml
```diff
@@ -395,7 +395,7 @@ jobs:
         with:
           skip-docker-setup: true
       - name: "Run workspace tests: general_test"
-        run: RUST_TEST_THREADS=1 cargo test --workspace --exclude aries-vcx-agent --exclude libvdrtools
+        run: RUST_TEST_THREADS=1 cargo test --workspace --lib --exclude aries-vcx-agent --exclude libvdrtools
```
Patrik-Stas, line 398: good catch

did_parser/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "did_parser"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+itertools = "0.10.5"
+once_cell = "1.17.1"
```
Patrik-Stas, line 10: also unused

did_parser/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "did_parser"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+itertools = "0.10.5"
```
Patrik-Stas, line 9: seems unused

did_parser/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+once_cell = "1.17.1"
+ouroboros = "0.15.6"
+regex = "1.7.3"
```
Patrik-Stas, line 12: :-)

mirgee: Good catch!

did_parser/Cargo.toml
mirgee: Good catch!

did_parser/Cargo.toml
mirgee: Good catch!

did_parser/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+name = "did_parser"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+itertools = "0.10.5"
+once_cell = "1.17.1"
+ouroboros = "0.15.6"
```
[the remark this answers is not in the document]
mirgee: Good catch!

did_parser/src/utils/parse.rs
mirgee: Good catch.

Note that this is partially responsibility of the DID method implementor, as DID methods may impose additional restrictions on `method-specific-id`. However, there are some restrictions that apply already on the level of DID Core specification.

did_resolver/src/traits/resolvable/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+
+#[async_trait]
+// #[cfg_attr(any(test, feature = "mock"), mockall::automock)]
+pub trait DIDResolvableMut: DIDResolvable {
+    async fn resolve_mut(
+        &mut self,
+        did: &ParsedDID,
+        options: &DIDResolutionOptions,
+    ) -> Result<DIDResolutionOutput, GenericError>;
+}
```
bobozaur, line 30: Why a mut variant?

did_resolver/src/traits/dereferenceable/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+// #[cfg_attr(test, mockall::automock)]
+pub trait DIDDereferenceableMut: DIDDereferenceable {
+    type Output: Read + Send + Sync;
+
+    async fn dereference_mut(
+        &mut self,
+        did: &ParsedDIDUrl,
+        options: &DIDDereferencingOptions,
+    ) -> Result<DIDDereferencingOutput<<Self as DIDDereferenceableMut>::Output>, GenericError>;
+}
```
bobozaur, line 39: What is this trait for and why does it have a mut variant?

bobozaur (changes requested):
I like the DID parser using ranges of the input DID string, great job!

To me it feels like a lot of places would need documentation and comments, like the structs and traits and the larger functions that are not really easy to follow. I'm honestly not able to grasp everything that's happening just from browsing through the code so I tried to start conversations where appropriate.

There are some minor things to adjust as far as idioms go, but nothing terrible. Another broader idea would be to incorporate some of the utility functions as associated methods, because it's easier to grasp what they are about if they belong to a namespace and most seem to have unique use cases.

Great job over all!

did_doc_builder/src/schema/service.rs
```diff
@@ -0,0 +1,180 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Service {
+    id: Uri,
+    r#type: ServiceTypeAlias,
+    service_endpoint: String,
+    #[serde(flatten)]
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    #[serde(default)]
+    extra: HashMap<String, Value>,
```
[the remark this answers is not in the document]
Patrik-Stas: I can answer this one

https://www.w3.org/TR/did-core/#services

> Each service extension MAY include additional properties and MAY further restrict the properties associated with the extension.

and so for example here
https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html#crud-operation-definitions
under the `Resolver DID Document Format`
there's DID Doc example which contains `routingKeys` and `recipientKeys` which are not part of w3c spec.

We are currently using, and will be using for a while more `did:sov` method.

did_doc_builder/src/schema/types/multibase.rs
```diff
@@ -0,0 +1,123 @@
...
+        self.base.encode(&self.bytes)
+    }
+}
+
+impl Deref for Multibase {
+    type Target = Vec<u8>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.bytes
+    }
```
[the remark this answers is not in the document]
mirgee: Good catch!

did_doc_builder/src/schema/verification_method.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    public_key_multibase: Option<Multibase>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    public_key_jwk: Option<JsonWebKey>,
+    #[serde(flatten)]
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    #[serde(default)]
+    extra: HashMap<String, Value>,
```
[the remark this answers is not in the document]
mirgee: > ... A [verification method](https://www.w3.org/TR/did-core/#dfn-verification-method) MAY include additional properties. ...
https://www.w3.org/TR/did-core/#verification-methods

did_resolver/src/traits/resolvable/mod.rs
mirgee: Removed as per discussion.

did_doc_builder/src/schema/did_doc.rs
```diff
@@ -0,0 +1,349 @@
...
+        self.capability_delegation.as_ref()
+    }
+
+    pub fn service(&self) -> &[Service] {
+        self.service.as_ref()
+    }
+
+    pub fn extra(&self, key: &str) -> Option<&Value> {
+        self.extra.get(key)
+    }
```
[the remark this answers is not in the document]
mirgee: Left in as is as discussed.

did_doc_builder/src/schema/verification_method.rs
```diff
@@ -0,0 +1,236 @@
...
+use crate::error::DIDDocumentBuilderError;
+
+use super::types::{jsonwebkey::JsonWebKey, multibase::Multibase};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(untagged)]
+pub enum VerificationMethodAlias {
+    VerificationMethod(VerificationMethod),
+    VerificationMethodReference(ParsedDIDUrl),
+}
```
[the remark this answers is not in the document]
mirgee: Added.

did_parser/src/utils/parse.rs
```diff
@@ -0,0 +1,128 @@
...
+    if &did_url[..method_start] != "did" {
+        return Err(ParseError::InvalidInput(did_url.to_string()));
+    }
+    let method_end = did_url[method_start + 1..]
+        .find(':')
+        .map(|i| i + method_start + 1)
+        .ok_or(ParseError::InvalidInput(did_url.to_string()))?;
```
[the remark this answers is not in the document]
mirgee: Improved through error-specific messages.

aries_vcx_core/Cargo.toml
mirgee: Removed.

did_doc_builder/src/schema/types/uri.rs
```diff
@@ -0,0 +1,88 @@
+use std::{ops::Deref, str::FromStr};
+
+use serde::{Deserialize, Serialize};
+
+use crate::error::DIDDocumentBuilderError;
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+pub struct Uri(uniresid::Uri);
```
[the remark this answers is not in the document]
mirgee: We decided to keep it as per discussion.

did_doc_builder/src/schema/verification_method.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub fn add_extra(mut self, key: String, value: Value) -> Self {
+        self.extra.insert(key, value);
+        self
+    }
+
+    pub fn build(self) -> Result<VerificationMethod, DIDDocumentBuilderError> {
+        if self.public_key_multibase.is_some() && self.public_key_jwk.is_some() {
+            return Err(DIDDocumentBuilderError::InvalidInput(
+                "Cannot specify both public_key_multibase and public_key_jwk".to_string(),
+            ));
```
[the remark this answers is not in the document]
mirgee: Done.

did_doc_builder/src/schema/did_doc.rs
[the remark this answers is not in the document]
bobozaur: How about renaming it then? The method looks like a getter of the map field instead of a lookup into the map. I agree though that the inner field should not be exposed.

did_doc_builder/src/schema/service.rs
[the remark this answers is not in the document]
bobozaur: As discussed, we might be able to substitute this with some generic type with well defined fields, but it might not be worth the investment right now since we don't even know whether we can pull it off for all did methods. I understand what this is about now, so thanks 😄

did_doc_builder/src/schema/types/multibase.rs
```diff
@@ -0,0 +1,123 @@
...
+use multibase::{decode, Base};
+use serde::{Deserialize, Serialize};
+
+use crate::error::DIDDocumentBuilderError;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Multibase {
+    base: Base,
+    bytes: Vec<u8>,
+}
```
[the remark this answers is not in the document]
bobozaur: Alright, I see. I still suggest to add a comment, even including that link, for people that might not be familiar with this. Having more details can't be that bad. Having less usually ends up costing way more time.

did_doc_builder/src/schema/verification_method.rs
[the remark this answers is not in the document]
bobozaur: I'm still a bit confused about this, so I'm just unresolving for visibility so we can discuss tomorrow.

did_doc_builder/src/schema/verification_method.rs
```diff
@@ -0,0 +1,236 @@
...
+        self.public_key_multibase.as_ref()
+    }
+
+    pub fn public_key_jwk(&self) -> Option<&JsonWebKey> {
+        self.public_key_jwk.as_ref()
+    }
+
+    pub fn extra(&self, key: &str) -> Option<&Value> {
+        self.extra.get(key)
+    }
```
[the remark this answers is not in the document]
bobozaur: How about a rename of the method?

did_doc_builder/src/schema/verification_method.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+#[derive(Debug, Default)]
+pub struct VerificationMethodBuilder {
+    id: ParsedDIDUrl,
+    controller: ParsedDID,
+    r#type: String,
+    public_key_multibase: Option<Multibase>,
+    public_key_jwk: Option<JsonWebKey>,
+    extra: HashMap<String, Value>,
```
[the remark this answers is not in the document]
bobozaur: How about a rename of the method?

did_parser/src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+extern crate regex;
+
+mod error;
+mod parsed_did;
+mod parsed_did_url;
+mod utils;
+
+use std::ops::Range;
+
+type DIDRange = Range<usize>;
```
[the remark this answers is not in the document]
bobozaur: You mean if you decided to use something else instead of a `Range<usize>`? I don't really think it would be a problem. Yeah, it's easier to change it in one place alone, but on the other hand the compiler will just complain if you happen to forget a spot. And it will be easier to see what the actual type is.

But I think this is more of a personal preference than anything else, so feel free to disregard it if it's more convenient for you like that. I generally just alias if the type def is very long 😅

did_parser/src/parsed_did_url.rs
```diff
@@ -0,0 +1,182 @@
...
+pub struct ParsedDIDUrl {
+    did_url: String,
+    did: Option<DIDRange>,
+    method: Option<DIDRange>,
+    id: Option<DIDRange>,
+    path: Option<DIDRange>,
+    fragment: Option<DIDRange>,
+    queries: HashMap<DIDRange, DIDRange>,
+    params: HashMap<DIDRange, DIDRange>,
+}
```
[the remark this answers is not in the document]
bobozaur: Just for posterity, as we discussed these do indeed have to be independent structures due to their capabilities and usage.

did_resolver/src/traits/dereferenceable/dereferencing_error.rs
```diff
@@ -0,0 +1,21 @@
...
+pub enum DIDDereferencingError {
+    InvalidDid,
+    NotFound,
+}
+
+impl Display for DIDDereferencingError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            DIDDereferencingError::InvalidDid => write!(f, "invalidDid"),
+            DIDDereferencingError::NotFound => write!(f, "notFound"),
```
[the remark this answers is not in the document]
bobozaur: Yeah, fair. I see what you mean now.

did_resolver/src/traits/resolvable/resolution_error.rs
```diff
@@ -0,0 +1,61 @@
...
+            DIDResolutionError::NotAllowedCertificate => write!(f, "notAllowedCertificate"),
+            DIDResolutionError::NotAllowedLocalDuplicateKey => {
+                write!(f, "notAllowedLocalDuplicateKey")
+            }
+            DIDResolutionError::NotAllowedLocalDerivedKey => {
+                write!(f, "notAllowedLocalDerivedKey")
+            }
+            DIDResolutionError::NotAllowedGlobalDuplicateKey => {
+                write!(f, "notAllowedGlobalDuplicateKey")
+            }
```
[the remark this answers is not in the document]
bobozaur: Touche. My bad!

did_resolver_registry/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    traits::resolvable::{
+        resolution_options::DIDResolutionOptions, resolution_output::DIDResolutionOutput,
+        DIDResolvable,
+    },
+};
+use error::DIDResolverRegistryError;
+
+pub struct ResolverRegistry {
+    resolvers: HashMap<String, Box<dyn DIDResolvable>>,
+}
```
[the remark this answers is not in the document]
bobozaur: I'm just thinking that we could perhaps localize this to where it would be used in the beginning rather than have a completely different crate for it. It seems to me that the workspace is getting quite crowded, but it might just be me. @Patrik-Stas what do you think?

did_doc_builder/src/schema/did_doc.rs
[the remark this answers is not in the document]
mirgee: Apologies, forgot about the suggestion to rename. What name do you suggest?

did_doc_builder/src/schema/types/multibase.rs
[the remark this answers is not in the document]
mirgee: Added.

did_doc_builder/src/schema/verification_method.rs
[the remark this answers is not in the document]
mirgee: Let's just keep one thread about this open.

did_parser/src/lib.rs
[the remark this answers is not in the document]
mirgee: Hm. I really do think type alias makes sense in this case. It is a matter of less than second to check the type in any modern IDE.

did_doc_builder/src/schema/did_doc.rs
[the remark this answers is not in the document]
bobozaur: Something like `lookup_extra_field`?

did_doc_builder/src/schema/verification_method.rs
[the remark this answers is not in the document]
mirgee: Let's keep just one thread open about this.

did_doc_builder/tests/serde.rs
```diff
@@ -0,0 +1,208 @@
...
+  "keyAgreement": [
+    {
+      "id": "did:web:did.actor:alice#zC8GybikEfyNaausDA4mkT4egP7SNLx2T1d1kujLQbcP6h",
+      "type": "X25519KeyAgreementKey2019",
+      "controller": "did:web:did.actor:alice",
+      "publicKeyBase58": "CaSHXEvLKS6SfN9aBfkVGBpp15jSnaHazqHgLHp8KZ3Y"
+    }
+  ]
+}
+"##;
```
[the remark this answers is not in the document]
bobozaur: Eh, my bad. I didn't look at *where* the file was 😅 .

did_parser/src/parsed_did.rs
```diff
@@ -0,0 +1,87 @@
...
+pub struct ParsedDID {
+    did: String,
+    method: DIDRange,
+    id: DIDRange,
+}
+
+impl ParsedDID {
+    pub fn parse(did: String) -> Result<Self, ParseError> {
+        if is_valid_did(&did) {
+            let (_, method, id) = parse_did_method_id(&did)?;
```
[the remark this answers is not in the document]
bobozaur: Fair enough.

did_doc_builder/src/schema/types/jsonwebkey.rs
```diff
@@ -0,0 +1,43 @@
...
+impl JsonWebKey {
+    pub fn new(jwk: String) -> Result<Self, DIDDocumentBuilderError> {
+        if is_valid_jwk(&jwk) {
+            Ok(Self(Value::from_str(&jwk)?))
+        } else {
+            Err(DIDDocumentBuilderError::InvalidInput(format!(
+                "Invalid JWK: {}",
+                jwk
+            )))
+        }
```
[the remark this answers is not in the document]
bobozaur: I think there could at least be a type that contains the required fields which you are validating later and, if it is so difficult to determine all the fields and their type, have a flattened map for the rest of the fields. This way the validation is done during deserialization and wouldn't require the `Value` validation.

did_resolver/src/traits/dereferenceable/mod.rs
bobozaur: Alright, thanks. I'm good since the mut variant was removed, but I don't think some comments would be bad here.

did_doc_builder/src/schema/service.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+impl ServiceBuilder {
+    pub fn new(id: Uri, service_endpoint: String) -> Result<Self, DIDDocumentBuilderError> {
+        if id.is_empty() {
+            return Err(DIDDocumentBuilderError::MissingField("id"));
+        }
+        if service_endpoint.is_empty() {
+            return Err(DIDDocumentBuilderError::MissingField("serviceEndpoint"));
+        }
```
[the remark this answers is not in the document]
bobozaur: Alright, the error string makes sense then, but I still think that checking for the empty string is not really of any importance here.

The idea is that if you had something like `fn gimme_any_non_empty_string(my_str: String) {}` then I'd get why you would validate that. However, the validation is more complex here, and simply because a String is non-empty does not mean it's a valid endpoint. Trying to parse it to an `Url`, though, would validate all cases, and this check becomes redundant.

bobozaur (changes requested):
Addressed the discussions on some of the updates/pending points.

did_doc_builder/src/schema/did_doc.rs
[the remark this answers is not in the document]
mirgee: How about `get_extra_field` instead of `lookup_extra_field`?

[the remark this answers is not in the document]
bobozaur: I think get is implicit. Maybe just `extra_field`?

did_resolver_registry/src/lib.rs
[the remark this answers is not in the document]
bobozaur: Let's leave it here as discussed.

did_resolver_sov/src/resolution/utils.rs
```diff
@@ -0,0 +1,181 @@
...
+    let endpoint: EndpointDidSov = serde_json::from_value(service_data["endpoint"].clone())?;
+
+    let txn_time = get_txn_time_from_response(resp)?;
+    let datetime = posix_to_datetime(txn_time);
+
+    let service = {
+        let mut service_builder =
+            Service::builder(service_id, endpoint.endpoint.as_str().try_into()?)?;
+        for t in endpoint.types {
+            if t != DidSovServiceType::Unknown {
```
Patrik-Stas, line 77: This is just a nice to have, and I toched on this elsewhere. Feel free to pass this, I think this could be nice good-first-issue.

If you leave this up for good-first-issue, please leave this comment un-resolved.

Although DID Core does not specify possible values of `type`, we could still make our code slightly developer friendlier making enum for types within our DidDocument structure. We could reuse `DidSovServiceType` - rename it  to`ServiceType`, move it to diddocument crate and tweak it a bit:
```
#[derive(Debug, Clone, Deserialize, PartialEq, Eq, Hash)]
pub enum ServiceType {
    #[serde(rename = "endpoint")] // AIP 1.0
    Endpoint,
    #[serde(rename = "did-communication")] // AIP 2.0
    DidCommunication,
    #[serde(rename = "DIDComm")] // DIDComm V2
    DIDComm,
    #[serde(other)]
    Unknown(value),
}
```
Just not sure about details how to get `Unknown(value)` such that the value parameter would get filled in.

did_resolver_sov/src/resolution/utils.rs
```diff
@@ -0,0 +1,181 @@
...
+    for c in id.chars() {
+        match c {
+            '1'..='9' | 'A'..='H' | 'J'..='N' | 'P'..='Z' | 'a'..='k' | 'm'..='z' => {}
+            _ => return false,
+        }
+    }
+    true
+}
+
+pub(super) async fn resolve_ddo(did: &str, resp: &str) -> Result<DIDResolutionOutput, DIDSovError> {
```
Patrik-Stas, line 64: feel like `map_ledger_response_to_ddo` would be more appropriate, as for function doing resolution I would expect it to read the ledger as well, but that's injected

did_resolver_sov/src/resolution/utils.rs
```diff
@@ -0,0 +1,181 @@
...
+            Service::builder(service_id, endpoint.endpoint.as_str().try_into()?)?;
+        for t in endpoint.types {
+            if t != DidSovServiceType::Unknown {
+                service_builder = service_builder.add_service_type(t.to_string())?;
+            };
+        }
+        service_builder.build()?
+    };
+
+    let ddo = DIDDocument::builder(ddo_id).add_service(service).build();
```
Patrik-Stas, line 84: Unless I am missing something, the DDO produced by this function is incomplete. Verification method is not set, and `resp` argument passed in argument is ledger response for getting attribute, but you didn't fetch DID's verkey anywhere

Patrik-Stas: It also seems that if ledger contains routing_keys or recipient keys, it wouldn't get set up in service object's extras

Patrik-Stas (changes requested):
left some comments, still continuing in review

did_doc_builder/src/schema/service.rs
```diff
@@ -0,0 +1,174 @@
...
+            id,
+            service_endpoint,
+            service_type: HashSet::new(),
+            extra: HashMap::new(),
+        })
+    }
+
+    pub fn add_service_type(
+        mut self,
+        service_type: String,
```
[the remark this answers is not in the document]
mirgee: [Here](https://www.w3.org/TR/did-spec-registries/#service-types) is a non-exhaustive list of the service types just for reference.

did_doc_builder/src/schema/types/jsonwebkey.rs
```diff
@@ -0,0 +1,44 @@
...
+    str::FromStr,
+};
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+use crate::error::DIDDocumentBuilderError;
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+// TODO: Introduce proper custom type
```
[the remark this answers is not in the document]
mirgee: https://www.rfc-editor.org/rfc/rfc7517
https://www.iana.org/assignments/jose/jose.xhtml

did_resolver_sov/src/service.rs
```diff
@@ -0,0 +1,146 @@
+use std::collections::HashSet;
+use std::fmt::Display;
+
+use serde::{Deserialize, Deserializer};
+
+#[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct EndpointDidSov {
+    pub endpoint: String,
```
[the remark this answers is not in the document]
mirgee: None other than that it introduces extra dependency. I will change it.

did_doc_builder/src/schema/verification_method.rs
```diff
@@ -0,0 +1,253 @@
...
+use serde_json::Value;
+
+use super::types::{jsonwebkey::JsonWebKey, multibase::Multibase};
+
+// Either a set of verification methods maps or DID URLs
+// https://www.w3.org/TR/did-core/#did-document-properties
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(untagged)]
+pub enum VerificationMethodKind {
+    VerificationMethodResolved(VerificationMethod),
```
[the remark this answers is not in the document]
mirgee: Done.

did_resolver_sov/src/resolution/utils.rs
```diff
@@ -0,0 +1,181 @@
...
+    let resp: serde_json::Value = serde_json::from_str(resp)?;
+    let txn_time = resp["result"]["txnTime"]
+        .as_i64()
+        .ok_or(DIDSovError::ParsingError(
+            ParsingErrorSource::LedgerResponseParsingError("Failed to parse txnTime".to_string()),
+        ))?;
+    Ok(txn_time)
+}
+
+fn posix_to_datetime(posix_timestamp: i64) -> Option<DateTime<Utc>> {
```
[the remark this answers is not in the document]
mirgee: Changed.

did_resolver_sov/src/resolution/utils.rs
mirgee: Good point, renamed to `ledger_response_to_ddo`.

did_doc_builder/src/schema/service.rs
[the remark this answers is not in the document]
mirgee: Added a serde test with did:sov extra fields.

did_parser/src/parsed_did_url.rs
```diff
@@ -0,0 +1,182 @@
...
+
+impl ParsedDIDUrl {
```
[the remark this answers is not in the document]
mirgee: Renamed `DID` to `Did` across the board.

Patrik-Stas (approved):
lgtm, can we move it from draft to ready state?

Patrik-Stas (dismissed):
lgtm, can we move it from draft to ready state?

did_doc_builder/src/schema/did_doc.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    pub fn add_service(mut self, service: Service) -> Self {
+        self.service.push(service);
+        self
+    }
+
+    pub fn add_extra_field(mut self, key: String, value: Value) -> Self {
+        self.extra.insert(key, value);
+        self
+    }
```
bobozaur, line 200: It's common for builder setter methods to just have re-use the field's name (as setting fields is pretty much all a builder does).

did_doc_builder/src/schema/types/uri.rs
```diff
@@ -0,0 +1,89 @@
...
+use crate::error::DidDocumentBuilderError;
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+pub struct Uri(uniresid::Uri);
+
+impl Uri {
+    pub fn new(uri: &str) -> Result<Self, DidDocumentBuilderError> {
+        Ok(Self(uniresid::Uri::try_from(uri).map_err(|e| {
+            DidDocumentBuilderError::InvalidInput(format!("Invalid URI: {}", e))
+        })?))
```
bobozaur, line 17: I think it would be cleaner and more readable to separate the `uniresid::Uri` construction into a variable.

did_parser/src/parsed_did.rs
```diff
@@ -0,0 +1,84 @@
...
+use std::fmt::{Display, Formatter};
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::DidUrl;
+use crate::{error::ParseError, utils::parse::parse_did_method_id, DidRange};
+
+#[derive(Default, Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Did {
+    did: String,
```
bobozaur, line 11: Maybe we could rename the `did` field to `full_did` or something like that? Might make things more intuitive as why there's a `did` field in the `Did` struct.

did_parser/src/parsed_did.rs
```diff
@@ -0,0 +1,84 @@
...
+impl TryFrom<&DidUrl> for Did {
+    type Error = ParseError;
+
+    fn try_from(did_url: &DidUrl) -> Result<Self, Self::Error> {
+        Self::parse(
+            did_url
+                .did()
+                .ok_or(Self::Error::InvalidInput("No DID provided in the DID URL"))?
+                .to_string(),
+        )
```
bobozaur, line 82: Shouldn't there be a better way to convert? The DID is already parsed, so it's just a matter of whether there are path arguments and stuff like that. If the `DidUrl` doesn't contain any of the additional stuff that makes it a did url, then the parts could be converted to a `Did`, right?

did_resolver_sov/src/resolution/utils.rs
bobozaur: This would require a custom impl for Serialize/Deserialize (or re-using the `MaybeKnown` enum?). The `#[serde(other)]` attribute only works with unit variants, unfortunately.

did_parser/src/parsed_did_url.rs
```diff
@@ -0,0 +1,182 @@
...
+
+use crate::{
+    error::ParseError,
+    utils::parse::{parse_did_method_id, parse_key_value, parse_path},
+    DidRange,
+};
+
+#[derive(Default, Debug, Clone, PartialEq)]
+pub struct DidUrl {
+    did_url: String,
```
bobozaur, line 13: Same thing, maybe it's worth naming this `full_did_url`?

bobozaur (changes requested):
Most of the things left are minor things or personal opinions, so feel free to merely comment or ignore them, but I think the conversion between `DidUrl` and `Did` can really be optimized to avoid double-parsing.

did_parser/src/parsed_did.rs
mirgee: Are you suggesting to expose the fields of `DidUrl` for this?

did_parser/src/parsed_did.rs
mirgee: I think it is quite obvious from the types.

did_parser/src/parsed_did.rs
bobozaur: Through `pub(crate)`, yeah. Why not? Or simply a constructor method could be added that's crate internal, and the from impl could be near the `ParsedDidUrl` type so the fields can stay private.

did_parser/src/parsed_did.rs
bobozaur: At least a comment then :D?

did_parser/src/parsed_did.rs
mirgee: I think the overhead caused by so negligible it's not worth additional code or exposing the now private fields at all. But I will do as you wish.

did_parser/src/parsed_did.rs
mirgee: I think adding a comment because of this is even worse, but will do as you wish.

did_doc_builder/src/schema/did_doc.rs
mirgee: I think the `add_` prefix is more descriptive. Do you want me to remove it everywhere?

did_parser/src/parsed_did.rs
bobozaur: The crate public constructor is probably the best way to go about it, so the fields aren't touched.

I think it's highly worth it. Why not optimize this when it's easy to do so? The difference between parsing and not parsing the string is huge in the number of executed instructions.

did_parser/src/parsed_did.rs
bobozaur: Why worse? A comment above the struct or just above the field would help everyone else who's looking at the code understand what's going on faster. 

For you and me it might be obvious because we discussed about this for some time, but newcomers would be saved from having to look at some other code to know exactly what the field is for. 

Although this is just my own opinion. @Patrik-Stas what do you think?

did_doc_builder/src/schema/did_doc.rs
bobozaur: I'd personally remove it everywhere, yeah. There are places where it's not already present, so this would make things consistent.

did_parser/src/parsed_did.rs
mirgee: Again, I don't think this is worth either changing the visibility of the fields or adding additional code. Keeping a struct's trait implementations in the same module it's defined in also makes the code navigable, orderly and well-organized. 

And again, a call to `parse_did_method_id` has absolutely negligible performance cost.

But since you think it's highly worth it, I will change it according to your preference.

did_doc_builder/src/schema/did_doc.rs
mirgee: Thank you for your feedback, but again, I think the prefix makes the purpose of the method clear and explicit.

did_parser/src/parsed_did.rs
mirgee: IMO comment is not only highly unnecessary in this case as the purpose of the field is obvious from context but almost detrimental as one has read the code and the superfluous comment as well.

bobozaur: As you wish, it's not a big deal.

did_parser/src/parsed_did_url.rs
```diff
@@ -0,0 +1,200 @@
...
+                method.to_owned(),
+                id.to_owned(),
+            ))
+        } else {
+            Err(ParseError::InvalidInput(
+                "Unable to construct a DID from relative DID URL",
+            ))
+        }
+    }
+}
```
bobozaur, line 200: Why from a `&DidUrl` and not a `DidUrl`?

bobozaur (changes requested):
Just a small comment regarding the TryFrom impl between `DidUrl` and `Did`

did_parser/src/parsed_did_url.rs
mirgee: Is there a need to own it for the conversion? It's used here:

```
    async fn dereference(
        &self,
        did_url: &DidUrl,
        _options: &DidDereferencingOptions,
    ) -> Result<DidDereferencingOutput<Self::Output>, GenericError> {
        let resolution_output = self
            .resolve(&did_url.try_into()?, &DidResolutionOptions::default())
            .await?;

        dereference_did_document(&resolution_output, did_url).map_err(|err| err.into())
    }
```

bobozaur (commented):
Can we also change the `parsed_did` and `parsed_did_url` module file names, for consistency?

did_parser/src/parsed_did_url.rs
bobozaur: It's more idiomatic. This limits the consumer API to only convert from a reference. This method can easily clone just as the consumer code can. The compiler is surely smart enough not to clone the other parts of `DidUrl` that are not used here in release mode.

bobozaur: Maybe we're looking at this the wrong way. What if the `Did` and `DidUrl` types would implement a common trait exposing `did`, `method` and `id`? Then the `resolve` method could take a generic implementing the trait and the conversion wouldn't even be necessary. 

What do you think?

mirgee: Implementing only owned version limits the consumer API as it must own `DidUrl`. There is no need to clone here as well as there is no need to rely on compiler optimizations.

Patrik-Stas (approved):

Patrik-Stas:
@mirgee please rerun the CI without `skip-ci` tag

did_parser/src/parsed_did_url.rs
bobozaur: Okay, this is your PR, you call the shots. I'm at least glad the `TryFrom` impl has changed. While I don't understand `asm`, parsing the `String` outputs 152 lines of code while the new impl outputs 37 lines. I find that to be huge.

bobozaur (approved):
If the CI runs, I'm okay with the PR.

Patrik-Stas (approved):

Patrik-Stas (approved):

PR #816, opened 2023-04-28
Fix README for deprecated link

swaptr:
`libvcx` is deprecated, the docs need to reflect that. The link in the docs to the deprecation notice seems to be broken. This PR fixes that.


Patrik-Stas (approved):

Patrik-Stas:
Good catch, thank you

PR #818, opened 2023-05-01
remove all feature flags

### LINKED ISSUE #799: Stop using feature flags for testing
Per @bobozaur's comment
https://github.com/hyperledger/aries-vcx/pull/797#pullrequestreview-1379809296

> Same thing I mentioned on one of @gmulhearn 's PRs - I think we should aim to remove these feature flags. Not now, as it would be too big a change, but down the line if we all agree on it.
> 
> The way I see it, we should have unit tests which should be fairly self-contained (and can use dev-dependencies) and integration tests which would be more extensive and maybe even require some mocks or other valid services altogether (and can also use dev-dependencies).
> 
> The mocks could then be placed into one/more common integration tests file/s. We could annotate some of the heavier tests with ignore and make use of naming patterns to run certain suites of tests (e.g all tests that start with test_pool_agency).
>
>The rationale is that these feature flags get in the way of code being compiled in IDEs so you don't really see that some change you made messes up some tests (at least not until you run them) and you might end up having made many changes which you then have to accommodate all tests to at once, rather than incrementally.
>
>The feature flags also kind of 'pollute' Cargo.toml, if you ask me.

swaptr:
This PR aims to implement changes suggested here:
https://github.com/hyperledger/aries-vcx/pull/797#pullrequestreview-1379809296.
Fixes #799

### DIFF
--- a/libvcx/Cargo.toml
+++ b/libvcx/Cargo.toml
@@ -11,12 +11,6 @@ path = "src/lib.rs"
 crate-type = ["staticlib", "cdylib", "lib"]
 doctest = false
 
-[features]
-test_utils = [ "libvcx_core/test_utils" ]
-pool_tests = [ "test_utils" ]
-general_test = [ "test_utils" ]
-fatal_warnings = []
-
 [dependencies]
 num-traits = "0.2.0"
 once_cell = { version = "1.15" }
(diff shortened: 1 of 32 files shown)


Patrik-Stas:
hi @swaptr the libvcx testing and clippy CI jobs are failing. Please have a look. Also, I feel like it will actually be a bit more effort than this, but you'll find out as you address those CI jobs.

Patrik-Stas:
Closing per inactivity, if you'd like to continue, feel free to reopen PR

PR #831, opened 2023-05-07
this PR fixes #820 (Credential and Presentation attributes mime-type adjustments)

### LINKED ISSUE #820: Credential and Presentation attributes mime-type adjustments
The attributes (`CredentialAttr` and `Attribute`) could benefit from handling the `value` and `mime-type` pair in a different manner. Right now both fields are deserialized and handling would have to be done based on whether a `MimeType` is present or not.

A better idea would be to use an `AttrValue` `#[serde(flatten)]` enum with two variants: 
- `Plain { value: String }`
- `Encoded { value: String, mime_type: MaybeKnown<MimeType> }`

Note the `MaybeKnown` enum use, which would also be a beneficial addition in case we get a mime_type we don't recognize.

Having the attribute value deserialized this way would make handling it easier since we can match and deduce that:
- if it's `Plain`, it's just a string,
- if it's `Encoded`, it's a base64 encoded string that then has to be processed according to the `MimeType`.

Furthermore, it would be best to rename `Attribute` to `PresentationAttr` to avoid confusion.

Steps:
    - [ ] Rename `Attribute` to `PresentationAttr`
    - [ ] Wrap `MimeType` in `MaybeKnown`
    - [ ] Create and use the `AttrValue` enum in both attribute types.

SumantxD:
as instructed AttrValue enum was created with two variants Plain and Encoded, the plain having { value: String } and Encoded with { value: String, mime_type: MaybeKnown<MimeType> }, which will facilitate in easy deserialization,

the AttrValue has been used in both the attributes (CredentialAttr and PresentationAttr), and the unit tests have been adjusted accordingly,

hope that this PR will solves the issue, any changes or feedback is welcomed

Thank You,


messages/src/msg_fields/protocols/cred_issuance/mod.rs
```diff
@@ -142,22 +142,31 @@ impl Serialize for CredentialPreviewMsgType {
...
+    Plain {
+        value: String,
+    },
+    Encoded {
+        value: String,
+        #[serde(rename = "mime-type")]
+        mime_type: MaybeKnown<MimeType>,
+    },
+}
+
```
bobozaur, line 157: While not important, I'm honestly not a big fan of enum variants with fields. I feel like a struct can provide better functionality at the cost of being slightly more verbose.

Also, the deserialization would have to be untagged. With that in mind, the `Encoded` variant would have to be attempted first, as the `Plain` variant should always get successfully deserialized because there's always going to be a `value` field (assuming the input is right).

messages/src/msg_fields/protocols/present_proof/propose.rs
```diff
@@ -98,27 +99,37 @@ impl Serialize for PresentationPreviewMsgType {
...
+    Plain {
+        #[serde(skip_serializing_if = "Option::is_none")]
+        value: Option<String>,
+    },
+    Encoded {
+        value: String,
+        #[serde(rename = "mime-type")]
+        mime_type: MaybeKnown<MimeType>,
+    },
+}
```
bobozaur, line 114: This should **not** be duplicated, but rather both attributes should reuse the same `AttrValue` enum. It should be placed somewhere where both protocols have access to. Maybe create a `common` module in `msg_fields` module with an `attr_value` module inside?

bobozaur (changes requested):
Not quite there yet, but not far off either. With some tweaking this will look like what we need. Good job so far!

SumantxD:
added the required modifications

messages/src/msg_fields/mod.rs
```diff
@@ -3,3 +3,27 @@
...
+        }
+        
+        #[derive(Debug, Serialize, Deserialize)]
+        pub struct EncodedAttrValue {
+            value: String,
+            #[serde(rename = "mime-type")]
+            mime_type: Option<MaybeKnown<MimeType>>,
+        }
+    }
+}
```
bobozaur, line 29: This would sit better as a separate file (including the attr_value module).

bobozaur (changes requested):
Looks good, but I'd love to see the modules separated in different files for readability.

SumantxD:
added those changes

bobozaur:
> added those changes

@SumantxD What I meant was for each of those to be individual modules, so the directory structure would look more like this:
- msg_fields
    - common
        - attr_value.rs
        - mod.rs
    - protocols
    - mod.rs

SumantxD:
@bobozaur I have fixed the directory structure

SumantxD:
@bobozaur a lot of checks were unsuccessful in the CI run, will it create problems? if so how can i fix it?

Patrik-Stas:
Closing per inactivity. Feel free to reopen if you would like to continue in this work.

PR #833, opened 2023-05-09
Fix/notification problem report

bobozaur:
As per the [RFC](https://github.com/hyperledger/aries-rfcs/tree/main/features/0015-acks#relationship-to-problem-report) the `ProblemReport` message can be a part of the `notification` message family (which only contains the `ack` message that's borrowed to other protocols).

This PR adds a `ProblemReport` message to the protocol to match the AFJ implementation (which is, technically, correct).


aries_vcx/src/protocols/issuance/holder/state_machine.rs
```diff
@@ -176,6 +177,11 @@ impl HolderSM {
                             return Some((uid, message));
                         }
                     }
+                    AriesMessage::Notification(Notification::ProblemReport(msg)) => {
+                        if matches_opt_thread_id!(msg, self.thread_id.as_str()) {
```
Patrik-Stas, line 181: Slightly out of PR scope, but consider this modification while we are at it:

```
#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq)]
pub struct ProblemReportDecorators {
    #[serde(rename = "~thread")]
    pub thread: Option<Thread>,
```
I am not sure if thread would ever be None. It's not expected to send problem reports out of blue, I think it would always be in relation to some conversation, hence I would expect ~thread decorator to always be present for that type of message.

aries_vcx/src/protocols/proof_presentation/verifier/state_machine.rs
```diff
@@ -239,12 +243,28 @@ impl VerifierSM {
...
                         let problem_report = build_problem_report_msg(Some(err.to_string()), &self.thread_id);
+
                         let sm = match err.kind() {
                             AriesVcxErrorKind::InvalidProof => VerifierFullState::Finished(
                                 (state, presentation, PresentationVerificationStatus::Invalid).into(),
                             ),
                             _ => VerifierFullState::Finished((state, problem_report.clone()).into()),
                         };
+
+                        // Use the ProblemReport message as part of
```
Patrik-Stas, line 254: Given that ProblemReport in Notification family is quite explicitly expressed with new type `NotificationProblemReport` I find this comment to not carry much information. But perhaps including link to
https://github.com/hyperledger/aries-rfcs/tree/main/features/0015-acks#relationship-to-problem-report
instead would be nice here

aries_vcx/src/protocols/issuance/holder/state_machine.rs
Patrik-Stas: i see, good observation. Yeah I also don't know, anyway not a big deal. Let's consider this resolved

Patrik-Stas (dismissed):

aries_vcx/src/protocols/proof_presentation/verifier/state_machine.rs
bobozaur: Removed.

The link itself would be more suitable in the type definition itself, for `ack` messages as well. Should I add them as part of this PR?

wrappers/vcx-napi-rs/Cargo.lock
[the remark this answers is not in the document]
bobozaur: It has a reason. It's useless. The metadata is held at the workspace level.

Patrik-Stas (approved):

PR #837, opened 2023-05-10
Add indy-vdr-proxy profile

mirgee:
Adds a new profile `VdrProxyProfile` which uses a running instance of [indy-vdr-proxy](https://github.com/hyperledger/indy-vdr/tree/main/indy-vdr-proxy) (through [indy-vdr-proxy-client](https://github.com/mirgee/indy-vdr/pull/1/files)) to read and write transactions instead of communicating with the ledger directly.

This may be useful in situations where
* a service is limited in its access to the public internet for security reasons or
* the client wants quick and easy access to a ledger without necessarily having to obtain the genesis file and set up and maintain their own ledger connection.

The `VdrProxyProfile` profile reuses existing `IndyVdrLedger` implementation of the `BaseLedger` trait and request submitting is injected through an implementation of the `RequestSubmitter` trait.


aries_vcx_core/Cargo.toml
```diff
@@ -39,6 +40,7 @@ lazy_static = "1.4.0"
 derive_builder = "0.12.0"
 uuid = { version = "1.3.0", default-features = false, features = ["v4"] }
 tokio = { version = "1.20" }
+indy-vdr-proxy-client = { git = "https://github.com/mirgee/indy-vdr.git", rev = "fab0535", optional = true }
```
Patrik-Stas, line 43: Very cool 👍 but I've noticed you perhaps by mistake created PR against your own repo, rather against hyperledger origin https://github.com/mirgee/indy-vdr/pull/1

mirgee: This is on purpose, I first want to make the maintainers are actually interested in the client before making a PR.

aries_vcx_core/src/errors/mapping_indyvdr_proxy.rs
```diff
@@ -0,0 +1,12 @@
+use indy_vdr_proxy_client::error::VdrProxyClientError;
+
+use super::error::{AriesVcxCoreError, AriesVcxCoreErrorKind};
+
+impl From<VdrProxyClientError> for AriesVcxCoreError {
```
Patrik-Stas, line 5: Currently we have 3 different ledger errors
```
#[error("Connection to Pool Ledger.")]
    PoolLedgerConnect,
    #[error("Ledger rejected submitted request.")]
    InvalidLedgerResponse,
    #[error("Ledger item not found.")]
    LedgerItemNotFound,
```

when using `VdrProxy` ledger trait implementation, all errors are squashed into `AriesVcxCoreErrorKind::InvalidLedgerResponse,` - that's fine with the first two, but remapping `LedgerItemNotFound` into `InvalidLedgerResponse` can cause some troubles. For example function `_try_get_cred_def_from_ledger` is handling `LedgerItemNotFound` in very particular way. 

It would be good if we could on level of `indy-vdr-proxy-client` distinquish and return "item not not found on ledger" vs "other issue with ledger response" errors, and that would enable us propagate more granular handling on aries-vcx level as well

aries_vcx_core/src/ledger/indy_vdr_ledger.rs
Patrik-Stas, line 1: vdr is very generic term, it's not obvious from file name as whether to relates to native indy client or indy proxy client. 
I'd suggest `indy_vdr_proxy_ledger` or `indy_proxy_ledger`. Then equally I'd rename `indy_ledger.rs` to `indy_native_ledger`/`indy_direct_ledger` or so

aries_vcx_core/src/errors/mapping_indyvdr_proxy.rs
Patrik-Stas: Just realized this is actually a problem of wider scope - `vdr_ledger` submitter, eg previously `indy_vdr_ledger.rs` also behaves differently 
```
let send_req_result: VdrSendRequestResult = recv
            .await
            .map_err(|e| AriesVcxCoreError::from_msg(AriesVcxCoreErrorKind::InvalidState, e))?;
        let (result, _) = send_req_result?;

```
It actually returns even different error.

I think you can still line up the error handling as suggested in previous comment, but I think what needs to happen in the end is:

- take out the ledger trait as separate crate
- the trait needs to return custom error type - which could be for example be enum with variants `PoolLedgerConnect`, `InvalidLedgerResponse`, `LedgerItemNotFound`
- Add some documentation to traits specifying what error needs to be returned in what case, ideally covered with some tests against the trait interface which any implementor of the trait could reuse (not sure how this would be practically done)

This should solve the deviations in error handling of the trait implementation, which now returns `AriesVcxCoreError` - way too much freedom for trait implementors.

Patrik-Stas (changes requested):
left some comments

aries_vcx_core/Cargo.toml
mirgee: In the end decided to open a PR against the Hyperledger repo. But IMO merging it should not be in any way a precondition for merging this PR.

Patrik-Stas: agreed

aries_vcx_core/src/ledger/indy_vdr_ledger.rs
Patrik-Stas, line 1: Per our discussion, I'll conclude things we discussed privately, for transparency and other reviewers

- there's lots of manual parsing going on here, because the `indy-vdr` doesn't contain model of ledger responses - the assumption is that `indy-vdr` has been typically used via FFI so far, so there wasn't value for implementors to maintained detailed types for ledger responses
- there are indy ledger responses in vdrtools under `libvdrtools/src/domain/ledger/*.rs` - we could reuse that, either pull it out from vdrtool as a small crate within `aries-vcx` repo, or even better make contribution to `indy-vdr` where we would map the ledger responses to those types. 

But we could do that in separate PR

Patrik-Stas (approved):

gmulhearn-anonyome:
`code-coverage-aries-vcx-modular-dependencies-integration-tests` is failing. I suspect because alot of `SetupProfile::run_indy` code has been changed to `SetupProfile::run`. where previously, `SetupProfile::run_indy` was being used to force modular_libs tests to avoid calling onto un-implemented modular_libs code (e.g. credx issuer functionality).

Would be ideal if `code-coverage-aries-vcx-modular-dependencies-integration-tests` can continue to run and pass. I'm not entirely sure how you could approach this to handle all 3 testing cases (vdrtools, modular_libs, vdr-proxy)..

aries_vcx/src/common/primitives/mod.rs
```diff
@@ -117,7 +117,7 @@ pub mod integration_tests {
     #[tokio::test]
     #[ignore]
     async fn test_pool_from_pool_ledger_with_id() {
```
Patrik-Stas, line 119: I know you didn't touch this, but let's give this a better name, perhaps `test_pool_create_and_get_schema`

Patrik-Stas (changes requested):
almost there

aries_vcx/src/common/anoncreds.rs
```diff
@@ -25,7 +25,7 @@ pub mod integration_tests {
     #[tokio::test]
     #[ignore]
     async fn test_pool_prover_get_credentials() {
-        SetupProfile::run(|setup| async move {
+        SetupProfile::run_indy(|setup| async move {
```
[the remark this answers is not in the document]
mirgee: Accident.

aries_vcx/src/common/primitives/mod.rs
mirgee: Done

Patrik-Stas (approved):

Patrik-Stas (approved):

aries_vcx_core/Cargo.toml
gmulhearn-anonyome: I agree, but it would be good to make a comment above this dependency; stating the changes you made and also that it is future TODO to monitor your indy-vdr PR and revert to using the main branch if/when your indy-vdr PR is merged

gmulhearn-anonyome (commented):
Looks really good, splitting the logic into a request_submitter was a good move.

Just noticed a couple of things:
1. `vdr_reader.rs` in `did_resolver_sov` may need to be updated to use the new request_submitter. It currently fails to build with `cargo build --features modular_libs --no-default-features`
2. The `vdr_proxy_profile` technically depends on the `vdrtools` dependency/feature-flag. I think this is ok, but I wonder if the `vdr_proxy_profile` would therefore need to be to hidden behind the `vdrtools` flag as well? e.g. `aries_vcx` currently fails to build with `cargo build --features vdr_proxy_ledger --no-default-features`. But maybe it's that's ok and you can ignore this comment, but something to think about.

Also had that other comment about commenting on your usage of indy-vdr fork.

Feel free to merge after you've addressed/considered my review! thnx

gmulhearn (approved):
LGTM assuming CI still good

Patrik-Stas (approved):

PR #842, opened 2023-05-13
refactor maybe_known.rs: MaybeKnown enum

### LINKED ISSUE #839: Move MaybeKnown enum to shared_vcx
The `MaybeKnown<T>` enum is used for deserializing a type which could also deserialize to `String`, essentially trying the `Known<T>` variant first with the specified type (generally some other enum `T`) and falling back to the `Unknown(String)` variant if the first one fails.

This is right now declared in the `messages` crate as that's where it is mainly used, but does not really belong there as it fits a more general purpose, hence should be moved to the `shared_vcx` crate.

Moreover, the type could be extended to fit some other use cases by using a second generic as `Unknown(U)` (which could very well default to `String` as that's the most common use case).

Steps:
    - [ ] Move `MaybeKnown<T>` to `shared_vcx`.
    - [ ] To make it more extensible, add a second generic to the type which defaults to `String` (as that's the most common use case and this way existing code wouldn't break).

nachiketkanore:
- fixes #839


messages/src/maybe_known.rs
```diff
@@ -1,21 +1,22 @@
...
-    Known(T),
-    Unknown(String),
-}
+// #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+// #[serde(untagged)]
+// pub enum MaybeKnown<T> {
+//     Known(T),
+//     Unknown(String),
+// }
+// implementation moved to shared_vcx::maybe_known::MaybeKnown
```
bobozaur, line 12: Delete this file, please. I assume you wanted an easy way to revert if needed, but you'd have to modify the paths anyway, and that's probably the most cumbersome part 😅

shared_vcx/src/maybe_known.rs
```diff
@@ -0,0 +1,13 @@
+use serde::{Deserialize, Serialize};
+
+/// Enum used to encapsulate `string-like` types which may
+/// have variants we haven't implement yet.
+///
+/// Deserialization will be first attempted to the [`MaybeKnown::Known`] variant
+/// and then, if that fails, to the [`MaybeKnown::Unknown`] variant.
```
bobozaur, line 7: The `string-like` thing is the default, but not the only way to use this. The comment should be updated to reflect that.

shared_vcx/src/maybe_known.rs
```diff
@@ -0,0 +1,13 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(untagged)]
+pub enum MaybeKnown<T, U = String> {
+    Known(T),
+    Unknown(U),
+}
```
bobozaur, line 13: Some tests would be nice. Previously the `Protocol` type was used for this, but it be better to keep the tests contained to this crate and just make up some enums in the test module.

bobozaur (changes requested):
Great job! There are just some small nitpicks that I commented on, but overall looks good. Thank you for picking this up!

shared_vcx/src/maybe_known.rs
nachiketkanore: The tests couldn't be moved to the new location, because there's a cyclic dependency introduced because of `Protocol`.

bobozaur: Right, which is why the way to go would be creating one or two enums in the tests module only, one that takes advantage of the `String` default and maybe one that uses something else for the `Unknown` variant.

bobozaur:
@nachiketkanore Did you merge `main` by any chance? If that's the case, we'd prefer if you rebased instead.

nachiketkanore:
Yes, I merged

bobozaur:
> Yes, I merged

Would you be so kind to rebase instead?

nachiketkanore:
> > Yes, I merged
> 
> Would you be so kind to rebase instead?

Sorry for not following the guidelines. I am new to this, will check and update accordingly!

bobozaur (approved):
LGTM overall, though I'd find it awesome if you also used some custom type for the `Unknown` variant as well. But that can also be done sometime later and it's not really a dealbreaker.

Provided that the CI runs successfully, I'm okay with merging this.

Issue #814, opened 2023-04-27
Typed interfaces/traits for BaseWallet


gmulhearn:
# Context
In `aries_vcx_core/src/wallet/base_wallet.rs` a trait `BaseWallet` is defined. When `BaseWallet` was originally defined, it kept the implementation 'simple' by taking and returning structures in the form of JSON Strings rather than typed rust `struct`s. Now that `BaseWallet` is well integrated, we should go back and properly add types to the `BaseWallet` interface.

# Requirements
* All JSON strings inputs and return types in `BaseWallet` should be replaced with rust structs.
* implementations of `BaseWallet` (currently mostly just `IndySdkWallet`) should be updated to use the new types and translate them appropriately
* consumers of `BaseWallet` in the code base should be updated to use the new types rather than the JSON strings

# Examples:
## Example 1:
take this method;
```
    async fn add_wallet_record(&self, xtype: &str, id: &str, value: &str, tags_json: Option<&str>)
        -> VcxCoreResult<()>;
```

`tags_json` is an optional JSON string that represents a `HashMap<String, String>`.

so the method should be updated as:
```
    async fn add_wallet_record(&self, xtype: &str, id: &str, value: &str, tags: Option<HashMap<String,String>>)
        -> VcxCoreResult<()>;
```

## Example 2
```
    async fn get_wallet_record(&self, xtype: &str, id: &str, options_json: &str) -> VcxCoreResult<String>;
```

This method takes a JSON string via `options_json` and returns a JSON string that represents a "wallet record"

they should both be replaced with struct types.

If we drill into the `libvdrtools` implementation of `get_record` (drilled into via the IndySdkWallet::get_wallet_record implementation), we get a clue for what these structures would look like:
```
    /// options_json: //TODO: FIXME: Think about replacing by bitmask
    ///  {
    ///    retrieveType: (optional, false by default) Retrieve record type,
    ///    retrieveValue: (optional, true by default) Retrieve record value,
    ///    retrieveTags: (optional, false by default) Retrieve record tags
    ///  }
    /// #Returns
    /// wallet record json:
    /// {
    ///   id: "Some id",
    ///   type: "Some type", // present only if retrieveType set to true
    ///   value: "Some value", // present only if retrieveValue set to true
    ///   tags: <tags json>, // present only if retrieveTags set to true
    /// }
```

# Tips
[...]

PR #843, opened 2023-05-14
Refactor basewallet and corresponding consumers

tech-bash:
fix: https://github.com/hyperledger/aries-vcx/issues/814

- All JSON strings inputs replaced with rust structs.
- consumers of `BaseWallet` in the code base  updated to use the new types rather than the JSON strings


Patrik-Stas:
hi @tech-bash note the failing jobs - take a look how we run those in CI and try to reproduce locally

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -606,10 +606,10 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
         let credential_id = cred_id.map_or(Uuid::new_v4().to_string(), String::from);
 
         let record_value = serde_json::to_string(&credential)?;
-        let tags_json = serde_json::to_string(&tags)?;
+        let tags_json: HashMap<String, String> = serde_json::from_str(&record_value)?;
```
gmulhearn, line 609: should not be serde_jsoning over `record_value` should be over `&tags`

gmulhearn: let tags_json: HashMap<String, String> = serde_json::from_value(tags)?;

aries_vcx_core/src/wallet/indy_wallet.rs
```diff
@@ -52,33 +53,45 @@ impl BaseWallet for IndySdkWallet {
...
     ) -> VcxCoreResult<()> {
+        let res = tags.map(|x| serde_json::to_string(&x)).transpose()?.to_owned();
+        let tags_json = res.as_deref();
         indy::wallet::add_wallet_record(self.wallet_handle, xtype, id, value, tags_json).await
     }
 
-    async fn get_wallet_record(&self, xtype: &str, id: &str, options_json: &str) -> VcxCoreResult<String> {
-        indy::wallet::get_wallet_record(self.wallet_handle, xtype, id, options_json).await
+    async fn get_wallet_record(&self, xtype: &str, id: &str, options: String) -> VcxCoreResult<String> {
+        let options_json = serde_json::to_string(&options)?;
```
gmulhearn, line 64: this is not neccessary, and converts the options into a weird string. you should be able to just pass `&options` straight into indy::waLLET::get_wallet_record

gmulhearn (changes requested):
Hey, nice work so far!

I noticed in some places you've changed the parameters of `BaseWallet` methods to take `String` instead of `&str`. For consistency, could you please change them back?

E.g.
you've changed:
```
async fn update_wallet_record_value(&self, xtype: &str, id: &str, value: &str)
```
to:
```
async fn update_wallet_record_value(&self, xtype: &str, id: &str, value: String)
```

but `value` should stay as `&str`.

This happens in a few other places too, just have a look thru the diff on this PR. thanks!

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
tech-bash: I did this on the first place, but it was not working so I was just trying things up!

gmulhearn: Just speculating, but perhaps when it failed, you were using serde_json::from_str. serde_json::from_value is what we'd use in this situation (as tags is a Value)

gmulhearn: Alternatively, we could redo the entire construction of tags to be a HashMap from the start, rather than constructing it as a Value and then converting to HashMap. 

However that's not necessarily in scope for this PR, can be fixed up later

tech-bash: > Just speculating, but perhaps when it failed, you were using serde_json::from_str. serde_json::from_value is what we'd use in this situation (as tags is a Value)
i actually tried with `from_value()` as soon as I got to know that tags are a value, but as far as I remember it was being converted to a diff type to what was needed!

libvcx_core/src/api_vcx/api_global/wallet.rs
```diff
@@ -123,16 +126,19 @@ pub async fn wallet_update_wallet_record_value(xtype: &str, id: &str, value: &st
...
 pub async fn wallet_add_wallet_record_tags(xtype: &str, id: &str, tags_json: &str) -> LibvcxResult<()> {
     let wallet = get_main_wallet();
-    map_ariesvcx_core_result(wallet.add_wallet_record_tags(xtype, id, tags_json).await)
+    let tags: HashMap<String, String> = serde_json::from_str(tags_json)?;
+    map_ariesvcx_core_result(wallet.add_wallet_record_tags(xtype, id, tags).await)
 }
 
 pub async fn wallet_delete_wallet_record_tags(xtype: &str, id: &str, tags_json: &str) -> LibvcxResult<()> {
     let wallet = get_main_wallet();
+    let tags_json = serde_json::from_str(tags_json)?;
```
gmulhearn, line 141: I believe this is not necessary and may cause bugs (unfortunately I believe this might be untested code, so CI is not reporting issues).

`wallet.delete_wallet_record_tags` still just takes in a regular `&str`. In the future we should fix the typing for `delete_wallet_record_tags`, but that can be a job for later.

gmulhearn (changes requested):
looking good just 1 comment for now

libvcx_core/src/api_vcx/api_global/wallet.rs
tech-bash: alright!

aries_vcx_core/src/wallet/base_wallet.rs
```diff
@@ -26,18 +28,28 @@ pub trait BaseWallet: std::fmt::Debug + Send + Sync {
...
+    async fn add_wallet_record(
+        &self,
+        xtype: &str,
+        id: &str,
+        value: &str,
+        tags: Option<HashMap<String, String>>,
+    ) -> VcxCoreResult<()>;
 
-    async fn get_wallet_record(&self, xtype: &str, id: &str, options_json: &str) -> VcxCoreResult<String>;
+    async fn get_wallet_record(&self, xtype: &str, id: &str, options: &str) -> VcxCoreResult<String>;
```
Patrik-Stas, line 39: The `options` could actually be well typed, but its type depends on what kind of options the underlying implementation accepts.

This is slightly more advanced, but we could use associated types here. You would defined something like

```
 pub trait BaseWallet: std::fmt::Debug + Send + Sync {
    type GetRecordOptions;
    type AddRecordOptions;

   ...
   ...
}
```
Then in particular implementation of the trait, for example in `impl IndySdkWallet `, you would inject the actual types. And of course, you would have to create those types (you can find out what the option objects should look like if you dig into `libvdrtools` code)

 So not a hard requirement, but if you could do this it would be really great!

Patrik-Stas (changes requested):
Thank you for PR, left some comments.

- One to address
- Second one to consider as bit more advanced improvement

aries_vcx_core/src/wallet/indy_wallet.rs
```diff
@@ -69,12 +72,19 @@ impl BaseWallet for IndySdkWallet {
...
+        tags: HashMap<String, String>,
+    ) -> VcxCoreResult<()> {
+        let tags_json = serde_json::to_string(&tags)?;
+        indy::wallet::update_wallet_record_tags(self.wallet_handle, xtype, id, &tags_json).await
     }
 
-    async fn add_wallet_record_tags(&self, xtype: &str, id: &str, tags_json: &str) -> VcxCoreResult<()> {
-        indy::wallet::add_wallet_record_tags(self.wallet_handle, xtype, id, tags_json).await
+    async fn add_wallet_record_tags(&self, xtype: &str, id: &str, tags: HashMap<String, String>) -> VcxCoreResult<()> {
+        let tags_json = serde_json::to_string(&tags)?;
```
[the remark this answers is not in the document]
tech-bash: We are taking the typed parameters and then converting them back to JSON before passing them `indy::wallet::abc`.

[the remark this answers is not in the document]
tech-bash: are you intending to say that we should only have hashmap as the parameter type and not &str?

aries_vcx_core/src/wallet/base_wallet.rs
tech-bash: alright!

Patrik-Stas (approved):

Issue #847, opened 2023-05-15
Typing of JSON strings on Prover handler APIs


gmulhearn:
There are a few APIs in the aries_vcx Prover implementation which use complicated JSON for API inputs & outputs. These should ideally be replaced with typed structs.

# retrieve_credentials
```rust
pub async fn retrieve_credentials(&self, profile: &Arc<dyn Profile>) -> VcxResult<String>
```
where the return JSON String should be replaced with some `RetrievedCredentials` struct.

looking at the internals of an implementation (specifically in `prover.rs` `fetch_credential`), we can see the intended structure:
```json
{
  "attrs": {
    "<referent (pred or attr referent)>": [{
      "cred_info": {
        "referent": "<credential identifier>",
        "attrs": { "<attr_name>": "<attr_raw_value>", ... },
        "schema_id": "...",
        "cred_def_id": "...",
        "rev_reg_id"?: "..."?,
        "cred_rev_id"?: "..."?,
      },
      "non_revoc_interval": { "from"?: int?, "to"?: int? },
    }],
  }
}
```

# generate_presentation
```rust
pub async fn generate_presentation(
        &mut self,
        profile: &Arc<dyn Profile>,
        credentials: String,
        self_attested_attrs: String,
    ) -> VcxResult<()>
```

where `credentials` should be replaced with some `SelectedCredentials` struct. and `self_attested_attrs` should be replaced with `HashMap<String, String>`.

looking at the internals of aries_vcx (specifically in `prover_internal.rs` `fn credential_def_identifiers`), we can see the intended structure of `credentials`:

```json
{
  "attrs": {
    "<referent (pred or attr referent)>": {
      "credential": {
        "cred_info": {
          "referent": "<credential identifier>",
          "schema_id": "...",
          "cred_def_id": "...",
          "rev_reg_id"?: "..."?,
          "cred_rev_id"?: "..."?,
          "revealed"?: boolean?
        }
      },
      "tails_file": "<path to tails dir>"
    }
  }
}
```

Both of these changes could be made in a non-breaking manner for libvcx consumers by if libvcx continues to take/return JSON strings, but serializes/deserializes them before passing them to aries_vcx. 

It would be a breaking change for the aries_vcx APIs though. Which I think is for the best, as IMO, aries_vcx should be embracing all the Rust type-safety-goodness that it can 🚀.

# Open questions?
[...]

PR #848, opened 2023-05-15
Issue #847: Prover Handler Types

gmulhearn:
Implements #847 

I went with the approach of keeping the legacy structures for now, they aren't toooo bad...

Changes revolve around changing the following handler APIs:
```
retrieve_credentials(&self, profile: &Arc<dyn Profile>) -> VcxResult<RetrievedCredentials>

generate_presentation(
        &mut self,
        profile: &Arc<dyn Profile>,
        credentials: SelectedCredentials,
        self_attested_attrs: HashMap<String, String>,
    ) -> VcxResult<()>
```

And then piping thru the typing in aries_vcx internally. And serde_json-ing the types externally (i.e. in our libvcx consumer code). The libvcx JSON strings should be backwards compatible still (which hopefully the CI is testing/enforcing).


gmulhearn:
@Patrik-Stas pending CI confirmation on some of my latest commits, but should be ready to review. 

If CI fails it should just be minor things I'll fix tomorrow!

bobozaur (approved):
These comments, @gmulhearn ... They are just beautiful.

PR #851, opened 2023-05-17
Ledger response parser

mirgee:
Extracts ledger response parsing capability into the `indy-ledger-response-parser` crate which
* returns types from `indy-data-types` (reexported by `indy-vdr`) and
* defines its own types (taken from `libvdrtools`) for ledger responses.

This crate is further integrated into `IndyVdrLedger` implementation to replace the current provisional constructions.

### DIFF
--- a/aries_vcx/src/core/profile/modular_libs_profile.rs
+++ b/aries_vcx/src/core/profile/modular_libs_profile.rs
@@ -3,9 +3,10 @@ use std::sync::Arc;
 use aries_vcx_core::anoncreds::base_anoncreds::BaseAnonCreds;
 use aries_vcx_core::anoncreds::credx_anoncreds::IndyCredxAnonCreds;
 use aries_vcx_core::ledger::base_ledger::BaseLedger;
-use aries_vcx_core::ledger::indy_vdr_ledger::IndyVdrLedger;
+use aries_vcx_core::ledger::indy_vdr_ledger::{IndyVdrLedger, IndyVdrLedgerConfig};
 use aries_vcx_core::ledger::request_submitter::vdr_ledger::{IndyVdrLedgerPool, IndyVdrSubmitter, LedgerPoolConfig};
 use aries_vcx_core::wallet::base_wallet::BaseWallet;
+use aries_vcx_core::ResponseParser;
 
 use crate::errors::error::VcxResult;
 
@@ -21,10 +22,16 @@ pub struct ModularLibsProfile {
 
 impl ModularLibsProfile {
     pub fn new(wallet: Arc<dyn BaseWallet>, ledger_pool_config: LedgerPoolConfig) -> VcxResult<Self> {
-        let ledger_pool = Arc::new(IndyVdrLedgerPool::new(ledger_pool_config)?);
-        let submitter = Arc::new(IndyVdrSubmitter::new(ledger_pool));
-        let ledger = Arc::new(IndyVdrLedger::new(Arc::clone(&wallet), submitter));
         let anoncreds = Arc::new(IndyCredxAnonCreds::new(Arc::clone(&wallet)));
+        let ledger_pool = Arc::new(IndyVdrLedgerPool::new(ledger_pool_config)?);
+        let request_submitter = Arc::new(IndyVdrSubmitter::new(ledger_pool));
+        let response_parser = Arc::new(ResponseParser::new());
+        let config = IndyVdrLedgerConfig {
+            wallet: wallet.clone(),
+            request_submitter,
+            response_parser,
+        };
+        let ledger = Arc::new(IndyVdrLedger::new(config));
         Ok(ModularLibsProfile {
             wallet,
             ledger,
--- a/aries_vcx_core/src/errors/mapping_vdrtools.rs
+++ b/aries_vcx_core/src/errors/mapping_vdrtools.rs
@@ -1,11 +1,23 @@
-use vdrtools::types;
-use vdrtools::types::errors::IndyErrorKind;
+#[cfg(feature = "vdrtools")]
+use vdrtools::types::{
+    errors::{IndyError, IndyErrorKind},
+    ErrorCode,
+};
+
+#[cfg(all(
+    not(feature = "vdrtools"),
+    any(feature = "modular_libs", feature = "vdr_proxy_ledger")
+))]
+use indy_ledger_response_parser::{
+    errors::{IndyError, IndyErrorKind},
+    ErrorCode,
+};
 
 use crate::errors::error::{AriesVcxCoreError, AriesVcxCoreErrorKind};
 
 impl From<IndyErrorKind> for AriesVcxCoreErrorKind {
     fn from(indy: IndyErrorKind) -> Self {
-        use types::errors::IndyErrorKind::*;
+        use IndyErrorKind::*;
 
         match indy {
             InvalidParam(_) => AriesVcxCoreErrorKind::InvalidLibindyParam,
@@ -33,15 +45,15 @@ impl From<IndyErrorKind> for AriesVcxCoreErrorKind {
             WalletAccessFailed => AriesVcxCoreErrorKind::WalletAccessFailed,
             ProofRejected => AriesVcxCoreErrorKind::ProofRejected,
             _ => {
-                let err_code = types::ErrorCode::from(indy) as u32;
+                let err_code = ErrorCode::from(indy) as u32;
                 AriesVcxCoreErrorKind::VdrToolsError(err_code)
             }
         }
     }
 }
 
-impl From<types::errors::IndyError> for AriesVcxCoreError {
-    fn from(indy: types::errors::IndyError) -> Self {
+impl From<IndyError> for AriesVcxCoreError {
+    fn from(indy: IndyError) -> Self {
         let vcx_kind: AriesVcxCoreErrorKind = indy.kind().into();
         AriesVcxCoreError::from_msg(vcx_kind, indy.to_string())
     }
--- a/aries_vcx_core/src/errors/mod.rs
+++ b/aries_vcx_core/src/errors/mod.rs
@@ -7,5 +7,5 @@ mod mapping_indyvdr;
 #[cfg(feature = "vdr_proxy_ledger")]
 mod mapping_indyvdr_proxy;
 mod mapping_others;
-#[cfg(feature = "vdrtools")]
+#[cfg(any(feature = "vdrtools", feature = "modular_libs", feature = "vdr_proxy_ledger"))]
 mod mapping_vdrtools;
--- a/aries_vcx_core/src/ledger/indy_vdr_ledger.rs
+++ b/aries_vcx_core/src/ledger/indy_vdr_ledger.rs
@@ -1,4 +1,5 @@
 use indy_credx::ursa::cl::RevocationRegistryDelta as UrsaRevocationRegistryDelta;
+use indy_ledger_response_parser::{ResponseParser, RevocationRegistryDeltaInfo, RevocationRegistryInfo};
 use indy_vdr as vdr;
 use std::fmt::{Debug, Formatter};
 use std::sync::Arc;
@@ -27,22 +28,33 @@ use crate::wallet::base_wallet::BaseWallet;
 use super::base_ledger::BaseLedger;
 use super::request_submitter::RequestSubmitter;
 
+pub struct IndyVdrLedgerConfig<T>
+where
+    T: RequestSubmitter + Send + Sync,
+{
+    pub wallet: Arc<dyn BaseWallet>,
+    pub request_submitter: Arc<T>,
+    pub response_parser: Arc<ResponseParser>,
+}
+
 pub struct IndyVdrLedger<T>
 where
     T: RequestSubmitter + Send + Sync,
 {
     wallet: Arc<dyn BaseWallet>,
     request_submitter: Arc<T>,
+    response_parser: Arc<ResponseParser>,
 }
 
 impl<T> IndyVdrLedger<T>
 where
     T: RequestSubmitter + Send + Sync,
 {
-    pub fn new(wallet: Arc<dyn BaseWallet>, request_submitter: Arc<T>) -> Self {
-        IndyVdrLedger {
-            wallet,
-            request_submitter,
+    pub fn new(config: IndyVdrLedgerConfig<T>) -> Self {
+        Self {
+            wallet: config.wallet,
+            request_submitter: config.request_submitter,
+            response_parser: config.response_parser,
         }
     }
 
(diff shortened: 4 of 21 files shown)


aries_vcx/src/core/profile/modular_libs_profile.rs
```diff
@@ -21,10 +22,16 @@ pub struct ModularLibsProfile {
...
 impl ModularLibsProfile {
     pub fn new(wallet: Arc<dyn BaseWallet>, ledger_pool_config: LedgerPoolConfig) -> VcxResult<Self> {
-        let ledger_pool = Arc::new(IndyVdrLedgerPool::new(ledger_pool_config)?);
-        let submitter = Arc::new(IndyVdrSubmitter::new(ledger_pool));
-        let ledger = Arc::new(IndyVdrLedger::new(Arc::clone(&wallet), submitter));
         let anoncreds = Arc::new(IndyCredxAnonCreds::new(Arc::clone(&wallet)));
+        let ledger_pool = Arc::new(IndyVdrLedgerPool::new(ledger_pool_config)?);
+        let request_submitter = Arc::new(IndyVdrSubmitter::new(ledger_pool));
+        let response_parser = Arc::new(ResponseParser::new());
+        let config = IndyVdrLedgerConfig {
```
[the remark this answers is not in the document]
Patrik-Stas: Maybe more related to PR - not sure if `request_submitter`, `response_parser` need to be Arc. I think `IndyVdrLedger` is going to be the only one to use them

aries_vcx_core/src/errors/mapping_vdrtools.rs
```diff
@@ -1,11 +1,23 @@
-use vdrtools::types;
-use vdrtools::types::errors::IndyErrorKind;
+#[cfg(feature = "vdrtools")]
+use vdrtools::types::{
+    errors::{IndyError, IndyErrorKind},
+    ErrorCode,
+};
+
+#[cfg(all(
+    not(feature = "vdrtools"),
```
Patrik-Stas, line 8: wouldn't simply stating this one basically imply 
```
any(feature = "modular_libs", feature = "vdr_proxy_ledger")
```
? Cause if you don't build `vdrtools`, you must be building with something else, no?

I might be missing something

aries_vcx_core/src/errors/mod.rs
```diff
@@ -7,5 +7,5 @@ mod mapping_indyvdr;
 #[cfg(feature = "vdr_proxy_ledger")]
 mod mapping_indyvdr_proxy;
 mod mapping_others;
-#[cfg(feature = "vdrtools")]
+#[cfg(any(feature = "vdrtools", feature = "modular_libs", feature = "vdr_proxy_ledger"))]
```
Patrik-Stas, line 10: isn't it redundant? Can there be case that we build without any of these?

aries_vcx_core/src/ledger/indy_vdr_ledger.rs
```diff
@@ -248,89 +260,20 @@ where
...
-    async fn get_schema(&self, schema_id: &str, submitter_did: Option<&str>) -> VcxCoreResult<String> {
-        let _ = submitter_did;
-        // TODO - future - try from cache first
-        // TODO - future - do we need to handle someone submitting a schema request by seq number?
-
-        let id = SchemaId::from_str(schema_id)?;
-
-        let request = self.request_builder()?.build_get_schema_request(None, &id)?;
-
+    async fn get_schema(&self, schema_id: &str, _submitter_did: Option<&str>) -> VcxCoreResult<String> {
```
Patrik-Stas, line 263: damn that's so much better, awesome 🥇

aries_vcx_core/src/errors/mod.rs
mirgee: It is more to clarify that one of these must be enabled to even build the module (regardless we have a "convention" of always building with one of them).

aries_vcx_core/src/errors/mapping_vdrtools.rs
mirgee: Using your suggestion, if you build with `vdrtools` (default feat. flag) and one of `modular_libs` and `vdr_proxy_ledger`, your suggestion would fail due to duplicate import.

aries_vcx/src/core/profile/modular_libs_profile.rs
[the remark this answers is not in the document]
mirgee: >  instead of putting things in this config, we could just tweak ndyVdrLedger::new signature and pass it directly. It's not like we have 10 optional things in the "config", all these things are pretty esential.

I put them into config very deliberately. I am planning to add more arguments and I am strongly convinced it is way easier to manage a single config object than many separate arguments.

aries_vcx_core/src/ledger/indy_vdr_ledger.rs
Patrik-Stas, line 1: Since we support for V0 and V1 responses, I think we should:
- either run tests with both V0 and V1 configurations (should be matter of indy-vdr configuration, what version of ledger protocol we want to use)
- or simply choose to primarily support V1, not test with V0, keep implementation as is, but log warnings anytime V0 response arrives

I suppose indy-vdr would be using V1 rn, but it would be good to make sure

Patrik-Stas, line 1: Since we support for V0 and V1 responses, I think we should:
- either run tests with both V0 and V1 configurations (should be matter of indy-vdr configuration, what version of ledger protocol we want to use)
- or simply choose to primarily support V1, not test with V0, keep implementation as is, but log warnings anytime V0 response arrives

I suppose indy-vdr would be using V1 rn, but it would be good to make sure

libvdrtools/indy-api-types/Cargo.toml
```diff
@@ -13,15 +13,15 @@ rust-base58 = ["bs58"]
...
+futures = { version = "0.3", default-features = false, features = ["std"] }
 log = { version = "0.4.17", features = ["std"] }
 libc = "0.2.114"
 openssl = {version = "0.10", optional = true}
 bs58 = {version = "0.4.0", optional = true}
 serde = "1.0.99"
 serde_json = "1.0.40"
 serde_derive = "1.0.99"
-sqlx = { version = "0.5.8", git = "https://github.com/jovfer/sqlx", branch = "feature/json_no_preserve_order_v5", features = [ "sqlite", "json_no_preserve_order" ], optional = true }
+sqlx = { version = "0.5.8", git = "https://github.com/jovfer/sqlx", branch = "feature/json_no_preserve_order_v5", features = [ "sqlite", "json_no_preserve_order", "runtime-tokio-rustls" ], optional = true }
```
Patrik-Stas, line 24: seems unrelated to pr, why the addition of `runtime-tokio-rustls`?

libvdrtools/indy-api-types/Cargo.toml
```diff
@@ -13,15 +13,15 @@ rust-base58 = ["bs58"]
 
 [dependencies]
 failure = "0.1.8"
-futures = { version = "0.3", default-features = false }
+futures = { version = "0.3", default-features = false, features = ["std"] }
```
Patrik-Stas, line 16: why needed?

Patrik-Stas (changes requested):
left some comments but nothing big, this is really great improvement

libvdrtools/indy-api-types/Cargo.toml
mirgee: On the `main` branch, `indy-api-types` and some of its dependents fail to build by themselves because

> one of the features ['runtime-actix-native-tls', 'runtime-async-std-native-tls', 'runtime-tokio-native-tls', 'runtime-actix-rustls', 'runtime-async-std-rustls', 'runtime-tokio-rustls'] must be enabled

The reason why `libvdrtools` or `aries_vcx` don't fail to build is because they add one of these features higher up the dependency chain.

libvdrtools/indy-api-types/Cargo.toml
mirgee: Same reason as above: `indy-api-types` fails to build without it. Actually, thanks to your comment, I found out the piece of code using the feature seems to have been dead, so I removed it along with the feature.

aries_vcx_core/src/ledger/indy_vdr_ledger.rs
mirgee: I am not sure I understand what you mean by "right now", but I think `indy-vdr` can be configured to use either protocol version and we set V1 by default.

Patrik-Stas (approved):

bobozaur (changes requested):
Looks good, but I think there are some improvements that can be made, especially regarding `parse_message`.

indy_ledger_response_parser/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+        )?; // FIXME: Review how we handle not found
+
+        match message {
+            Message::Reject(response) | Message::ReqNACK(response) => Err(err_msg(
+                IndyErrorKind::InvalidTransaction,
+                format!("Transaction has been failed: {:?}", response.reason),
+            )),
+            Message::Reply(reply) => Ok(reply),
+        }
+    }
```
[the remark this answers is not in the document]
mirgee: I agree that this piece is an ugly workaround. Your suggestion is good in spirit, but it needed a little bit of tweaking to work, as it incorrectly assumes that `type` is a sibling field of `op`, but is in fact part of `Reply` result. To be specific, this is a simplified version of a response:
```
{
  "op": "REPLY",
  "result": {
     ...
    "type": "107"
  }
}
```
(The part about `#[serde(try_from = "MsgWithReplyType")]` on `Message` seems to contradict the rest of your suggestion, where you are proposing to deserialize into `MsgWithTypedReply`, and then convert directly into `Reply`. Moreover, assuming you mean `MsgWithTypedReply`, it would also take generic parameters to account for. But I assume it is no longer relevant.)

indy_ledger_response_parser/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+                        format!("Nym not found"),
+                    ))
+                    .and_then(|data| {
+                        serde_json::from_str(&data).map_err(|err| {
+                            IndyError::from_msg(
+                                IndyErrorKind::InvalidState,
+                                format!("Cannot parse GET_NYM response: {}", err),
+                            )
+                        })
+                    })?;
```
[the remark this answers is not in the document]
mirgee: Good point about `ok_or_else`, addressed. I don't find extracting into a variable more readable.

indy_ledger_response_parser/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        let (revoc_reg_def_id, revoc_reg) = match reply.result() {
+            GetRevocRegDeltaReplyResult::GetRevocRegDeltaReplyResultV0(res) => (res.revoc_reg_def_id, res.data),
+            GetRevocRegDeltaReplyResult::GetRevocRegDeltaReplyResultV1(res) => {
+                (res.txn.data.revoc_reg_def_id, res.txn.data.value)
+            }
+        };
+
+        let revoc_reg_delta = RevocationRegistryDeltaV1 {
+            value: json!(UrsaRevocationDelta::from_parts(
```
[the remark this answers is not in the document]
mirgee: Good point! Fixed.

indy_ledger_response_parser/src/lib.rs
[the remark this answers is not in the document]
bobozaur: Aaaahhh, my bad! It was obvious but I missed that. It can still be done in the same fashion though, but it becomes a bit annoying because of the manually specified bounds. I'll close this and start a new convo to make the suggestion.

indy_ledger_response_parser/src/domain/response.rs
```diff
@@ -0,0 +1,129 @@
...
+{
+    type Error = IndyError;
+    fn try_from(value: MessageWithTypedReply<'a, T>) -> Result<Self, Self::Error> {
+        match value {
+            MessageWithTypedReply::ReqNACK(r) => Ok(Message::ReqNACK(r)),
+            MessageWithTypedReply::Reply(r) => Ok(Message::Reply(r.result().try_into()?)),
+            MessageWithTypedReply::Reject(r) => Ok(Message::Reject(r)),
+        }
+    }
+}
```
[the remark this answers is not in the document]
mirgee: It doesn't work, you seem to have reverted the change which actually made it possible to deserialize into `MessageWithTypedReply`.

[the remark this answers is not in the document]
bobozaur: Ok lemme actually test. Give me a few minutes.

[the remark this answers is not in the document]
bobozaur: Yeah so, if there was any doubt that I'm an idiot, I just provided the confirmation 😅 . I made the exact same mistake that you pointed out before, about the `type` being a field in the `result` and not alongside it.

Modified the suggestion (after testing) to accommodate the field being at the right level in the JSON.

Tested locally with this:
```rust
#[test]
fn test_reply_deserialization() {
    #[derive(Deserialize, Debug)]
    struct Test {
        field: u8,
    }

    impl ReplyType for Test {
        fn get_type<'a>() -> &'a str {
            "107"
        }
    }

    let raw_str = r#"{
        "op": "REPLY",
        "result": {
            "field": 0,
            "type": "107"
        }
      }"#;

    let reply: Message<Test> = serde_json::from_str(raw_str).unwrap();
    println!("{reply:?}");
}
```

indy_ledger_response_parser/src/domain/response.rs
```diff
@@ -0,0 +1,129 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct ReplyV1<T> {
+    pub data: ReplyDataV1<T>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct ReplyDataV1<T> {
+    pub result: Vec<ReplyV0<T>>,
```
bobozaur, line 41: While working on the suggestion to avoid double deserialization, I was wondering why is this a `Vec`, since it looks like in the `result()` method the first element is removed and the others, if any, are lost. If there's only one element to be expected, how about making it an array?
The element can then be extracted with `result.into_iter().next().unwrap()` which is sound since you cannot instantiate an unfilled array.

```suggestion
    pub result: [ReplyV0<T>; 1],
```

bobozaur (commented):
Modified the last suggestion for the `Message<T>`, and also wondering about this `Vec` in `ReplyV1`. I'm curious of what your thoughts are on this.

indy_ledger_response_parser/src/domain/response.rs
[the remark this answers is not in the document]
mirgee: Good, the `Reply` version is preserved now! Applied the suggestion.

indy_ledger_response_parser/src/lib.rs
```diff
@@ -0,0 +1,244 @@
+#[macro_use]
+extern crate serde;
+
+#[macro_use]
+extern crate serde_json;
+
+mod domain;
+
+use domain::response::MessageWithTypedReply;
```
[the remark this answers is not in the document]
mirgee: Removed.

indy_ledger_response_parser/src/domain/response.rs
mirgee: If there is always only one element to be expected, I don't see the point of making it any kind of array in the first place. I don't know why it is a `Vec` - perhaps this is to support some sort of batch transactions in the future...? Then the current implementation of using only the first element would be a temporary workaround, and so I would probably prefer to not accommodate the type to the current use case. But again, as I said, I don't know.

bobozaur: Why not deal with the idea of batch transactions if/when it will be a thing? Whether the workaround is temporary or not, I'd say it could at least be a better workaround. But feel free to ignore this and resolve the conversation if it's not important for you.

bobozaur (dismissed):
LGTM

indy_ledger_response_parser/src/domain/response.rs
mirgee: Alright, changed the `Vec` to an array.

bobozaur (approved):
LGTM

Patrik-Stas (approved):

PR #854, opened 2023-05-18
Feature/credx issuer

bobozaur:
This PR completes the credx anoncreds implementation by adding proper handling for issuer operations. For tests, the `mixed_breed` feature flag was added (with teary eyes, since we want to get rid of them not add more) so that a new profile can be constructed.

The mixed breed profile uses the credx anoncreds impl and the indy ledger and wallet impls.

The code here is bad from a lot of perpectives, but to me that's not a problem since this will all be deleted/refactored soon enough as we transition to `anoncreds-rs`.

### DIFF
--- a/aries_vcx/src/core/profile/mixed_breed_profile.rs
+++ b/aries_vcx/src/core/profile/mixed_breed_profile.rs
@@ -4,7 +4,7 @@ use aries_vcx_core::{
     anoncreds::{base_anoncreds::BaseAnonCreds, credx_anoncreds::IndyCredxAnonCreds},
     ledger::{
         base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerWrite, IndyLedgerRead, IndyLedgerWrite},
-        indy_ledger::IndySdkLedger,
+        indy_ledger::{IndySdkLedgerRead, IndySdkLedgerWrite},
     },
     wallet::{base_wallet::BaseWallet, indy_wallet::IndySdkWallet},
     PoolHandle, WalletHandle,
@@ -26,15 +26,16 @@ impl MixedBreedProfile {
     pub fn new(indy_wallet_handle: WalletHandle, indy_pool_handle: PoolHandle) -> Self {
         let wallet: Arc<dyn BaseWallet> = Arc::new(IndySdkWallet::new(indy_wallet_handle));
         let anoncreds = Arc::new(IndyCredxAnonCreds::new(Arc::clone(&wallet)));
-        let ledger = Arc::new(IndySdkLedger::new(indy_wallet_handle, indy_pool_handle));
+        let ledger_read = Arc::new(IndySdkLedgerRead::new(indy_wallet_handle, indy_pool_handle));
+        let ledger_write = Arc::new(IndySdkLedgerWrite::new(indy_wallet_handle, indy_pool_handle));
 
         MixedBreedProfile {
             wallet,
             anoncreds,
-            anoncreds_ledger_read: ledger.clone(),
-            anoncreds_ledger_write: ledger.clone(),
-            indy_ledger_read: ledger.clone(),
-            indy_ledger_write: ledger,
+            anoncreds_ledger_read: ledger_read.clone(),
+            anoncreds_ledger_write: ledger_write.clone(),
+            indy_ledger_read: ledger_read,
+            indy_ledger_write: ledger_write,
         }
     }
 }
--- a/aries_vcx/src/core/profile/mixed_breed_profile.rs
+++ b/aries_vcx/src/core/profile/mixed_breed_profile.rs
@@ -1,0 +1,44 @@
+use std::sync::Arc;
+
+use aries_vcx_core::{
+    anoncreds::{base_anoncreds::BaseAnonCreds, credx_anoncreds::IndyCredxAnonCreds, indy_anoncreds::IndySdkAnonCreds},
+    ledger::{base_ledger::BaseLedger, indy_ledger::IndySdkLedger},
+    wallet::{base_wallet::BaseWallet, indy_wallet::IndySdkWallet},
+    PoolHandle, WalletHandle,
+};
+
+use super::profile::Profile;
+
+#[derive(Debug)]
+pub struct MixedBreedProfile {
+    wallet: Arc<dyn BaseWallet>,
+    ledger: Arc<dyn BaseLedger>,
+    anoncreds: Arc<dyn BaseAnonCreds>,
+}
+
+impl MixedBreedProfile {
+    pub fn new(indy_wallet_handle: WalletHandle, indy_pool_handle: PoolHandle) -> Self {
+        let wallet = Arc::new(IndySdkWallet::new(indy_wallet_handle));
+        let ledger = Arc::new(IndySdkLedger::new(indy_wallet_handle, indy_pool_handle));
+        let anoncreds: Arc<IndyCredxAnonCreds> = Arc::new(IndyCredxAnonCreds::new(wallet.clone()));
+        MixedBreedProfile {
+            wallet,
+            ledger,
+            anoncreds,
+        }
+    }
+}
+
+impl Profile for MixedBreedProfile {
+    fn inject_ledger(self: Arc<Self>) -> Arc<dyn BaseLedger> {
+        Arc::clone(&self.ledger)
+    }
+
+    fn inject_anoncreds(self: Arc<Self>) -> Arc<dyn BaseAnonCreds> {
+        Arc::clone(&self.anoncreds)
+    }
+
+    fn inject_wallet(&self) -> Arc<dyn BaseWallet> {
+        Arc::clone(&self.wallet)
+    }
+}
--- a/aries_vcx/src/core/profile/mixed_breed_profile.rs
+++ b/aries_vcx/src/core/profile/mixed_breed_profile.rs
@@ -1,8 +1,11 @@
 use std::sync::Arc;
 
 use aries_vcx_core::{
-    anoncreds::{base_anoncreds::BaseAnonCreds, credx_anoncreds::IndyCredxAnonCreds, indy_anoncreds::IndySdkAnonCreds},
-    ledger::{base_ledger::BaseLedger, indy_ledger::IndySdkLedger},
+    anoncreds::{base_anoncreds::BaseAnonCreds, credx_anoncreds::IndyCredxAnonCreds},
+    ledger::{
+        base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerWrite, IndyLedgerRead, IndyLedgerWrite},
+        indy_ledger::IndySdkLedger,
+    },
     wallet::{base_wallet::BaseWallet, indy_wallet::IndySdkWallet},
     PoolHandle, WalletHandle,
 };
@@ -12,30 +15,49 @@ use super::profile::Profile;
 #[derive(Debug)]
 pub struct MixedBreedProfile {
     wallet: Arc<dyn BaseWallet>,
-    ledger: Arc<dyn BaseLedger>,
     anoncreds: Arc<dyn BaseAnonCreds>,
+    anoncreds_ledger_read: Arc<dyn AnoncredsLedgerRead>,
+    anoncreds_ledger_write: Arc<dyn AnoncredsLedgerWrite>,
+    indy_ledger_read: Arc<dyn IndyLedgerRead>,
+    indy_ledger_write: Arc<dyn IndyLedgerWrite>,
 }
 
 impl MixedBreedProfile {
     pub fn new(indy_wallet_handle: WalletHandle, indy_pool_handle: PoolHandle) -> Self {
-        let wallet = Arc::new(IndySdkWallet::new(indy_wallet_handle));
+        let wallet: Arc<dyn BaseWallet> = Arc::new(IndySdkWallet::new(indy_wallet_handle));
+        let anoncreds = Arc::new(IndyCredxAnonCreds::new(Arc::clone(&wallet)));
         let ledger = Arc::new(IndySdkLedger::new(indy_wallet_handle, indy_pool_handle));
-        let anoncreds: Arc<IndyCredxAnonCreds> = Arc::new(IndyCredxAnonCreds::new(wallet.clone()));
+
         MixedBreedProfile {
             wallet,
-            ledger,
             anoncreds,
+            anoncreds_ledger_read: ledger.clone(),
+            anoncreds_ledger_write: ledger.clone(),
+            indy_ledger_read: ledger.clone(),
+            indy_ledger_write: ledger,
         }
     }
 }
 
 impl Profile for MixedBreedProfile {
-    fn inject_ledger(self: Arc<Self>) -> Arc<dyn BaseLedger> {
-        Arc::clone(&self.ledger)
+    fn inject_indy_ledger_read(self: Arc<Self>) -> Arc<dyn IndyLedgerRead> {
+        Arc::clone(&self.indy_ledger_read)
+    }
+
+    fn inject_indy_ledger_write(self: Arc<Self>) -> Arc<dyn IndyLedgerWrite> {
+        Arc::clone(&self.indy_ledger_write)
     }
 
     fn inject_anoncreds(self: Arc<Self>) -> Arc<dyn BaseAnonCreds> {
         Arc::clone(&self.anoncreds)
+    }
+
+    fn inject_anoncreds_ledger_read(self: Arc<Self>) -> Arc<dyn AnoncredsLedgerRead> {
+        Arc::clone(&self.anoncreds_ledger_read)
+    }
+
+    fn inject_anoncreds_ledger_write(self: Arc<Self>) -> Arc<dyn AnoncredsLedgerWrite> {
+        Arc::clone(&self.anoncreds_ledger_write)
     }
 
     fn inject_wallet(&self) -> Arc<dyn BaseWallet> {
--- a/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
+++ b/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
@@ -330,9 +330,14 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
             .add_wallet_record(CATEGORY_CRED_KEY_PROOF, &cred_def_id.0, &str_cred_key_proof, None)
             .await?;
 
-        self.wallet
+        let store_schema_res = self
+            .wallet
             .add_wallet_record(CATEGORY_CRED_SCHEMA, schema.id(), schema_json, None)
-            .await?;
+            .await;
+
+        if let Err(e) = store_schema_res {
+            warn!("Storing schema {schema_json} failed - {e}. It's possible it is already stored.")
+        }
 
         let str_schema_id = serde_json::to_string(schema.id())?;
 
--- a/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
+++ b/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
@@ -943,9 +943,12 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
     }
 
     async fn revoke_credential_local(&self, tails_dir: &str, rev_reg_id: &str, cred_rev_id: &str) -> VcxCoreResult<()> {
-        let cred_rev_id = cred_rev_id
-            .parse()
-            .map_err(|e| AriesVcxCoreError::from_msg(AriesVcxCoreErrorKind::InvalidInput, e))?;
+        let cred_rev_id = cred_rev_id.parse().map_err(|e| {
+            AriesVcxCoreError::from_msg(
+                AriesVcxCoreErrorKind::InvalidInput,
+                format!("Invalid cred_rev_id {cred_rev_id} - {e}"),
+            )
+        })?;
 
         let rev_reg = self.get_wallet_record_value(CATEGORY_REV_REG, rev_reg_id).await?;
 
--- a/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
+++ b/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
@@ -355,8 +355,10 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
             .add_wallet_record(CATEGORY_CRED_SCHEMA, schema.id(), schema_json, None)
             .await?;
 
+        let str_schema_id = serde_json::to_string(schema.id())?;
+
         self.wallet
-            .add_wallet_record(CATEGORY_CRED_SCHEMA_ID, &cred_def_id.0, schema.id(), None)
+            .add_wallet_record(CATEGORY_CRED_SCHEMA_ID, &cred_def_id.0, &str_schema_id, None)
             .await?;
 
         // Return the ID and the cred def
@@ -370,12 +372,12 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
             .get_wallet_record_value(CATEGORY_CRED_KEY_PROOF, cred_def_id)
             .await?;
 
-        let schema_id = match &cred_def {
-            CredentialDefinition::CredentialDefinitionV1(c) => &c.schema_id,
-        };
+        let schema_id = self
+            .get_wallet_record_value(CATEGORY_CRED_SCHEMA_ID, cred_def_id)
+            .await?;
 
         // If cred_def contains schema ID, why take it as an argument here...?
-        let offer = credx::issuer::create_credential_offer(schema_id, &cred_def, &correctness_proof)?;
+        let offer = credx::issuer::create_credential_offer(&schema_id, &cred_def, &correctness_proof)?;
 
         serde_json::to_string(&offer).map_err(From::from)
     }
--- a/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
+++ b/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
@@ -253,7 +253,7 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
             return Ok((rev_reg_id.0, rev_reg, rev_reg_def));
         }
 
-        let (rev_reg_def, rev_reg_def_priv, rev_reg, rev_reg_delta) = credx::issuer::create_revocation_registry(
+        let (rev_reg_def, rev_reg_def_priv, rev_reg, _rev_reg_delta) = credx::issuer::create_revocation_registry(
             &issuer_did,
             &cred_def,
             tag,
@@ -292,12 +292,6 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
 
         self.wallet
             .add_wallet_record(CATEGORY_REV_REG, &rev_reg_id.0, &str_rev_reg, None)
-            .await?;
-
-        let str_rev_reg_delta = serde_json::to_string(&rev_reg_delta)?;
-
-        self.wallet
-            .add_wallet_record(CATEGORY_REV_REG_DELTA, &rev_reg_id.0, &str_rev_reg_delta, None)
             .await?;
 
         Ok((rev_reg_id.0, str_rev_reg_def, str_rev_reg))
@@ -473,9 +467,10 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
         let str_rev_reg = rev_reg.as_ref().map(serde_json::to_string).transpose()?;
         let str_rev_reg_delta = rev_reg_delta.as_ref().map(serde_json::to_string).transpose()?;
 
-        if let (Some(rev_reg_id), Some(str_rev_reg), Some(str_rev_reg_delta), Some((_, _, _, rev_reg_info, _))) =
-            (rev_reg_id, &str_rev_reg, &str_rev_reg_delta, revocation_config_parts)
+        let cred_rev_id = if let (Some(rev_reg_id), Some(str_rev_reg), Some((_, _, _, rev_reg_info, _))) =
+            (rev_reg_id, &str_rev_reg, revocation_config_parts)
         {
+            let cred_rev_id = rev_reg_info.curr_id.to_string();
             let str_rev_reg_info = serde_json::to_string(&rev_reg_info)?;
 
             self.wallet
@@ -483,17 +478,17 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
                 .await?;
 
             self.wallet
-                .update_wallet_record_value(CATEGORY_REV_REG_DELTA, &rev_reg_id, str_rev_reg_delta)
-                .await?;
-
-            self.wallet
                 .update_wallet_record_value(CATEGORY_REV_REG_INFO, &rev_reg_id, &str_rev_reg_info)
                 .await?;
+
+            Some(cred_rev_id)
+        } else {
+            None
         };
 
         let str_cred = serde_json::to_string(&cred)?;
 
-        Ok((str_cred, str_rev_reg, str_rev_reg_delta))
+        Ok((str_cred, cred_rev_id, str_rev_reg_delta))
     }
 
     /// * `requested_credentials_json`: either a credential or self-attested attribute for each requested attribute
@@ -954,8 +949,6 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
 
         let rev_reg = self.get_wallet_record_value(CATEGORY_REV_REG, rev_reg_id).await?;
 
-        let rev_reg_delta = self.get_wallet_record_value(CATEGORY_REV_REG_DELTA, rev_reg_id).await?;
-
         let rev_reg_def = self.get_wallet_record_value(CATEGORY_REV_REG_DEF, rev_reg_id).await?;
 
         let tails_file_hash = match &rev_reg_def {
@@ -968,7 +961,17 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
         let (rev_reg, new_rev_reg_delta) =
             credx::issuer::revoke_credential(&rev_reg_def, &rev_reg, cred_rev_id, &tails_reader)?;
 
-        let rev_reg_delta = credx::issuer::merge_revocation_registry_deltas(&rev_reg_delta, &new_rev_reg_delta)?;
+        let old_str_rev_reg_delta = self.get_rev_reg_delta(rev_reg_id).await?;
+
+        let rev_reg_delta = old_str_rev_reg_delta
+            .as_ref()
+            .map(|s| serde_json::from_str(s))
+            .transpose()?;
+
+        let rev_reg_delta = rev_reg_delta
+            .map(|rev_reg_delta| credx::issuer::merge_revocation_registry_deltas(&rev_reg_delta, &new_rev_reg_delta))
+            .transpose()?
+            .unwrap_or(new_rev_reg_delta);
 
         let str_rev_reg = serde_json::to_string(&rev_reg)?;
         let str_rev_reg_delta = serde_json::to_string(&rev_reg_delta)?;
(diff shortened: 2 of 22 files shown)


aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -211,13 +303,91 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+        if let Ok(cred_def) = self
+            .wallet
+            .get_wallet_record(CATEGORY_CRED_DEF, &cred_def_id.0, &options)
+            .await
+        {
+            return Ok((cred_def_id.0, cred_def));
+        }
+
+        // Otherwise, create cred def
+        let (cred_def, cred_def_priv, cred_key_correctness_proof) =
```
Patrik-Stas, line 335: Just observation for other reviewers and myself

The function here has following signature
```rust
pub fn create_credential_definition(
    origin_did: &DidValue,
    schema: &Schema,
    tag: &str,
    signature_type: SignatureType,
    config: CredentialDefinitionConfig,
) -> Result<(
    CredentialDefinition,
    CredentialDefinitionPrivate,
    CredentialKeyCorrectnessProof,
)> {
```

In vdrtools, there is similar function returning similar stuff:
```rust
async fn _create_credential_definition(
        &self,
        attr_names: &AttributeNames,
        support_revocation: bool,
    ) -> IndyResult<(
        CredentialDefinitionData,
        CredentialPrivateKey,
        CredentialKeyCorrectnessProof,
    )> 
```
and the function is used similarly. But you can note that credx one returns `CredentialDefinition` while vdrtools returns `CredentialDefinitionData`. 

This is what credx structure looks like
```rust
#[derive(Debug)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize), serde(tag = "ver"))]
pub enum CredentialDefinition {
    #[cfg_attr(feature = "serde", serde(rename = "1.0"))]
    CredentialDefinitionV1(CredentialDefinitionV1),
}

pub struct CredentialDefinitionV1 {
    pub id: CredentialDefinitionId,
    pub schema_id: SchemaId,
    #[cfg_attr(feature = "serde", serde(rename = "type"))]
    pub signature_type: SignatureType,
    pub tag: String,
    pub value: CredentialDefinitionData,
}

#[derive(Debug)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
pub struct CredentialDefinitionData {
    pub primary: ursa_cl!(CredentialPrimaryPublicKey),
    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
    pub revocation: Option<ursa_cl!(CredentialRevocationPublicKey)>,
}

pub const CL_SIGNATURE_TYPE: &str = "CL";

#[derive(Copy, Clone, Debug, PartialEq, Eq)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
pub enum SignatureType {
    CL,
}
```

And this is what `vdrtools`'s `CredentialDefinitionData` looks like
```
#[derive(Debug, Serialize, Deserialize)]
pub struct CredentialDefinitionData {
    pub primary: CredentialPrimaryPublicKey,
    pub revocation: Option<CredentialRevocationPublicKey>,
}
``` 
So essentially, `credx` returns richer structure with more metadata, while `vdrtools` returned only the core piece - `CredentialDefinitionData` which is the same between the 2 implementations.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -211,13 +303,91 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+            credx::issuer::create_credential_definition(&issuer_did, &schema, tag, sig_type, config)?;
+
+        let str_cred_def = serde_json::to_string(&cred_def)?;
+
+        // Store stuff in wallet
+        self.wallet
+            .add_wallet_record(CATEGORY_CRED_DEF, &cred_def_id.0, &str_cred_def, None)
```
Patrik-Stas, line 342: as per note above, we not store the data under new custom "type". Every wallet item is identified by tuple "id + type"  pair. Since we store different data, we want to make sure not to mix things up.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
Patrik-Stas: In order to implement migration of old data to new, we will have to map out the `vdrtools`'s `CredentialDefinitionData` to credx's `CredentialDefinition`. 
Since the additional metadata contained in credx datastructure can be known, this should be pretty fairly doable

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -211,13 +303,91 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+        self.wallet
+            .add_wallet_record(CATEGORY_CRED_DEF_PRIV, &cred_def_id.0, &str_cred_def_priv, None)
+            .await?;
+
+        let str_cred_key_proof = serde_json::to_string(&cred_key_correctness_proof)?;
+
+        self.wallet
+            .add_wallet_record(CATEGORY_CRED_KEY_PROOF, &cred_def_id.0, &str_cred_key_proof, None)
+            .await?;
+
```
Patrik-Stas, line 356: in `vdrtools` we also stored one more record

```rust
self._wallet_set_schema_id(wallet_handle, &cred_def_id.0, &schema_id)
            .await?; // TODO: FIXME delete temporary storing of schema id
```
but it was apparently something the original implementors wanted to get rid off. So I suppose everthing works with the credx implementation as is, we can ignore that

Patrik-Stas (commented):
only went to creation of credential definition, looks very promising

left comments, but not really something to address, just observations, feel free to either ignore them or discuss

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: I think it's actually important. It seems to be a mapping between the credential definition ID and the schema ID. Not really sure why, but I'll implement it just so that things are the same. We can evaluate and optimize all this stuff later when we do the refactor.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
Patrik-Stas: I had a type there, now fixed "not -> now"

Anyway, yeah I was just a vocal observation that the wallet interface requires specifying "id" and "type" to store/lookup some record (though vdrtools defines the values of "type" on its own, while here we have to do it ourselves).

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
Patrik-Stas: okay 👌

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -199,8 +212,89 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+            .add_wallet_record(CATEGORY_REV_REG, &rev_reg_id.0, &str_rev_reg, None)
+            .await?;
+
+        let str_rev_reg_delta = serde_json::to_string(&rev_reg_delta)?;
+
+        self.wallet
+            .add_wallet_record(CATEGORY_REV_REG_DELTA, &rev_reg_id.0, &str_rev_reg_delta, None)
+            .await?;
+
+        Ok((rev_reg_id.0, str_rev_reg_def, str_rev_reg))
```
Patrik-Stas, line 297: Obivously the string make it difficult to assure we are returning what calling code expects to be returned in these strings, but I found there's a minor incompatibiltiy on 2nd element `str_rev_reg_def`

While `vdrtools` returned instance of (`libvdrtools/src/domain/anoncreds/revocation_registry_definition.rs`)
```
#[derive(Deserialize, Debug, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct RevocationRegistryDefinitionV1 {
    pub id: RevocationRegistryId,
    pub revoc_def_type: RegistryType,
    pub tag: String,
    pub cred_def_id: CredentialDefinitionId,
    pub value: RevocationRegistryDefinitionValue,
}
```
the credx implementation returns instance of 
```
#[derive(Clone, Debug)]
#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
#[cfg_attr(feature = "serde", serde(tag = "ver"))]
pub enum RevocationRegistryDefinition {
    #[cfg_attr(feature = "serde", serde(rename = "1.0"))]
    RevocationRegistryDefinitionV1(RevocationRegistryDefinitionV1),
}

#[derive(Clone, Debug)]
#[cfg_attr(
    feature = "serde",
    derive(Deserialize, Serialize),
    serde(rename_all = "camelCase")
)]
pub struct RevocationRegistryDefinitionV1 {
    pub id: RevocationRegistryId,
    pub revoc_def_type: RegistryType,
    pub tag: String,
    pub cred_def_id: CredentialDefinitionId,
    pub value: RevocationRegistryDefinitionValue,
}
```

Currently I would expect this to cause failure in `generate_rev_reg` (`aries_vcx/src/common/primitives/revocation_registry.rs`) where we try to deserialize this into custom aries-vcx type `RevocationRegistryDefinition` (`aries_vcx/src/common/primitives/revocation_registry.rs:285`)

Patrik-Stas: The third element of tuple also check out ✅  - produced json looks as expected
```json
{
    "ver": "1.0",
    "value": {
        "accum": "21 129DC07087DB2972AF5B794DF713B445B8C1B0C5BF157C8FF6F0661A0E17ECF10 21 13862D252ED89BD0522029AB666E6B8296457A3322B4730A7CB0AB93BB15EFB75 6 7F9B8103C80DD5A9C5ABD3FE265027D01B8491C4293DE0C0BC8CDF0AF245309D 4 293A8BF9A068F8F6031B626449307EFF8D83523265CDCED77FFA776BB49B0515 6 613376EC16552DD2ED51594C2EBC1A2E1212E770DB57F3F650293A9FD9BAAEBC 4 2B2DD93B763AEA054EC70ECFFB20AA5222177C5FCD7566A5EF62DA82D1433C82"
    }
}
```

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: The trait method now replicates the old behavior.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: Are you sure about this? It looks like they returned `RevocationRegistryDefinition` as well (the parent enum).

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: I'm actually going to remove this commit because it looks like this is wrong. The `_create_credential_definition` function returns `CredentialDefinitionData`, but that data is then used to construct an actual `CredentialDefinition` (parent enum), which is what gets returned.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
Patrik-Stas: yeah you are right, the returns on create revocation registry method lgtm ✅

gmulhearn:
looking good to me so far, pls tag me once Draft is removed

bobozaur:
@gmulhearn Ping

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -679,21 +921,99 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
     async fn clear_rev_reg_delta(&self, rev_reg_id: &str) -> VcxCoreResult<()> {
-        Err(unimplemented_method_err("credx clear_rev_reg_delta"))
+        if self.get_rev_reg_delta(rev_reg_id).await?.is_some() {
+            self.wallet
+                .delete_wallet_record(CATEGORY_REV_REG_DELTA, rev_reg_id)
+                .await?;
+        }
+
+        Ok(())
     }
```
gmulhearn, line 1018: Note, you should be able to remove the `unimplemented_method_err` function and the whole `unit_tests` block from the bottom of the file.

This should be very satisfying to do!

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -679,21 +921,99 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
         let schema_json = serde_json::to_string(&schema)?;
         let schema_id = &schema.id().0;
 
-        // TODO - future - store as cache against issuer_did
+        // TODO - future - store as cache against issuer_did - why? indy_anoncreds.rs doesn't seem to cache it either
```
gmulhearn, line 924: i think i wrote this comment originally. I was most likely using ACA-py as a reference and noticed that they store their `CATEGORY_SCHEMA`/`CATEGORY_CRED_SCHEMA` wallet schema record here. 

Whereas in your implementation, you seem to be storing it in `issuer_create_and_store_credential_def`.

I'm not sure if that is what vdrtools does, but it's a little confusing to me. As an issuer can create many cred defs for the same schema. So wouldn't this throw a duplicate record error if an issuer creates a 2nd cred def for the same schema/schema_id?

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -211,13 +288,75 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+            .await?;
+
+        let str_cred_key_proof = serde_json::to_string(&cred_key_correctness_proof)?;
+
+        self.wallet
+            .add_wallet_record(CATEGORY_CRED_KEY_PROOF, &cred_def_id.0, &str_cred_key_proof, None)
+            .await?;
+
+        self.wallet
+            .add_wallet_record(CATEGORY_CRED_SCHEMA, schema.id(), schema_json, None)
```
gmulhearn, line 332: pls see my comment further down about storing cred_schema here vs at the creation of the schema

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -28,14 +30,34 @@ use credx::{
...
 
-const CATEGORY_CREDENTIAL: &str = "VCX_CREDENTIAL";
 const CATEGORY_LINK_SECRET: &str = "VCX_LINK_SECRET";
 
+const CATEGORY_CREDENTIAL: &str = "VCX_CREDENTIAL";
+const CATEGORY_CRED_DEF: &str = "VCX_CRED_DEF";
+const CATEGORY_CRED_KEY_PROOF: &str = "VCX_CRED_DEF_PROOF";
+const CATEGORY_CRED_DEF_PRIV: &str = "VCX_CRED_DEF_PRIV";
+const CATEGORY_CRED_SCHEMA: &str = "VCX_CRED_SCHEMA";
+const CATEGORY_CRED_SCHEMA_ID: &str = "VCX_CRED_SCHEMA_ID";
```
gmulhearn, line 46: IMO this record type is pretty confusing. It's not intuitive that this is used to map cred_def_ids to schema_ids. I'm hopeful that we can remove  `CATEGORY_CRED_SCHEMA_ID` soon, as it seems unnecessary.

But if we don't want to remove it right now, perhaps you could add a comment to make it more obvious what this record type's purpose is?

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -211,13 +288,75 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+
+        let correctness_proof = self
+            .get_wallet_record_value(CATEGORY_CRED_KEY_PROOF, cred_def_id)
+            .await?;
+
+        let schema_id = self
+            .get_wallet_record_value(CATEGORY_CRED_SCHEMA_ID, cred_def_id)
+            .await?;
+
+        // If cred_def contains schema ID, why take it as an argument here...?
```
gmulhearn, line 356: I agree. Note that ACA-py does not do this weird `CATEGORY_CRED_SCHEMA_ID` record id mapping stuff. They just get the `schema_id` from the cred_def;

https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/indy/credx/issuer.py#L179

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: Also, because this represents a mapping between the cred_def_id and the schema_id, it has to be done when the credential is actually created. Again, this is how libvdrtools does it.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: `libvdrtools` didn't store it when creating the schema, it stored it when creating the credential definition.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: I'll modify the name.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: Done. It was satisfying indeed 😅 !

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
gmulhearn: Yes IMO that is a valid scenario, and should be possible; I tend to do this workflow at work a lot using ACA-py. 

E.g. creating a generic `University_Degree` schema and then creating several cred defs against that schema for different mock universities `ABC_University_Degree`, `XYZ_University_Degree` etc. 

So this line would not allow for that behaviour, as `schema.id()` might already exist in the wallet if already used for a cred def previously;
```
        self.wallet
            .add_wallet_record(CATEGORY_CRED_SCHEMA, schema.id(), schema_json, None)
            .await?;
```

gmulhearn: Had a look in libvdrtools, the difference is that they ignore errors on this schema storage, so they are silencing the duplicate wallet record error;

```
        let _ = self
            .wallet_service
            .add_indy_object(wallet_handle, &schema_id.0, &schema_, &HashMap::new())
            .await
            .ok();
```

note `.ok` rather than `?`

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
gmulhearn: kk fair enough. i'll resolve this thread but pls see the other thread about silencing duplicate records for schema id

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
gmulhearn: kk fair enough

gmulhearn (dismissed):
seeing ModularLibsProfile complete brings me joy! Nice work

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -679,21 +921,99 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+        let res_rev_reg_delta = res_rev_reg_delta.ok().as_ref().map(serde_json::to_string).transpose();
+
+        if let Err(err) = &res_rev_reg_delta {
+            warn!(
+                "get_rev_reg_delta >> Unable to deserialize rev_reg_delta cache for rev_reg_id: {}, error: {}",
+                rev_reg_id, err
+            );
+        }
+
+        Ok(res_rev_reg_delta.ok().flatten())
```
mirgee, line 1012: Can't the same be done using a single `match` statement? Perhaps it would be more indented, yes, but more readable still?

aries_vcx/src/core/profile/mixed_breed_profile.rs
```diff
@@ -0,0 +1,44 @@
...
+pub struct MixedBreedProfile {
+    wallet: Arc<dyn BaseWallet>,
+    ledger: Arc<dyn BaseLedger>,
+    anoncreds: Arc<dyn BaseAnonCreds>,
+}
+
+impl MixedBreedProfile {
+    pub fn new(indy_wallet_handle: WalletHandle, indy_pool_handle: PoolHandle) -> Self {
+        let wallet = Arc::new(IndySdkWallet::new(indy_wallet_handle));
+        let ledger = Arc::new(IndySdkLedger::new(indy_wallet_handle, indy_pool_handle));
```
mirgee, line 22: Why create a new profile? The proxy profile could have been used with `IndySdkAnoncreds` swapped for `IndyCredxAnonCreds`.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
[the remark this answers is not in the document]
bobozaur: This whole thing is very bad, from a performance, readability, code quality point of view. Your suggestion is good, but I'm honestly not concerned with this right now as this is just a stepping stone towards `anoncreds-rs`, strong typing and splitting the trait and all. This whole thing will be removed soon. So if you don't mind too much I will discard the readability comments for this PR.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -199,8 +220,64 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+
+        let rev_reg_id =
+            credx::issuer::make_revocation_registry_id(&issuer_did, &cred_def, tag, RegistryType::CL_ACCUM)?;
+
+        let res_rev_reg = self.get_wallet_record_value(CATEGORY_REV_REG, &rev_reg_id.0).await;
+        let res_rev_reg_def = self.get_wallet_record_value(CATEGORY_REV_REG_DEF, &rev_reg_id.0).await;
+
+        if let (Ok(rev_reg), Ok(rev_reg_def)) = (res_rev_reg, res_rev_reg_def) {
+            return Ok((rev_reg_id.0, rev_reg, rev_reg_def));
+        }
```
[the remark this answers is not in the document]
bobozaur: I disagree. Long lines with function invocations don't improve readability to me. I much prefer intermediate variables and more but shorter lines.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -199,8 +220,64 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+            id: rev_reg_id.clone(),
+            curr_id: 0,
+            used_ids: HashSet::new(),
+        };
+
+        let str_rev_reg_info = serde_json::to_string(&rev_reg_info)?;
+
+        self.wallet
+            .add_wallet_record(CATEGORY_REV_REG_INFO, &rev_reg_id.0, &str_rev_reg_info, None)
+            .await?;
```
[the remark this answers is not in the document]
bobozaur: One day it will, when all of this is done properly.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -199,8 +220,64 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+
+        let (rev_reg_def, rev_reg_def_priv, rev_reg, _rev_reg_delta) = credx::issuer::create_revocation_registry(
+            &issuer_did,
+            &cred_def,
+            tag,
+            RegistryType::CL_ACCUM,
+            IssuanceType::ISSUANCE_BY_DEFAULT,
+            max_creds,
+            &mut tails_writer,
+        )?;
```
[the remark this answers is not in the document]
bobozaur: Hm, fair. I'll keep this in mind when refactoring this stuff.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -28,14 +30,34 @@ use credx::{
...
+const CATEGORY_REV_REG_DELTA: &str = "VCX_REV_REG_DELTA";
+const CATEGORY_REV_REG_INFO: &str = "VCX_REV_REG_INFO";
+const CATEGORY_REV_REG_DEF: &str = "VCX_REV_REG_DEF";
+const CATEGORY_REV_REG_DEF_PRIV: &str = "VCX_REV_REG_DEF_PRIV";
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct RevocationRegistryInfo {
+    pub id: RevocationRegistryId,
+    pub curr_id: u32,
+    pub used_ids: HashSet<u32>,
```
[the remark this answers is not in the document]
bobozaur: I just copy pasted the info from `libvdrtools`. But yeah, these are better. However, this revocation info stuff is just a very bad way of tracking the current revocation ID. Right now we just replicate what `libvdrtools` was doing, but this will not make it in the refactor.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -211,13 +288,75 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+            .add_wallet_record(CATEGORY_CRED_SCHEMA, schema.id(), schema_json, None)
+            .await?;
+
+        let str_schema_id = serde_json::to_string(schema.id())?;
+
+        self.wallet
+            .add_wallet_record(CATEGORY_CRED_SCHEMA_ID, &cred_def_id.0, &str_schema_id, None)
+            .await?;
+
+        // Return the ID and the cred def
```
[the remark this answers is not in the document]
bobozaur: These were just for me, as this was one of the first things to implement and it's easier for me to follow what's going on, especially with all the loose typing. It will get removed in a future iteration.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -228,8 +367,111 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+
+                let rev_reg = self.get_wallet_record_value(CATEGORY_REV_REG, rev_reg_id).await?;
+                let rev_reg_info: RevocationRegistryInfo =
+                    self.get_wallet_record_value(CATEGORY_REV_REG_INFO, rev_reg_id).await?;
+
+                Some((rev_reg_def, rev_reg_def_priv, rev_reg, rev_reg_info, tails_dir))
+            }
+            (None, None) => None,
+            (tails_dir, rev_reg_id) => {
+                warn!("Missing revocation config params: tails_dir: {tails_dir:?} - {rev_reg_id:?}; Issuing non revokable credential");
```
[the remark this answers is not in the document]
bobozaur: This was how `libvdrtools` was doing it. I really only care about replicating that in credx right now, and improve on all this mess later.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -228,8 +367,111 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+                let tails_file_hash = match rev_reg_def {
+                    RevocationRegistryDefinition::RevocationRegistryDefinitionV1(rev_reg_def) => {
+                        if rev_reg_info.curr_id > rev_reg_def.value.max_cred_num {
+                            return Err(AriesVcxCoreError::from_msg(
+                                AriesVcxCoreErrorKind::ActionNotSupported,
+                                "RevocationRegistryAccumulator is full",
+                            ));
+                        }
+
+                        if rev_reg_def.value.issuance_type == IssuanceType::ISSUANCE_ON_DEMAND {
```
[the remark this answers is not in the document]
bobozaur: Yeah, agreed. But right now this is what `libvdrtools` was doing.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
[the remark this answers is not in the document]
mirgee: Almost all code will change one day, the question is how soon and I don't think we know for sure. But it was just a suggestion.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -228,8 +367,111 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+                            rev_reg_info.used_ids.insert(rev_reg_info.curr_id);
+                        }
+
+                        &rev_reg_def.value.tails_hash
+                    }
+                };
+
+                let tails_file_path = format!("{}/{}", tails_dir, tails_file_hash);
```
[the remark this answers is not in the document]
bobozaur: Okay, sure. This was a last minute addition because `credx`, unlike `libvdrtools`, does not create the tails dir if it doesn't exist.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -211,13 +288,75 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+            .get_wallet_record_value(CATEGORY_CRED_SCHEMA_ID, cred_def_id)
+            .await?;
+
+        // If cred_def contains schema ID, why take it as an argument here...?
+        let offer = credx::issuer::create_credential_offer(&schema_id, &cred_def, &correctness_proof)?;
+
+        serde_json::to_string(&offer).map_err(From::from)
     }
 
     async fn issuer_create_credential(
```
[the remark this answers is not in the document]
bobozaur: Or deleted entirely and rewritten from scratch, I agree.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: Probably could. Not going to wonder about that now tbh, but I definitely will in the future.

aries_vcx/src/core/profile/mixed_breed_profile.rs
bobozaur: Because @Patrik-Stas said he'd like to have cross-libraries testing and it would be clear to remove this if it is it's own individual type rather than tweaking some other type. And I really wanted to use that name somewhere 🤣 . But it will go away soon.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -228,8 +367,111 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+        let revocation_config = match &mut revocation_config_parts {
+            Some((rev_reg_def, rev_reg_def_priv, rev_reg, rev_reg_info, tails_dir)) => {
+                rev_reg_info.curr_id += 1;
+
+                let tails_file_hash = match rev_reg_def {
+                    RevocationRegistryDefinition::RevocationRegistryDefinitionV1(rev_reg_def) => {
+                        if rev_reg_info.curr_id > rev_reg_def.value.max_cred_num {
+                            return Err(AriesVcxCoreError::from_msg(
+                                AriesVcxCoreErrorKind::ActionNotSupported,
+                                "RevocationRegistryAccumulator is full",
```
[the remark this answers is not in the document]
bobozaur: Changed.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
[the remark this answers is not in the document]
bobozaur: Changed.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -679,21 +921,99 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+
+        let rev_reg = self.get_wallet_record_value(CATEGORY_REV_REG, rev_reg_id).await?;
+
+        let rev_reg_def = self.get_wallet_record_value(CATEGORY_REV_REG_DEF, rev_reg_id).await?;
+
+        let tails_file_hash = match &rev_reg_def {
+            RevocationRegistryDefinition::RevocationRegistryDefinitionV1(r) => &r.value.tails_hash,
+        };
+
+        let tails_file_path = format!("{tails_dir}/{tails_file_hash}");
```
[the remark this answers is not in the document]
bobozaur: Changed.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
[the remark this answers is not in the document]
bobozaur: Once this is merged, the migration from libvdrtools -> credx is up. Once that's done, using `anoncreds-rs` instead of `credx` is up. Then this whole mess will be properly done. I'd personally like to avoid investing more time into improving the quality of some code that will be removed soon and it was a mess to begin with.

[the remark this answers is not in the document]
bobozaur: Maybe also worth noting in terms of why I think like this is that, while it was good exercise for me to get familiar with this stuff, I also would've personally not bother with doing this in the first place and moved straight to `anoncreds-rs` and some high quality implementation. 

But the decision was to take it one step at a time. So I'm just aiming for some common ground of having this implemented, be it even poorly, and then merely use it as a loose reference point on the next implementation.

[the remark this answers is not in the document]
mirgee: OK, so then next time please say so in the PR description. I don't see the point in reviewing the code carefully if the code is not "for real" (this time) and comments are to be automatically discarded.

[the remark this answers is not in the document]
bobozaur: My bad here, sorry. We've been talking about these steps in the weekly calls for weeks and I was under the impression that everyone was aware of the plan. I'll be more careful with the descriptions from now.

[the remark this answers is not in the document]
mirgee: I knew this implementation is to be replaced with `anoncreds-rs`. I did not know that it means that this code should be  "very bad, from a performance, readability, code quality point of view" and "whole mess" and that comments will be automatically discarded.

aries_vcx/src/core/profile/mixed_breed_profile.rs
Patrik-Stas: I wasn't thinking in particular of trying this combination, why not. It's good to know this also works :-) 
I agree we can remove it in some subsequent PR.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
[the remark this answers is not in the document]
Patrik-Stas: I agree this would make migration a bit more difficult than it needs to be, as these structs are and were stored in wallet. if anything, we can do renaming in separate step.

Patrik-Stas (dismissed):
i tried hard to find any significant deviations compared to original code, but couldn't find any.

LGTM

Patrik-Stas:
@bobozaur I guess one more rebase needed, plus consider the one comment I've left

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -28,14 +30,36 @@ use credx::{
...
 use uuid::Uuid;
 
 use super::base_anoncreds::BaseAnonCreds;
 
-const CATEGORY_CREDENTIAL: &str = "VCX_CREDENTIAL";
 const CATEGORY_LINK_SECRET: &str = "VCX_LINK_SECRET";
 
+const CATEGORY_CREDENTIAL: &str = "VCX_CREDENTIAL";
+const CATEGORY_CRED_DEF: &str = "VCX_CRED_DEF";
+const CATEGORY_CRED_KEY_PROOF: &str = "VCX_CRED_DEF_PROOF";
```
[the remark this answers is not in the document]
bobozaur: Done.

Patrik-Stas (dismissed):

mirgee (approved):

PR #861, opened 2023-05-24
Split BaseLedger trait

mirgee:
Splits `BaseLedger` trait into smaller ones:
* `AnoncredsLedgerRead` - methods for resolving Anoncreds primitives
* `AnoncredsLedgerWrite` - methods for registering Anoncreds primitives
* `IndyLedgerRead` - methods for reading Indy-specific transactions from the ledger
* `IndyLedgerWrite` - methods for writing Indy-specific transactions to the Ledger

Among other things, this allows eventual elimination of TAA stored in global state by building the "reader" which allows to obtain the TAA from ledger and use said TAA to construct a "writer" which needs TAA to write to the ledger.

The associated methods `set_endorser` and `endorse_transaction` may stand out:
* The method `set_endorser`, temporarily associated with `IndyLedgerRead`, does not read from the ledger, but merely modifies a transaction.
* The method `endorse_transaction`, temporarily associated with `IndyLedgerWrite`, again only manipulates an existing transaction (adds a multisignature) before submitting.
The pattern with the remaining methods is that a ledger-specific request is built from scratch using the supplied data and submitted to the ledger. Perhaps therefore, local modification of existing requests should not be the responsibility of mentioned traits.

### DIFF
--- a/agents/rust/aries-vcx-agent/src/services/schema.rs
+++ b/agents/rust/aries-vcx-agent/src/services/schema.rs
@@ -34,7 +34,7 @@ impl ServiceSchemas {
     }
 
     pub async fn schema_json(&self, thread_id: &str) -> AgentResult<String> {
-        let ledger = Arc::clone(&self.profile).inject_ledger();
+        let ledger = Arc::clone(&self.profile).inject_anoncreds_ledger_read();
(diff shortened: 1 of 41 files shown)


Patrik-Stas (approved):
love to see it

this is great

PR #862, opened 2023-05-24
Split IndyVdrLedger and IndySdkLedger

mirgee:
Splits 
* `IndyVdrLedger` -> `IndyVdrLedgerRead` & `IndyVdrLedgerWrite`, and
* `IndySdkLedger` -> `IndySdkLedgerRead` & `IndySdkLedgerWrite`

This allows each implementation to use only what it needs (e.g. `IndyVdrLedgerRead` variant doesn't need `RequestSigner`, `IndyVdrLedgerWrite` doesn't need `ResponseCacher` and `ResponseParser`). Moreover, since both are initialized separately, a `IndyLedgerRead` implementation can be used to initialize a `IndyLedgerWrite` implementation.

### DIFF
--- a/aries_vcx_core/src/ledger/indy_ledger.rs
+++ b/aries_vcx_core/src/ledger/indy_ledger.rs
@@ -6,14 +6,29 @@ use crate::{indy, PoolHandle, WalletHandle};
 use super::base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerWrite, IndyLedgerRead, IndyLedgerWrite};
 
 #[derive(Debug)]
-pub struct IndySdkLedger {
+pub struct IndySdkLedgerRead {
     indy_wallet_handle: WalletHandle,
     indy_pool_handle: PoolHandle,
 }
 
-impl IndySdkLedger {
+impl IndySdkLedgerRead {
     pub fn new(indy_wallet_handle: WalletHandle, indy_pool_handle: PoolHandle) -> Self {
-        IndySdkLedger {
+        IndySdkLedgerRead {
+            indy_wallet_handle,
+            indy_pool_handle,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct IndySdkLedgerWrite {
+    indy_wallet_handle: WalletHandle,
+    indy_pool_handle: PoolHandle,
+}
+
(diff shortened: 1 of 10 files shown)


aries_vcx_core/src/ledger/indy_ledger.rs
```diff
@@ -6,22 +6,37 @@ use crate::{indy, PoolHandle, WalletHandle};
 use super::base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerWrite, IndyLedgerRead, IndyLedgerWrite};
 
 #[derive(Debug)]
-pub struct IndySdkLedger {
+pub struct IndySdkLedgerRead {
     indy_wallet_handle: WalletHandle,
```
Patrik-Stas, line 10: perhaps can be removed now?

Patrik-Stas (dismissed):

aries_vcx_core/src/ledger/indy_ledger.rs
bobozaur: Maybe not worth it doing it here, but how about taking a generic implementing `BaseWallet` along with the `submitter_did`? Something like `Option(&str, &T)` where `T: BaseWallet`. Shouldn't that remove the need to store the wallet handle while also making it easier to work with that scenario since the arguments are either provided together or not at all?

bobozaur (commented):
I merely have an observation. Otherwise LGTM.

did_resolver_sov/src/reader/vdr_reader.rs
```diff
@@ -28,13 +28,13 @@ impl TryFrom<LedgerPoolConfig> for ConcreteAttrReader {
             .capacity(1000)?
             .build();
         let response_cacher = Arc::new(InMemoryResponseCacher::new(cacher_config));
-        let config = IndyVdrLedgerConfig {
+        let config = IndyVdrLedgerReadConfig {
             request_signer,
```
bobozaur, line 32: I tried pulling the branch locally and I'm getting an error from cargo check that there's no `request_signer` in the struct. How did the CI pass though? Am I doing something wrong? As it does seem that the field was indeed removed from the struct.

Patrik-Stas (approved):

PR #863, opened 2023-05-24
Remove reliance of IndyVdrLedger on global state

mirgee:
Makes `ProtocolVersion` and `TxnAuthrAgrmtAcceptanceData` injectable to `IndyVdrLedgerRead` and `IndyVdrLedgerWrite` on construction, thus removing their reliance on global `settings`. Also adds TAA ledger response parsing.


aries_vcx_core/src/ledger/indy_vdr_ledger.rs
```diff
@@ -11,22 +11,22 @@ use vdr::ledger::requests::schema::{Schema, SchemaV1};
...
-use crate::global::settings;
+use crate::errors::error::{AriesVcxCoreError, AriesVcxCoreErrorKind, VcxCoreResult};
 
 use super::base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerWrite, IndyLedgerRead, IndyLedgerWrite};
 use super::request_signer::RequestSigner;
 use super::request_submitter::RequestSubmitter;
 use super::response_cacher::ResponseCacher;
 
+// TODO: Should implement builders for these configs...
+// Good first issue?
```
Patrik-Stas, line 29: I am curious what @bobozaur @gmulhearn you guys think about `IndyVdrLedgerReadConfig`, `IndyVdrLedgerWriteConfig` in general.

aries_vcx_core/src/ledger/indy_vdr_ledger.rs
```diff
@@ -162,11 +167,14 @@ where
...
     }
 
-    async fn get_txn_author_agreement(&self) -> VcxCoreResult<String> {
+    async fn get_txn_author_agreement(&self) -> VcxCoreResult<Option<String>> {
         let request = self
             .request_builder()?
             .build_get_txn_author_agreement_request(None, None)?;
-        self.request_submitter.submit(request).await
+        let response = self.request_submitter.submit(request).await?;
+        not_found_to_none(self.response_parser.parse_get_txn_author_agreement_response(&response))?
```
Patrik-Stas, line 175: small naming suggestion, I'd go for `map_error_not_found_to_none`

aries_vcx_core/src/ledger/indy_vdr_ledger.rs
Patrik-Stas, line 1: It seems that `publish_nym` does not use `append_txn_author_agreement_to_request` but should be

Patrik-Stas (changes requested):
left minor comments, but generally lgtm ✅ 👍

aries_vcx_core/src/ledger/indy_vdr_ledger.rs
mirgee: I am a little confused because I am not entirely sure what exactly is being asked by either of you. The context here is that there was a single small config struct here in the beginning. It was split into two and multiple fields were added. As their complexity grows, I just wanted to simplify their construction and so I wrote this comment. This was just the most simple and straightforward thing to do.

I have no issue with doing a builder pattern for the target struct, it just seemed like an overkill in the beginning.

I have a feeling that there is a miscommunication happening here and @Patrik-Stas incorrectly assumes my position is always and in all cases of a config struct aficionado, whereas I only said a few times that I prefer a single config object as an argument instead of a separate argument for each field resulting in functions taking zillion arguments :)

bobozaur: Fair enough. I just think we should take a note of this and maybe create an issue when this is merged. Whether you wanna do it now or in a later refinement, I think they would be both fine. Replacing the config structs with builders would also be a good first issue, I think.

Patrik-Stas: it's not a biggie, just wanted to know other opinions, as we've discussed the similar things in past few times.

Also the term "a config struct aficionado" really got me lolled :-D

Patrik-Stas (changes requested):
just needs a rebase

Patrik-Stas (changes requested):

PR #864, opened 2023-05-30
draft: DidDocument integration

mirgee:
Replaces the use of
* `diddoc_legacy::aries::diddoc::AriesDidDoc` with `did_doc::schema::did_doc::DidDocument`, and
* `diddoc_legacy::aries::service::AriesService` with `did_doc::schema::service::Service`
across the aries-vcx codebase.

Integration with mediated connection is skipped and `diddoc_legacy` will be removed along with `mediated_connection`.

The method specific fields of the service struct are now set via a type associated with the `DidResolvable` trait, where only fields specific to the sovrin method are used in the codebase for now. This choice impacts the current implementation of the resolver registry, which needs to be reconsidered.

Moreover, currently, the service type is set manually during the construction of the `DidDocument`, and the `accept` field is not set at all. However, as per the [Sovrin DID method specification](https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html#crud-operation-definitions), the value of those two fields is implied by the composition of / fields included in the service. This suggests that potentially we might want to wrap the `DidDocumentBuilder` in some kind of `DidDocumentBuilderSov` responsible for this method-specific logic in the future.

This would be also useful because the extra fields are defined in `did_resolver_sov`. In some use-cases, the entire resolver is imported only for the extra fields needed in order to name the `DidDocument` or `Service` type.

The DID parser fails to parse any did which is not fully qualified. This is desirable in cases where fully qualified DID is required as it forces early failure. But in situations where the method is implied, an unqualified DID may also be valid and usable. Therefore, perhaps a separate type should be created to be used in these cases.

As it stands, this integration is a breaking change as it changes the format of the `did_doc` field in the serialized connection state machines in both non-backwards-compatible and non-forwards-compatible way (e.g. some fields optional in the legacy format are now required and `publicKey` used in the legacy format is not a valid DDO field - it can probably be mapped to another valid field when deserializing, but currently isn't).


wrappers/node/test/helpers/mockdata.ts
```diff
@@ -62,7 +63,7 @@ export const ARIES_CONNECTION_REQUEST = {
         {
           id: 'did:example:123456789abcdefghi;indy',
           priority: 0,
-          recipientKeys: ['2RjtVytftf9Psbh3E8jqyq#1'],
```
mirgee, line 65: Currently, after converting from the new to legacy DDO format, resolving key by reference fails.

mirgee:
Postponed until
* mediated connection is removed,
* libvdrtools is removed,
* `DidExchange` protocol is implemented,
* Sovrin-specific wrapper for `DidDocument` is implemented (incl. dereferencing), and potentially
* fully qualified DIDs are used across the codebase in backwards-compatible manner.

Issue #822, opened 2023-05-02
Use static JSON in messages crate tests


bobozaur:
Some corners were cut when developing the tests for the `messages` crate during its refactor in the sense that the JSON was being composed off of already tested structs or fields.

It would be better to use plain text JSON to ensure that any API change won't slip by the testing suite.

SumantxD:
@bobozaur i would like to work on this issue can you help me navigate it in the project directory 
Thank You.

bobozaur:
> @bobozaur i would like to work on this issue can you help me navigate it in the project directory Thank You.

Not sure what you mean. This is about most tests in `msg_fields` module (might be others in other modules). 

The ones about `msg_type` deserialization are fine, but the tests that are about deserializing an actual message with fields should ideally use static messages instead of serializing certain types to avoid the situation where both the serialization and the deserialization are incorrect and thus the test passes but the result is incorrect.

tech-bash:
I would like to try this out.

bobozaur:
> I would like to try this out.

Have at it :).

tech-bash:
In the above example you have set `~thread` to two same names i.e `field1` so is there any specific reason behind it or as I am thinking that, the other value could be `field2` as well!

bobozaur:
> In the above example you have set `~thread` to two same names i.e `field1` so is there any specific reason behind it or as I am thinking that, the other value could be `field2` as well!

Ah, my bad. Edited. They are supposed to be different fields, yes.

PR #866, opened 2023-06-01
Refactor messages crate tests

tech-bash:
fix: #822 

Just want to know that the direction in which I'm going is upright! For using the `Static JSON` in message crate tests


Patrik-Stas:
closing per inactivity

PR #867, opened 2023-06-01
Feature/cred migrator

bobozaur:
This PR introduces the `wallet_migrator` library crate, currently just for the purpose of migrating anoncreds related objects from their `libvdrtools` implementation to `indy-credx`. This is accomplished by creating a new `Indy` wallet, so the old one remains as a backup.  

The migrator is flexible enough though to support migrating anything in terms of an `Indy` wallet.

In the future the crate could be extended to assist with migrating from an `Indy` wallet to a different one too.

The PR also adds a testing feature (unfortunately necessary) to accommodate current tests in `aries-vcx` so that wallets get migrated in the middle of the tests to ensure things keep working post migration.

Stubs are already in place for the `indy-credx` to `anoncreds-rs` migration, but that will be fully implemented in a future PR.


aries_vcx/src/core/profile/profile.rs
```diff
@@ -6,6 +6,7 @@ use aries_vcx_core::{
     anoncreds::base_anoncreds::BaseAnonCreds,
     ledger::base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerWrite, IndyLedgerRead, IndyLedgerWrite},
     wallet::base_wallet::BaseWallet,
+    WalletHandle,
```
gmulhearn, line 9: hmmm, not sure about `Profile` re-exporting a type from indy/vdrtools. It implies that consumers of a Profile now have to import the whole vdrtools crate. 

The CI/CD "check variants" step should stop this. I wonder if you could just have `wallet_handle(..)` return `i32?` so that it's not exposing vdrtools

bobozaur: Well, it did 😅 . I don't really like having to export this either, but:
1) This whole profile thing will go away in the near future.
2) It's a necessary evil for testing because the handle is needed for migration, no two ways about it.

I will accommodate it differently, though.

bobozaur: I put the whole method behind the `migration` feature flag, as that's really the only place it is needed (for testing).

bobozaur: Also, while we could have the method export the inner type, since the scope is limited to testing I think it's better to just have it behind the feature flag. I just really forgot it was there as I did that a couple of weeks ago 😅.

gmulhearn: kk sounds good.

good news, the check step was failing i believe; https://github.com/hyperledger/aries-vcx/actions/runs/5344789783/jobs/9689606086?pr=867#step:5:1246.

IMO it's v important that we continue to allow consumers to import aries_vcx without any vdrtools deps (or without any credx/modular-libs deps if they prefer, etc)

bobozaur: Yeah, it did. God bless CI 😅 .
I agree with you regarding the deps. It was really just a snafu on my part as it was a quick addition when I started this since the wallet handle of the current `Indy` wallet was needed.

Patrik-Stas (changes requested):
left few comments

I'd suggest to work a little bit more on testing - the current approach includes migration to existing tests, but usually the migration happens at the same spot, after issuing credentials.
We could probably take some tests and tweak around the position of migration, but I think following scenario could deserve a dedicated test:

Scenario 1:
- create creddef, revocation registry
- issue credential A, revoke locally
- issue credential B
- migrate
- revoke locally B
- issue another credential C
- revoke locally C
- publish revocation

Mainly to check that revocation registries and accumulated deltas properly survive across the migration

libvdrtools/indy-wallet/src/lib.rs
```diff
@@ -704,6 +704,61 @@ impl WalletService {
         Ok(())
     }
 
+    pub async fn migrate_records<E>(
```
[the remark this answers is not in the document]
bobozaur: Most of the internal vdrtools wallet API is not public (I assume by design, so that weird things cannot be done from outside). So to have this someplace else means exposing a lot of the wallet API, which I'm not necessarily against but don't find as a good idea either.

There's no reason this wouldn't work for credx -> anoncreds-rs. This is the wallet API, not the anoncreds API.

libvdrtools/indy-wallet/src/wallet.rs
```diff
@@ -352,7 +353,14 @@ impl Wallet {
                 None => {
                     // no item in cache, lets retrieve it and put it in cache.
                     let metrics_fut = cache_hit_metrics.inc_cache_miss(type_);
-                    let full_options = RecordOptions::id_value_tags();
+                    let full_options = RecordOptions {
```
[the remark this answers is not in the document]
bobozaur: Honestly, I have no idea whether setting this to `true` all the time is a good idea and I think finding out is outside the scope of this PR. 

I changed this simply because there was no way to set it to `true`, which was bad and was needed for the migration, but if the original record options don't set it to `true`, then it remains `false` and everything happens as before.

Furthermore, the scenario you described will never happen, because when the item is added to the cache, it is always passed the `type_`.

wallet_migrator/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+        ($wh:expr, $category:expr, $res:ty) => {{
+            let val = get_wallet_item_raw($wh, $category).await;
+            serde_json::from_str::<$res>(&val).unwrap()
+        }};
+    }
+
+    #[tokio::test]
+    async fn test_migration() {
+        let wallet_key = "8dvfYSt5d1taSd6yJdpjq4emkwsPDDLYxkNFysFD2cZY".to_owned();
+        let (credentials, config) = make_wallet_reqs(wallet_key.clone(), "wallet_test_migration".to_owned());
```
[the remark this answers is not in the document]
bobozaur: Yeah

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
bobozaur: I think I can arrange something, but I'd really do something more like the dummy data structures used in the test in this file (I guess you had the same thing in mind).

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -344,10 +344,8 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
             warn!("Storing schema {schema_json} failed - {e}. It's possible it is already stored.")
         }
 
-        let str_schema_id = serde_json::to_string(schema.id())?;
-
         self.wallet
-            .add_wallet_record(CATEGORY_CRED_MAP_SCHEMA_ID, &cred_def_id.0, &str_schema_id, None)
+            .add_wallet_record(CATEGORY_CRED_MAP_SCHEMA_ID, &cred_def_id.0, &schema.id().0, None)
```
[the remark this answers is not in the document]
bobozaur: Yeah, it was a weird one that took a shameful amount of time to troubleshoot. I messed up the implementation initially and stored the serialized `SchemaId`, whereas the `libvdrtools` impl stored the inner string itself. 

I changed this so we'd be in sync with libvdrtools, but this is really going to go away in the `anoncreds-rs` impl as we can retrieve this schema ID from the credential definition, so we don't store all this stuff all over again and have some different category in place just for it.

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
bobozaur: Added some test to use the `mysql` wallet too. Note that to avoid having to duplicate code all over the place I moved the mysql prepation (the migration file contents) and the database creation inside `libvdrtools`. They do the database setup for `sqlite`, so I think they should be doing it for `mysql` as well. 

This allows for easier testing/setup of mysql wallets since we don't have to do anything from outside libvdrtools.

bobozaur:
@Patrik-Stas The requested test scenario was added. I'll be monitoring the CI to check on the latest changes, but apart from that I think the points you mentioned in your review are covered now.

wallet_migrator/src/vdrtools2credx/conv.rs
Patrik-Stas, line 1: I think we should add some debug logs to each migration step. It might be desirable to have logs about what and how many records have been migrated.

wallet_migrator/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+    let new_wh = locator
+        .wallet_controller
+        .open(config.clone(), credentials.clone())
+        .await?;
+
+    debug!("Migrating records from wallet with handle {wallet_handle:?} to wallet with handle {new_wh:?}");
+
+    let res = locator
+        .wallet_controller
+        .migrate_records(wallet_handle, new_wh, migrate_fn)
```
Patrik-Stas, line 48: Would be nice to have dry run option. If I run this on server with many records, it'd be nice to first just check the migration can succeed given the current wallet state, avoiding any risks/possible issue related to writing to db.

wallet_migrator/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+        .await;
+
+    locator.wallet_controller.close(new_wh).await?;
```
Patrik-Stas, line 51: This can potentially fail and then the following wallet deletion won't be run. 

I would rather:
- have dry run option I can use first
- on failure not delete at all (to avoid this issue with skipping deletion if closing fails - just make it users responsibility to manage wallets they might have created while using this)
- make migration idempotent, so I could migrate to same target, even if I initially get failure on attempt to migrate some wallet item

Patrik-Stas (changes requested):
left some comments

wallet_migrator/src/lib.rs
Patrik-Stas: It would be cool if we could also count number of records as part of the dry run (or perhaps regardless)

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -383,6 +384,7 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
         let cred_request = serde_json::from_str(cred_req_json)?;
         let cred_values = serde_json::from_str(cred_values_json)?;
 
+        // TODO: Might need to qualify with offer method or something - look into how vdrtools does it
```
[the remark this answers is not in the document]
bobozaur: It still is a TODO, yeah. I left it because I noticed this now and I don't really know how to approach this whole qualification thing. I was thinking of tackling this in a separate PR, maybe directly in the `anoncreds-rs` implementation.

libvdrtools/indy-wallet/src/storage/mysql/mod.rs
```diff
@@ -811,12 +811,87 @@ impl WalletStorageType for MySqlStorageType {
...
+
+        // Replace the previous single use pool
+        // with the actual one, get a connection
+        // and create the required tables
+        pool = self._connect(false, config, credentials).await?;
+        con = pool.acquire().await?;
+
+        sqlx::query(
+            r#"
+            CREATE TABLE IF NOT EXISTS `items` (
```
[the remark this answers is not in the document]
bobozaur: The `id` is mandatory, yeah. I don't really know how it worked without it there, but it complained until I added it. It is defined for the SQLite storage, which is what made me think it should exist. Having it autoincrement means it must be the `PRIMARY KEY` too. But I'll add a `UNIQUE` constraint on the previous `PRIMARY KEY`

wallet_migrator/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+    migrate_fn: impl FnMut(Record) -> Result<Option<Record>, E>,
+) -> MigrationResult<()>
+where
+    E: Display,
+{
+    info!("Starting wallet migration...");
+    let locator = Locator::instance();
+
+    debug!("Creating new wallet {} ...", config.id);
+    locator
```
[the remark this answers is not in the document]
bobozaur: Can do. I just thought that taking care of the wallet creation would mean we *always* create a new one and *ensure* we don't mess up the old one.

How is the users thing a problem, though? You have to pass the `Config` and the `Credentials` to construct the new wallet, which can be virtually anything.

wallet_migrator/src/vdrtools2credx/conv.rs
bobozaur: Can do.

wallet_migrator/src/lib.rs
bobozaur: Hmmmmmm... It would be a bit tricky to implement as part of the library because it would have to mess up with all the internals of the migration code.

Would an "ephemeral" SQLite wallet do the trick? I'd personally just migrate (and create a new wallet) and ensure to delete it after (maybe use some UUID as part of the name for uniqueness and to avoid concurrency issues).

I could add some helper function for this but idk, it feels like something too small and specific to be worth it. What do you think?

libvdrtools/indy-wallet/src/storage/mysql/mod.rs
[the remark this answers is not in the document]
bobozaur: On a second thought, the `id` is mandatory because the query for getting all records has an `ORDER BY id` clause. It's not used apart from that. Since this was not being used in any of the code so far, the problem never surfaced.

The ordering is really not important, so instead of doing the unique key shenanigan I modified the `SELECT` to not order the records anymore "¯\_(ツ)_/¯ ".

wallet_migrator/src/lib.rs
bobozaur: Looking at the wallet closing code, it can really fail only for two reasons: mutex poisoning or unknown wallet handle.

The latter is practically impossible since we create the wallet and the former is not really something within our control (the mutex gets poisoned if a panic happens in a different thread that acquired it but didn't get the chance to release it).

I could tweak `libvdrtools` a bit but I think the end result is the same, so instead I'll just log the error from closing the wallet, just like the one from the delete call (which can theoretically also fail).

In the event of an error, this still allows the user to be able to close/delete the wallet themselves (they provided all the stuff required for creating and opening it).

wallet_migrator/src/lib.rs
bobozaur: Just to be more specific, the helper function would literally just do something like:
```rust
async fn dry_run( 
    wallet_handle: WalletHandle,
    config: Config,
    credentials: Credentials,
    migrate_fn: impl FnMut(Record) -> Result<Option<Record>, E>,
) -> MigrationResult<()>
where
    E: Display,
{
    if migrate_wallet(wallet_handle, config, credentials, migrate_fn).await.is_ok() {
        let new_wh = Locator::instance()
            .wallet_controller
            .open(config.clone(), credentials.clone())
            .await
            .unwrap();

        Locator::instance().wallet_controller.close(new_wh).await.ok();
        Locator::instance().wallet_controller.delete(config, credentials).await.ok();
    }    
}
```

I guess my point is that if someone wants to do all this, they can. But I don't have a strong opinion on this, so if you'd rather have this implemented and it would aid you, I'll add it.

wallet_migrator/src/vdrtools2credx/conv.rs
bobozaur: Done. 

Note that the log level is `trace` because iterating all records can get very verbose and essentially users shouldn't be too interested in seeing their wallet items in clear text. I was even considering removing these logs together, honestly.

aries_vcx/migrations/20210903123945_migration0.sql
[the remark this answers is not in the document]
bobozaur: They should be the same now.

libvdrtools/indy-wallet/src/storage/mysql/mod.rs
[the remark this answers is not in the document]
Patrik-Stas: I see, good catch, I didn't know this. So then just putting in attention that our old wallet does not have `id`, field, but that would get created in migrated wallet. I think it should work fine though.

libvdrtools/indy-wallet/src/lib.rs
```diff
@@ -704,6 +703,65 @@ impl WalletService {
...
+    {
+        let old_wallet = self.get_wallet(old_wh).await?;
+        let new_wallet = self.get_wallet(new_wh).await?;
+
+        let mut records = old_wallet.get_all().await?;
+        let mut num_records = 0;
+
+        while let Some(WalletRecord {
+            type_,
+            id,
```
Patrik-Stas, line 723: Given that we found that the wallet schema is expected to have`id` field, I would suggest to rename this field in `WalletRecord` from `id` to `name`. Because the `id` field in WalletRecord / StorageRecord structs is originating from `name` column in the db schema.

image

libvdrtools/indy-wallet/src/storage/mysql/mod.rs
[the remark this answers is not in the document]
Patrik-Stas: This kinda brought me to notice and suggest following, to avoid confusion between the database schema "id" and `id` used in WalletRecord/StorageRecord
https://github.com/hyperledger/aries-vcx/pull/867#discussion_r1248142713

wallet_migrator/src/lib.rs
Patrik-Stas: I see, good point. I think we can add some dry run functionality like this later on vcx-napi-rs level.

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
Patrik-Stas: Touche, you are right, you can provide `Config` and `Credentials` with different db, different users etc. 

Thinking about how this migration API will be propagate to `vcx-napi-rs`, and how you would work with that from a NodeJS code in practice, how would you execute migration in server environment, I would find it cleaner if we take 
```
    source_config: Config,
    source_credentials: Credentials,
    target_config: Config,
    target_credentials: Credentials,
```
as  arguments into this function, and as suggested before, in the beginning of the function check that both wallets exists and be connected to, then proceed with migration, do not perform deletion at the end of the process.

That would give caller more flexibility how to approach creation of new wallet, how to possibly do dry runs with sqlite wallets, make the process for caller bit safer and "phased" by enabling him to create wallet himself first.

wallet_migrator/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+        .wallet_controller
+        .create(config.clone(), credentials.clone())
+        .await?;
+
+    debug!("Opening new wallet...");
```
Patrik-Stas, line 37: Assuming you agree on https://github.com/hyperledger/aries-vcx/pull/867/files#r1248167895
and we would work in this function with `source_config` and `target_config`, I'd tweak logs to include respective "config" when referring to one of the wallets. For example, here I would have logs like:

```
info!("Opening source wallet {:?}", source_config)
info!("Opening target wallet {:?}", target_config)
```

Also I think these high level processing logs deserve `info` rather than `debug`

wallet_migrator/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+        .await;
+
+    if let Err(e) = locator.wallet_controller.close(new_wh).await {
+        error!("Closing the wallet failed: {e}");
+    }
+
+    if let Err(e) = &res {
+        error!("Migration error encountered: {e}");
+
+        match locator.wallet_controller.delete(config, credentials).await {
```
Patrik-Stas, line 58: if you agree on https://github.com/hyperledger/aries-vcx/pull/867/files#r1248167895 then I'd suggest to omit any deletion of wallets as well - since we leave creation of wallets up to them, it would be consistent to let them handle deletion (if they want to) as well

libvdrtools/indy-wallet/src/storage/mysql/mod.rs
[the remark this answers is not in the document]
bobozaur: I removed the `id` column. It was only used in an irrelevant order by clause.

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
bobozaur: I don't see how that is better in any way, tbh. If they provide the wallet handle for the source wallet, that means they already created and maybe populated one.

If we get the arguments to open a wallet, we might have to treat the fact that the wallet might not even exist.

libvdrtools/indy-wallet/src/lib.rs
bobozaur: I'd do the opposite. I'd rename the name column to id, because that's exactly what gets passed in.

wallet_migrator/src/lib.rs
bobozaur: The configs contain some keys and other stuff that's not important outside of a debugging situation. 

You really only care about what happens in the info logs, with the minimal identification data (like the wallet.name).

But I will include the credentials and config, perhaps with some secret values.

wallet_migrator/src/lib.rs
bobozaur: I don't really see any benefit in it, really. But it might be just me.

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
Patrik-Stas: Imagine the code calling this, it will be working with the exact same arguments as I flashed out above, but then for the source wallet open it first, deal with wallet handle, but not for the target one. Why such a assymmetric approach?

You could say that "because the source wallet already exists, and the target one doesn't" - but I argue against that as well, and target wallet should also already exist. Migrator should just migrate IMO, not do other things - like create or try to delete a wallet, for the reasons argued above - perhaps I first need to convince you about this.

libvdrtools/indy-wallet/src/lib.rs
Patrik-Stas: ahh I see you've removed the `id` autoincremented field, so it's possible. Okay then - although being renamed to `id` I would usually expect that to be the primary key, which it won't be - I'll leave the final decision to your judgement.

wallet_migrator/src/lib.rs
Patrik-Stas: Ahha I see, I missed that config can contain secret values actually, good catch. Then I agree, lets just log wallet name (rather the wallet handle)

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
bobozaur: My reasons for finding this the way to go are:

1) As you said, the source wallet must already exist. There's no point in being provided with a set of `config` and `credentials` to open the wallet and maybe even have to deal with it not existing at all. The purpose of the migrator is to migrate existent data, and for that a `WalletHandle` is sufficient. Creating a source wallet is in no way the responsibility of the migrator.

2) The destination wallet is a different story. I find this whole thing to be sensitive enough that we cannot afford data loss or inconsistencies, and no matter how much we test this we should always warn about that risk (which is what ACA-Py also does).  

    So, if we accept a `WalletHandle` there's nothing preventing a user from messing up and providing the same handle as for the source wallet. We could add a simple equality check, but that's not enough either. A wallet can be opened multiple times and you'd get different handles pointing to the same underlying wallet. 

    In contrast, if we ask for a set of `config` and `credentials` which we can use to create a new wallet, there's 0% chance of migrating within the source wallet, because the creation would fail if the same wallet info is provided (honestly I haven't directly tested this, but given that a new schema is created for the wallet I am confident that the database would throw an error on trying to create a duplicate). 

    So we are absolutely sure that whatever happens, no matter the error, only the destination wallet can get messed up and the source wallet will remain intact. We do our best to cleanup the destination wallet if that happens, but even if we fail to delete it, it's not the end of the world and the users can clean up. As for the source wallet, in the event of a successful migration consumers can delete it themselves (preferably after thorough testing).

3) You could also argue that the migrator does not delete records, so even if the same wallet is used as source and destination and an error occurs somewhere, the old records would still be there (this assuming that we use only `WalletHandle`s and remove the wallet deletion code on failure). And you'd be right. But the records that *did* get migrated will cause duplicate errors to occur on subsequent migrations. We can definitely handle all this, but this adds overhead, increases complexity, and all for, in my view, no good reason.

I personally think this more than justifies the asymmetric function arguments, which I honestly find to be quite a small thing anyway. I understand that maybe modifying this might make it easier for testing purposes and whatnot, but I'll always stand by my belief that the tests should accommodate production code, not the other way around.

Don't take this the wrong way, but I really doubt that my opinion on this will change. I am waaaaaaay more concerned with what calling this does than what the code that calls this looks like. So maybe we can loop @mirgee and @gmulhearn  to chip in and whatever the majority decides, we implement.

libvdrtools/indy-wallet/src/lib.rs
bobozaur: Well, I looked further into this and the problem is that if we were to update the `name` column to `id`, we should do that everywhere, including in the `SQLite` impl.

But the `SQLite` impl has an `id` field already, and it's being used for other purposes too. So... maybe we should just leave it like this. We're not really aiming to maintain libvdrtools, right? So someday I guess we'll want to drop even the `Indy` wallet.

And I'm also saying this because this is the kind of "problem" that wouldn't have surfaced if not for the modifications I made to `libvdrtools` to assist with the migration.

wallet_migrator/src/lib.rs
bobozaur: There are also `trace` logs in `libvdrtools` with the full `credentials` and `config`, so maybe we can also leave this as it is, with just the wallet name getting logged as `debug`.

libvdrtools/indy-wallet/src/lib.rs
Patrik-Stas: okay, makes sense. Since we removed the `id` autoincrement in the schema, that mostly eliminates the potential confusion I was pointing out in my original comment.

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
bobozaur: Just adding the update of our discussion here:

We're still not sure how to parametrize this function, although I guess we'd all rather have two `Config` and `Credentials` sets, one for the old wallet and one for the new wallet, rather than two `WalletHandle`s. The only downside is that we have to error out if the source wallet does not exist (which is by no means a big deal, we just try to open it).

If people would rather see it like that, then I'll implement it that way.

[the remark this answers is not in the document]
Patrik-Stas: So we had some internal discussion and I am open to leaving things as they are. 

Some of my concern was also coming from the fact that create wallet (for mysql wallet type) now also create db schemas. But it does make sense considering sqlite type performs analogical setup, and if we ever need to migrate the schema, we can support that directly from the vdrtools wallet code.

Patrik-Stas (approved):

wallet_migrator/src/lib.rs
Patrik-Stas: When I run the migration, I would like to have easy way to see migrator's logs (without having to fiddle with rust_log and setup logs for migrator crate specifically)
I would like to see:
- when migrations process starts (info)
- creation of wallet - start and finish - as info
- want to see it's progressing (on per record that could be debug or trace, but you didn't address that), ideally perhaps every 100th migration record also info log
- when migration finish (info)
- deletion of wallet (on error) as warn

Also in general:
- be bit more explicit in logs, eg:
```
debug!("Opening new wallet...");
```
would be better to have something like:
```
debug!("Opening migration target wallet ${config.name}");
```

Additionally, I can't help it but:
```
debug!("Migrating records from wallet with handle {wallet_handle:?} to wallet with handle {new_wh:?}");
```
I find the inability to produce meaningful log (wallet handle its are not) message in **migrator** about 2 most important things - source and destination - are codesmell related to the signature. This kinda brings me again thoughts that migrator should be responsible for opening both wallets, and should be aware of what the source and destination wallets are.

I find logging pretty important for tool for this kind of operational tool like migrator. If something goes wrong, you really want to know what happened.

Patrik-Stas (changes requested):
left further comment on logging

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
mirgee: So to summarise, my very limited understanding of the tradeoffs in consideration here are: either

1. both the source and the target wallets are created and opened outside of the migration script and two wallet handles are passed in, or
2. both the source and the target wallets are created and opened outside of the migration script and two sets of configs are passed in, or
3. the source wallet is opened outside, destination wallet created and opened inside the script and source wallet handle and dest. wallet config are passed in.

As for 1., on one hand it makes sense to decouple opening the source wallet from the actual migration, but on the other hand it is done at the cost of exposing a lower-level primitive - the wallet handle.

As for 2., it shields the user away from the wallet handles, mitigates the danger of source and wallet handle being of the same wallet, and a small drawback is that in order to perform any preparatory steps, the wallets need to be opened and closed before performing the migration, which opens and closes them again.

As for 3., in a sense it provides the worst of all possible worlds, as it allows for passing wallet handles of the same wallet in addition to the drawback of 1., but it allows for the flexibility of doing preparatory steps in between opening the wallets and performing the migration.

Note the limited practicality of preparatory steps mentioned as wallet migration is already performed during the wallet creation (which makes sense, as the schema is inherently coupled to indy-wallet implementation).

Given the above, going with 2. makes the most sense to me, but it may be just because I missed some critical insight.

wallet_migrator/src/lib.rs
bobozaur: Most of the stuff was already there...

1) when migrations process starts (info) - definitely was already there.
2) creation of wallet - start and finish - as info - the creation was there, though as `debug`. I assumed the finish is implied.
3) want to see it's progressing (on per record that could be debug or trace, but you didn't address that), ideally perhaps every 100th migration record also info log -> This is done in the `migrate_fn`, the one provided as `vdrtools2credx::migrate_any_records` already logs the initial and migrated records as `trace`. And, by all means, an arbitrary closure can be passed that wraps this function, or per type ones in `conv.rs`
4) when migration finish (info) - wasn't there, but I again assumed this is implied.
5) deletion of wallet (on error) as warn - was there, but as `error`.

Personally I don't see this as a "tool" (it's not like some CLI), but a library. And it's flexible enough that you can do whatever you want on a record by record basis.

In any case, I've been working on adding logging and change the signature to use two `Config` and `Credentials` sets when I decided to look into `WalletHandle` some more. I don't know why I was so convinced that multiple wallet instances can be open with different handles, but it turns out they can't. Opening the same wallet a second time fails. This is my bad, sorry.

This essentially means that accepting two handles and checking for equality is sufficient to ensure that the source and destination wallets are not the same, so I implemented that. This should solve the asymmetry concern and make most of the wallet logging obsolete.

bobozaur: Maybe on short, the function now takes two `WalletHandle`s, errors out if they are the same, and the record logging was already there at a `trace` level. Both the old and the migrated records get logged.

Please let me know if this is not sufficient.

wallet_migrator/src/lib.rs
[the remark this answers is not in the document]
bobozaur: Well, the `Config` and `Credentials` structs are also part of `libvdrtools`, so `WalletHandle` isn't that bad, I guess... 
Nevertheless, after a much closer look I realized that I was very wrong and that the same wallet cannot be opened twice. So I modified the migrator to only take two `WalletHandle`s and error out if they are equal. This also is more in line with what you were saying about the migrator only really dealing with the migration, and not managing wallets.

Patrik-Stas (approved):

mirgee (approved):

PR #869, opened 2023-06-05
Extend Profile trait with update_taa_configuration method

Patrik-Stas:
In PR https://github.com/hyperledger/aries-vcx/pull/863 we have eliminated global state used by modular profile. However:
- we didn't have way to set up TAA in runtime without having to rebuild entire Profile
- that was manifested by `ModularLibsProfile` having fields with `public(crate)` visibility so we could accommodate different initialization flow which needs indy writer with TAA (also resulting in some code duplication in `aries_vcx/src/utils/devsetup.rs`)

- This PR introduces interior mutability within `ModularLibsProfile` such that it's possible to update TAA setup for ledger writer in runtime - doing so by wrapping the `anoncreds_ledger_write`, `indy_ledger_write` in `RwLock` - note these wrapped trait objects are the same instance under the hood.
- As RwLocks has been added wrapping the ledger write trait objects, in order to prevent having to modify codebase across many files and deal with locks everywhere, obtaining locks has been encapsulated into `IndyLedgerWriteProxy`, `AnoncredsLedgerWriteProxy` which implement `IndyLedgerWriteProxy`, `AnoncredsLedgerWriteProxy` respectively. This enabled us to introduce RwLocks without surrounding code consuming Profile trait being aware of it.

### DIFF
--- a/aries_vcx/Cargo.toml
+++ b/aries_vcx/Cargo.toml
@@ -50,6 +50,7 @@ derive_builder = "0.10.2"
 tokio = { version = "1.20.4" }
 thiserror = "1.0.37"
 url = { version = "2.3", features = ["serde"] }
+async-lock = "2.7.0"
 
 [target.'cfg(target_os = "android")'.dependencies]
 android_logger = "0.5"
--- a/aries_vcx/src/core/profile/mixed_breed_profile.rs
+++ b/aries_vcx/src/core/profile/mixed_breed_profile.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use aries_vcx_core::ledger::base_ledger::TxnAuthrAgrmtOptions;
 use aries_vcx_core::{
     anoncreds::{base_anoncreds::BaseAnonCreds, credx_anoncreds::IndyCredxAnonCreds},
     ledger::{
@@ -9,6 +10,7 @@ use aries_vcx_core::{
     wallet::{base_wallet::BaseWallet, indy_wallet::IndySdkWallet},
     PoolHandle, WalletHandle,
 };
+use async_trait::async_trait;
 
 use super::profile::Profile;
 
@@ -40,6 +42,7 @@ impl MixedBreedProfile {
     }
 }
 
+#[async_trait]
 impl Profile for MixedBreedProfile {
     fn inject_indy_ledger_read(self: Arc<Self>) -> Arc<dyn IndyLedgerRead> {
         Arc::clone(&self.indy_ledger_read)
@@ -63,5 +66,9 @@ impl Profile for MixedBreedProfile {
 
     fn inject_wallet(&self) -> Arc<dyn BaseWallet> {
         Arc::clone(&self.wallet)
+    }
+
+    async fn update_taa_configuration(self: Arc<Self>, taa_options: TxnAuthrAgrmtOptions) {
+        todo!()
     }
 }
--- a/aries_vcx/src/core/profile/modular_libs_profile.rs
+++ b/aries_vcx/src/core/profile/modular_libs_profile.rs
@@ -1,18 +1,23 @@
+use async_lock::RwLock;
+use std::ops::Deref;
 use std::sync::Arc;
 use std::time::Duration;
 
 use aries_vcx_core::anoncreds::base_anoncreds::BaseAnonCreds;
 use aries_vcx_core::anoncreds::credx_anoncreds::IndyCredxAnonCreds;
-use aries_vcx_core::ledger::base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerWrite, IndyLedgerRead, IndyLedgerWrite};
+use aries_vcx_core::errors::error::VcxCoreResult;
+use aries_vcx_core::ledger::base_ledger::{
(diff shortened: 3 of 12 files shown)


aries_vcx/src/core/profile/mixed_breed_profile.rs
```diff
@@ -64,4 +67,8 @@ impl Profile for MixedBreedProfile {
     fn inject_wallet(&self) -> Arc<dyn BaseWallet> {
         Arc::clone(&self.wallet)
     }
+
+    async fn update_taa_configuration(self: Arc<Self>, taa_options: TxnAuthrAgrmtOptions) {
+        todo!()
```
mirgee, line 72: Just don't forget to put error log here instead, please.

aries_vcx/src/core/profile/vdrtools_profile.rs
```diff
@@ -63,4 +65,8 @@ impl Profile for VdrtoolsProfile {
     fn inject_wallet(&self) -> Arc<dyn BaseWallet> {
         Arc::clone(&self.wallet)
     }
+
+    async fn update_taa_configuration(self: Arc<Self>, taa_options: TxnAuthrAgrmtOptions) {
+        todo!()
```
mirgee, line 70: ... and here.

ci/alpine_core.dockerfile
```diff
@@ -18,8 +18,8 @@ RUN apk update && apk upgrade && \
 USER indy
 WORKDIR /home/indy
 
-ARG RUST_VER="1.65.0"
+ARG RUST_VER="1.66.0"
```
mirgee, line 21: Can we use one Rust version consistently across the codebase?

aries_vcx/Cargo.toml
```diff
@@ -50,6 +50,7 @@ derive_builder = "0.10.2"
 tokio = { version = "1.20.4" }
 thiserror = "1.0.37"
 url = { version = "2.3", features = ["serde"] }
+async-lock = "2.7.0"
```
mirgee, line 53: Why introduce a new crate instead of using `tokio::sync::RwLock`, since we already depend on `tokio`.

aries_vcx/src/core/profile/vdrtools_profile.rs
```diff
@@ -63,4 +65,8 @@ impl Profile for VdrtoolsProfile {
     fn inject_wallet(&self) -> Arc<dyn BaseWallet> {
         Arc::clone(&self.wallet)
     }
+
+    async fn update_taa_configuration(self: Arc<Self>, taa_options: TxnAuthrAgrmtOptions) {
+        panic!("update_taa_configuration not implemented for VdrtoolsProfile")
```
mirgee, line 70: Why panic here and error in `VdrProxyProfile`.

mirgee: And isn't it possible to update the global in settings upon calling this?

aries_vcx/src/core/profile/vdr_proxy_profile.rs
```diff
@@ -93,4 +95,8 @@ impl Profile for VdrProxyProfile {
     fn inject_wallet(&self) -> Arc<dyn BaseWallet> {
         Arc::clone(&self.wallet)
     }
+
+    async fn update_taa_configuration(self: Arc<Self>, taa_options: TxnAuthrAgrmtOptions) {
+        error!("update_taa_configuration not implemented for MixedBreedProfile")
```
mirgee, line 100: Incorrect error message. Also, why can't it be implemented?

aries_vcx/src/core/profile/modular_libs_profile.rs
```diff
@@ -21,18 +26,125 @@ use super::profile::Profile;
...
+    async fn publish_nym(
+        &self,
+        submitter_did: &str,
+        target_did: &str,
+        verkey: Option<&str>,
+        data: Option<&str>,
+        role: Option<&str>,
+    ) -> VcxCoreResult<String> {
+        self.indy_ledger_write
+            .read()
```
mirgee, line 68: As we discussed, I am not a fan of the overhead associated with acquiring the read lock on every write operation. (Nor am I a fan of having the ability to randomly change TAA during runtime.) It would be good if we managed to find a solution which avoids this downside.

aries_vcx/Cargo.toml
Patrik-Stas: removed, not needed after all

aries_vcx/src/utils/devsetup.rs
```diff
@@ -411,7 +413,7 @@ impl SetupProfile {
         .unwrap();
         let pool_handle = open_test_pool().await;
 
-        let profile: Arc<dyn Profile> = Arc::new(VdrtoolsProfile::init(wallet_handle, pool_handle.clone()));
+        let mut profile = Arc::new(VdrtoolsProfile::init(wallet_handle, pool_handle.clone()));
```
mirgee, line 416: Why is mutable reference here?

mirgee (approved):
OK

PR #871, opened 2023-06-05
Sovrin-specific DDO facade

mirgee:
Introduces wrappers around `DidDocument` and `Service` in order to
* handle logic specific to the Sovrin method, e.g.
    * setting and validating `type` and `accept` fields based on the composition of the service extra fields,
    * dereferencing service keys (?)
* remove definition of the Sovrin extra fields from `did_resolver_sov`,
* simplify common Sovrin DDO usage patterns (e.g. retrieving recipient / routing key from the first service in the `service` field, etc.).


did_doc/src/schema/service.rs
```diff
@@ -55,6 +52,13 @@ where
 
 #[derive(Debug)]
 pub struct ServiceBuilder<E> {
+    id: Uri,
+    service_endpoint: Url,
+    extra: E,
+}
+
+#[derive(Debug)]
+pub struct ServiceBuilderWithServiceType<E> {
```
Patrik-Stas, line 61: What is this for? Where is `service_type` specified? (can't find it in did:sov method or didcore)

did_doc_sov/src/error.rs
```diff
@@ -0,0 +1,13 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum DidDocumentSovError {
+    #[error("Attempted to access empty collection: {0}")]
+    EmptyCollection(&'static str),
+    #[error("DID document builder error: {0}")]
+    DidDocumentBuilderError(#[from] did_doc::error::DidDocumentBuilderError),
```
Patrik-Stas, line 8: seems unused so far

Patrik-Stas (commented):
left some comments

After this PR, perhaps it's really time to put all the did/ddo stuff into a directory, the number of crates is quite expanding, but I really like the separation of concerns in each of them

did_doc/src/schema/service.rs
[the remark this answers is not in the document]
mirgee: First, note that this is a builder, not the final struct. Second, note that the final struct is (de)serialized with this field renamed to "type"; `service_type` is used since `type` is a reserved Rust keyword.

did_doc_sov/src/extra_fields/didcommv2.rs
```diff
@@ -0,0 +1,43 @@
+use serde::{Deserialize, Serialize};
+
+use super::AcceptType;
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct ExtraFieldsDidCommV2 {
```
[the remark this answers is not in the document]
mirgee: The Sovrin method spec you linked in the comment above.

did_doc_sov/src/extra_fields/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+use serde::{Deserialize, Deserializer, Serialize};
+
+use crate::error::DidDocumentSovError;
+
+pub mod aip1;
+pub mod didcommv1;
+pub mod didcommv2;
+
+#[derive(Serialize, Clone, Debug, PartialEq)]
+pub enum AcceptType {
```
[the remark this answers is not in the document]
mirgee: Why do you find it misplaced here? Not only is `AcceptType` part of extra fields (only), but is linked with the "kind" of extra fields, as you say.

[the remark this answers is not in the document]
Patrik-Stas: ahha right, it's part of `ExtraFields`. Touche, nevermind :-)

did_doc_sov/src/error.rs
mirgee: It is used (quite a bit actually).

did_doc_sov/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    pub fn service(&self) -> Result<ServicesList, DidDocumentSovError> {
+        Ok(ServicesList::new(
+            self.did_doc
+                .service()
+                .iter()
+                .map(|s| ServiceSov::try_from(s.to_owned()))
+                .collect::<Result<Vec<ServiceSov>, _>>()?,
+        ))
```
mirgee, line 57: TODO: The conversion should be be done once during deserialization / building of `DidDocumentSov`.

did_doc_sov/src/extra_fields/aip1.rs
```diff
@@ -0,0 +1,4 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Default)]
+pub struct ExtraFieldsAIP1 {}
```
bobozaur, line 4: There's a type in `shared_vcx` that I made when working on the `messages` one (it was in messages initially but then moved to shared_vcx in a good first issue) and it serves the exact purpose of serializing to nothing and deserializing from thin air. 

It's called `SerdeIgnored` or something like that. What do you think about re-using that instead of another type like `ExtraFieldsAIP1`?

mirgee: Interesting! Didn't know about the flattening behavior, I guess I was just lucky in this case 😅 

I have no issue with reusing `SerdeIgnored`, except I am not too fond of pulling in a dependency (`shared_vcx`) just for that, to be honest...

bobozaur: Hmmm... I get you, but I thought the whole point of `shared_vcx` was to allow reusing utilities across the other crates, or am I misunderstanding?

If it isn't, maybe it would be worth creating one like that. The `OneOrList` type you made could go in there too.

I guess it shouldn't be a big deal to pull in a local dependency from a compilation standpoint, as it will always have one version across all places used.

bobozaur: Just to clarify, this isn't a showstopper so disregard if it's unimportant 😅 . I'm just brainstorming about reusing utils that we make out of necessity.

PR #872, opened 2023-06-06
Refactor/do not consume profile

Patrik-Stas:


gmulhearn (changes requested):
lgtm, but IMO now would probably be the best time to update all "clones" of `Arc<dyn Profile>`. As previously the code would call `Arc::clone(&profile)` whenever it needed to inject ledger/anoncreds.

e.g. in `holder/state_machine.rs`'s `_store_credential`:
```
    let ledger = Arc::clone(profile).inject_anoncreds_ledger_read();
    let anoncreds = Arc::clone(profile).inject_anoncreds();
    ```

can now just be:
```
    let ledger = profile.inject_anoncreds_ledger_read();
    let anoncreds = profile.inject_anoncreds();
```

**Comment by gmulhearn** [comment]
^ i think if you do a search for `Arc::clone` in the codebase, pretty much every `Arc::clone` call is to clone a profile, which is now unnecessary.

could also probably argue that the signature could/should be updated to `&Arc` rather than `Arc`. Which could allow implementations of `Profile` to not have to `Arc:;clone` before returning..

e.g. :
```
    fn inject_indy_ledger_read(&self) -> &Arc<dyn IndyLedgerRead>;

    fn inject_indy_ledger_write(&self) -> &Arc<dyn IndyLedgerWrite>;

    fn inject_anoncreds(&self) -> &Arc<dyn BaseAnonCreds>;

    fn inject_anoncreds_ledger_read(&self) -> &Arc<dyn AnoncredsLedgerRead>;

    fn inject_anoncreds_ledger_write(&self) -> &Arc<dyn AnoncredsLedgerWrite>;

    fn inject_wallet(&self) -> &Arc<dyn BaseWallet>;
```

this would have no downside, as if the consumer really does need an `Arc`, then can always just `Arc::clone` over the `&Arc`.

But up to you whether you want to include that in this PR

Patrik-Stas:
@gmulhearn thanks for review and comments, however I'd like to suggest to push the suggested updates further down the road as this https://github.com/hyperledger/aries-vcx/pull/873 is building on top of this PR and generally eliminate profiles from non-testing portion of aries-vcx codebase. Profiles will still exist post-#873 however, so we can apply your suggestion afterwards. Right now addressing your suggestion would require many changes which would be subsequently removed by 873.

gmulhearn (approved):
lgtm!

gmulhearn:
yea my bad, forgot about the profile removal work coming soon. approved!

PR #873, opened 2023-06-06
Inject only required components, not entire profiles

Patrik-Stas:
`Profile` abstraction has been introduced to enable smooth transition from `vdrtools` to modular libs (credx/anoncreds, indy-vdr, askar). `Profile` represents basket of trait objects to perform variety of operations.

Previously `Profile` was widely present in `aries-vcx` codebase as a convenient way to access various lower-level APIs. However this was at the price of:
- passing references to objects down the stack, even though they were never used,
- requiring `aries-vcx` consumers to build entire "Profile", even though the APIs they've consequently used only used limited number of profile's components (eg, you don't need access to ledger in order to build credential proposal, for example)
- forcing `aries-vcx` to store building blocks in particular way (Profiles) - but consuming applications might want to store/handle things differntly.

This PR modifies `aries-vcx` and `aries-vcx-core` method signatures to be unaware of `Profile` abstraction - the APIs now only speak in terms of individual components such as `dyn BaseAnoncreds`, `dyn AnoncredsLedgerRead` - so the APIs ask only for what they really need.
Additional extra benefit is that it's now much easier to spot fishy stuff, like "oh, this method requires access to ledger? That's weird, why is that???"

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -555,6 +555,7 @@ jobs:
       - name: "Run libvcx tests: pool_tests"
         run: |
           RUST_TEST_THREADS=1 cargo test --manifest-path="libvcx/Cargo.toml" -F "pool_tests";
+          RUST_TEST_THREADS=1 cargo test --manifest-path="libvcx_core/Cargo.toml" -F "pool_tests";
 
   test-integration-resolver:
     needs: workflow-setup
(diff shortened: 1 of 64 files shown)


mirgee (approved):

Patrik-Stas:
@gmulhearn if this is fine with you, feel free to merge

Big delta but trivial changes mostly

bobozaur (approved):

PR #880, opened 2023-06-09
Refactor/testing

Patrik-Stas:
- Aries-vcx: Streamlines and deduplicates test initialization
- Libvcx: Minimize usage of "handle" vocabulary in favor of working with trait object components. This was preliminary work to enable running libvcx with different than vdrtools profile

# Aries-vcx Alice/Faber testing
- Implement `create_faber` similarly like we had `create_alice`, which construct the test agent with regard to to feature flags. This significantly increases coverage of issuer/verifier tests running modular libs profile.

- Use `SetupProfile` as basis to build out components for Faber/Alice (`create_faber`/`create_alice`) in integration tests in favor of logic duplicated in `setup_modular_profile`, `setup_indy_wallet`, `setup_indy_wallet` (now deleted). This again significantly increases coverage for different profile feature flags

# Libvcx
- Rework global storage/Profile handling in `libvcx_core`. Remove notion of "handles" from libvcx global state, instead is replaced with dealing with individual components (base wallet, anoncreds etc.). This opens up door for swapping different implementation on libvcx level, such as using modural libs components

- Remove testing feature infra from libvcx, run libvcx unit tests in CI (this was dropped somewhere along the way)
- Decrease usage of global "settings" variables used in libvcx tests

# Various smaller changes
- removal of unused or under-utilized "Setup" objects
- removal of commented tests
- removal of low value tests - typically 1 or 2 tests using specific "Setup" object (which itself would require refactoring) while not delivering much value - removed tests along with the setup
- split Faber and Alice integration test agents to separate files
- Rename `SetupPool` to `SetupPoolDirectory` and change behaviour, such that it doesn't connect to pool, only prepares path to store temporary pool genesis file. Previously in some tests, we been doing extra unused connection to pool. 
- Delete `TestSetupCreateWallet` and the single libvcx (deprecated component) test utilizing it 'test_vcx_init_called_twice_passes_after_shutdown' - wallet/pool re-initialization is still indirectly covered by NodeJS integration tests
- Delete "SetupPoolConfig" and libvcx test utilizing it "test_full_init" (similar story like above)
- Use unique poolname (for vdrtools bases profiles) to increase resilience of test suites
- Change `state_vcx_shutdown` behaviour - ignore `delete` argument, which deletes wallet and pool. This was unused and is practically rather dangerous than useful.


bobozaur (approved):
LGTM

bobozaur (approved):

PR #883, opened 2023-06-21
Implementation of did:peer:2 DID method

mirgee:
Implementation of did:peer method 2:

* construction of did:peer:2 peer DID based on a DID document
* construction of DID document based on did:peer:2 peer DID.

The main non-functional requirement is maximal interoperability with the already existing implementations in other Aries frameworks.

### DIFF
--- a/did_doc/src/schema/verification_method/public_key.rs
+++ b/did_doc/src/schema/verification_method/public_key.rs
@@ -1,0 +1,112 @@
+use std::str::FromStr;
+
+use base64::{engine::general_purpose, Engine};
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    error::DidDocumentBuilderError,
+    schema::types::{jsonwebkey::JsonWebKey, multibase::Multibase},
+};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(untagged)]
+#[serde(deny_unknown_fields)]
+pub enum PublicKeyField {
+    #[serde(rename_all = "camelCase")]
+    Multibase { public_key_multibase: String },
+    #[serde(rename_all = "camelCase")]
+    Jwk { public_key_jwk: JsonWebKey },
+    #[serde(rename_all = "camelCase")]
+    Base58 { public_key_base58: String },
+    #[serde(rename_all = "camelCase")]
+    Base64 { public_key_base64: String },
+    #[serde(rename_all = "camelCase")]
+    Hex { public_key_hex: String },
+    #[serde(rename_all = "camelCase")]
+    Pem { public_key_pem: String },
+    #[serde(rename_all = "camelCase")]
+    Pgp { public_key_pgp: String },
+}
+
+impl PublicKeyField {
+    pub fn key_decoded(&self) -> Result<Vec<u8>, DidDocumentBuilderError> {
+        match self {
+            PublicKeyField::Multibase {
+                public_key_multibase,
+            } => {
+                let multibase = Multibase::from_str(public_key_multibase)?;
+                Ok(multibase.as_ref().to_vec())
+            }
+            PublicKeyField::Jwk { public_key_jwk } => public_key_jwk.to_vec(),
+            PublicKeyField::Base58 { public_key_base58 } => {
+                Ok(bs58::decode(public_key_base58).into_vec()?)
+            }
+            PublicKeyField::Base64 { public_key_base64 } => {
+                Ok(general_purpose::STANDARD_NO_PAD.decode(public_key_base64.as_bytes())?)
+            }
+            PublicKeyField::Hex { public_key_hex } => Ok(hex::decode(public_key_hex)?),
+            PublicKeyField::Pem { public_key_pem } => {
+                Ok(pem::parse(public_key_pem.as_bytes())?.contents().to_vec())
+            }
+            PublicKeyField::Pgp { public_key_pgp: _ } => Err(
+                DidDocumentBuilderError::UnsupportedPublicKeyField("publicKeyPgp"),
+            ),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    static PUBLIC_KEY_MULTIBASE: &str = "z6LSbysY2xFMRpGMhb7tFTLMpeuPRaqaWM1yECx2AtzE3KCc";
+    static PUBLIC_KEY_BASE58: &str = "6LSbysY2xFMRpGMhb7tFTLMpeuPRaqaWM1yECx2AtzE3KCc";
+    static PUBLIC_KEY_BASE64: &str = "7AEEiIVxASfd1+8HamOWE5BCi6vqNfL13mzYUoQk1M4mKQ";
+    static PUBLIC_KEY_HEX: &str =
+        "ec01048885710127ddd7ef076a63961390428babea35f2f5de6cd8528424d4ce2629";
+    static PUBLIC_KEY_BYTES: [u8; 34] = [
+        236, 1, 4, 136, 133, 113, 1, 39, 221, 215, 239, 7, 106, 99, 150, 19, 144, 66, 139, 171,
+        234, 53, 242, 245, 222, 108, 216, 82, 132, 36, 212, 206, 38, 41,
+    ];
+
+    #[test]
+    fn test_multibase() {
+        let public_key_field = PublicKeyField::Multibase {
+            public_key_multibase: PUBLIC_KEY_MULTIBASE.to_string(),
+        };
+        assert_eq!(public_key_field.key_decoded().unwrap(), PUBLIC_KEY_BYTES);
+    }
+
+    #[test]
+    fn test_base58() {
+        let public_key_field = PublicKeyField::Base58 {
+            public_key_base58: PUBLIC_KEY_BASE58.to_string(),
+        };
+        assert_eq!(
+            public_key_field.key_decoded().unwrap(),
+            PUBLIC_KEY_BYTES.to_vec()
+        );
+    }
+
+    #[test]
+    fn test_base64() {
+        let public_key_field = PublicKeyField::Base64 {
+            public_key_base64: PUBLIC_KEY_BASE64.to_string(),
+        };
+        assert_eq!(
+            public_key_field.key_decoded().unwrap(),
+            PUBLIC_KEY_BYTES.to_vec()
+        );
+    }
+
+    #[test]
+    fn test_hex() {
+        let public_key_field = PublicKeyField::Hex {
+            public_key_hex: PUBLIC_KEY_HEX.to_string(),
+        };
+        assert_eq!(
+            public_key_field.key_decoded().unwrap(),
+            PUBLIC_KEY_BYTES.to_vec()
+        );
+    }
+}
--- a/did_doc_sov/src/extra_fields/didcommv1.rs
+++ b/did_doc_sov/src/extra_fields/didcommv1.rs
@@ -68,11 +68,8 @@ impl ExtraFieldsDidCommV1Builder {
         self
     }
 
-    pub fn set_accept<I>(mut self, accept: I) -> Self
-    where
-        I: IntoIterator<Item = AcceptType>,
-    {
-        self.accept = accept.into_iter().collect();
+    pub fn set_accept(mut self, accept: Vec<AcceptType>) -> Self {
+        self.accept = accept;
         self
     }
 
@@ -86,7 +83,7 @@ impl ExtraFieldsDidCommV1Builder {
             priority: self.priority,
             recipient_keys: self.recipient_keys,
             routing_keys: self.routing_keys,
-            accept: self.accept.into_iter().collect(),
+            accept: self.accept,
         }
     }
 }
--- a/did_doc_sov/src/extra_fields/didcommv1.rs
+++ b/did_doc_sov/src/extra_fields/didcommv1.rs
@@ -8,6 +8,7 @@ pub struct ExtraFieldsDidCommV1 {
     priority: u32,
     recipient_keys: Vec<KeyKind>,
     routing_keys: Vec<KeyKind>,
+    #[serde(default)]
     accept: Vec<AcceptType>,
 }
 
@@ -33,11 +34,22 @@ impl ExtraFieldsDidCommV1 {
     }
 }
 
-#[derive(Default)]
 pub struct ExtraFieldsDidCommV1Builder {
     priority: u32,
     recipient_keys: Vec<KeyKind>,
     routing_keys: Vec<KeyKind>,
+    accept: Vec<AcceptType>,
+}
+
+impl Default for ExtraFieldsDidCommV1Builder {
+    fn default() -> Self {
+        Self {
+            priority: 0,
+            recipient_keys: Vec::new(),
+            routing_keys: Vec::new(),
+            accept: vec![AcceptType::DIDCommV1],
+        }
+    }
 }
 
 impl ExtraFieldsDidCommV1Builder {
@@ -56,12 +68,25 @@ impl ExtraFieldsDidCommV1Builder {
         self
     }
 
+    pub fn set_accept<I>(mut self, accept: I) -> Self
+    where
+        I: IntoIterator<Item = AcceptType>,
+    {
+        self.accept = accept.into_iter().collect();
+        self
+    }
+
+    pub fn add_accept(mut self, accept: AcceptType) -> Self {
+        self.accept.push(accept);
+        self
+    }
+
     pub fn build(self) -> ExtraFieldsDidCommV1 {
         ExtraFieldsDidCommV1 {
             priority: self.priority,
             recipient_keys: self.recipient_keys,
             routing_keys: self.routing_keys,
-            accept: vec![AcceptType::DIDCommV1],
+            accept: self.accept.into_iter().collect(),
         }
     }
 }
--- a/did_doc_sov/src/extra_fields/didcommv2.rs
+++ b/did_doc_sov/src/extra_fields/didcommv2.rs
@@ -44,11 +44,8 @@ impl ExtraFieldsDidCommV2Builder {
         self
     }
 
-    pub fn set_accept<I>(mut self, accept: I) -> Self
-    where
-        I: IntoIterator<Item = AcceptType>,
-    {
-        self.accept = accept.into_iter().collect();
+    pub fn set_accept(mut self, accept: Vec<AcceptType>) -> Self {
+        self.accept = accept;
         self
     }
 
@@ -60,7 +57,7 @@ impl ExtraFieldsDidCommV2Builder {
     pub fn build(self) -> ExtraFieldsDidCommV2 {
         ExtraFieldsDidCommV2 {
             routing_keys: self.routing_keys,
-            accept: self.accept.into_iter().collect(),
+            accept: self.accept,
         }
     }
 }
--- a/did_doc_sov/src/extra_fields/didcommv2.rs
+++ b/did_doc_sov/src/extra_fields/didcommv2.rs
@@ -5,8 +5,9 @@ use super::{AcceptType, KeyKind};
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct ExtraFieldsDidCommV2 {
-    accept: Vec<AcceptType>,
     routing_keys: Vec<KeyKind>,
+    #[serde(default)]
+    accept: Vec<AcceptType>,
 }
 
 impl ExtraFieldsDidCommV2 {
@@ -23,9 +24,18 @@ impl ExtraFieldsDidCommV2 {
     }
 }
 
-#[derive(Default)]
 pub struct ExtraFieldsDidCommV2Builder {
     routing_keys: Vec<KeyKind>,
+    accept: Vec<AcceptType>,
+}
+
+impl Default for ExtraFieldsDidCommV2Builder {
+    fn default() -> Self {
+        Self {
+            routing_keys: Vec::new(),
+            accept: vec![AcceptType::DIDCommV2],
+        }
+    }
 }
 
 impl ExtraFieldsDidCommV2Builder {
@@ -34,10 +44,23 @@ impl ExtraFieldsDidCommV2Builder {
         self
     }
 
+    pub fn set_accept<I>(mut self, accept: I) -> Self
+    where
+        I: IntoIterator<Item = AcceptType>,
+    {
+        self.accept = accept.into_iter().collect();
+        self
+    }
+
+    pub fn add_accept(mut self, accept: AcceptType) -> Self {
+        self.accept.push(accept);
+        self
+    }
+
     pub fn build(self) -> ExtraFieldsDidCommV2 {
         ExtraFieldsDidCommV2 {
-            accept: vec![AcceptType::DIDCommV2],
             routing_keys: self.routing_keys,
+            accept: self.accept.into_iter().collect(),
         }
     }
 }
--- a/did_parser/src/did.rs
+++ b/did_parser/src/did.rs
@@ -4,6 +4,7 @@ use std::str::FromStr;
 
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
+use crate::DidUrl;
 use crate::{error::ParseError, utils::parse::parse_did_method_id, DidRange};
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
@@ -13,6 +14,7 @@ pub struct Did {
     id: DidRange,
 }
 
+// TODO: Add a builder / constructor so that we don't have to create strings and parse them?
 impl Did {
     pub fn parse(did: String) -> Result<Self, ParseError> {
         let (_, method, id) = parse_did_method_id(&did)?;
@@ -89,5 +91,11 @@ impl<'de> Deserialize<'de> for Did {
     {
         let did = String::deserialize(deserializer)?;
         Self::parse(did).map_err(serde::de::Error::custom)
+    }
+}
+
+impl From<Did> for DidUrl {
+    fn from(did: Did) -> Self {
+        Self::from_did_parts(did.did().to_string(), 0..did.did.len(), did.method, did.id)
     }
 }
--- a/did_parser/src/did_url.rs
+++ b/did_parser/src/did_url.rs
@@ -146,6 +146,48 @@ impl DidUrl {
             })
             .collect()
     }
+
+    // TODO: Ideally we would have a builder instead of purpose-specific constructors
+    pub fn from_fragment(fragment: String) -> Result<Self, ParseError> {
+        // TODO: Better validation
+        if fragment.contains("#") {
+            return Err(ParseError::InvalidInput(
+                "Fragment cannot contain '#' character",
+            ));
+        }
+        if fragment.is_empty() {
+            return Err(ParseError::InvalidInput("Empty fragment"));
+        }
+        let len = fragment.len();
+        Ok(Self {
+            did_url: format!("#{}", fragment),
+            did: None,
+            method: None,
+            id: None,
+            path: None,
+            fragment: Some(1..len + 1),
+            queries: HashMap::new(),
+            params: HashMap::new(),
+        })
+    }
+
+    pub(crate) fn from_did_parts(
+        did_url: String,
+        did: DidRange,
+        method: DidRange,
+        id: DidRange,
+    ) -> Self {
+        Self {
+            did_url,
+            did: Some(did),
+            method: Some(method),
+            id: Some(id),
+            path: None,
+            fragment: None,
+            queries: HashMap::new(),
+            params: HashMap::new(),
+        }
+    }
 }
 
 impl TryFrom<String> for DidUrl {
--- a/did_peer/src/error.rs
+++ b/did_peer/src/error.rs
@@ -1,0 +1,69 @@
+use std::convert::Infallible;
+
+use did_doc::schema::verification_method::VerificationMethodType;
+use thiserror::Error;
+
+use crate::peer_did::Numalgo;
+
+#[derive(Debug, Error)]
+pub enum DidPeerError {
+    #[error("DID parser error: {0}")]
+    DidParserError(#[from] did_parser::ParseError),
+    #[error("DID validation error: {0}")]
+    DidValidationError(String),
+    #[error("DID document builder error: {0}")]
+    DidDocumentBuilderError(#[from] did_doc::error::DidDocumentBuilderError),
+    #[error("Invalid key reference: {0}")]
+    InvalidKeyReference(String),
+    #[error("Invalid service type")]
+    InvalidServiceType,
+    #[error("Sovrin DID document builder error: {0}")]
+    DidDocumentSovBuilderError(#[from] did_doc_sov::error::DidDocumentSovError),
+    #[error("Unsupported numalgo: {0}")]
+    UnsupportedNumalgo(Numalgo),
+    #[error("Invalid numalgo character: {0}")]
+    InvalidNumalgoCharacter(char),
+    #[error("Unsupported purpose character: {0}")]
+    UnsupportedPurpose(char),
+    #[error("Unsupported multicodec descriptor: {0}")]
+    UnsupportedMulticodecDescriptor(u64),
+    #[error("Unsupported verification method type: {0}")]
+    UnsupportedVerificationMethodType(VerificationMethodType),
+    #[error("Base 64 decoding error")]
+    Base64DecodingError(#[from] base64::DecodeError),
+    #[error("Multibase decoding error")]
+    MultibaseDecodingError(#[from] multibase::Error),
+    #[error("Varint decoding error")]
+    VarintDecodingError(#[from] VarintDecodingError),
+    #[error("JSON error: {0}")]
+    JsonError(#[from] serde_json::Error),
+    #[error("Regex error: {0}")]
+    RegexError(#[from] regex::Error),
+}
+
+impl From<Infallible> for DidPeerError {
+    fn from(_: Infallible) -> Self {
+        panic!("Attempted to convert an Infallible error")
+    }
+}
+
+#[derive(Debug, Error)]
+pub struct VarintDecodingError(unsigned_varint::decode::Error);
+
+impl std::fmt::Display for VarintDecodingError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Varint decoding error: {}", self.0)
+    }
+}
+
+impl From<unsigned_varint::decode::Error> for VarintDecodingError {
+    fn from(error: unsigned_varint::decode::Error) -> Self {
+        Self(error)
+    }
+}
+
+impl From<unsigned_varint::decode::Error> for DidPeerError {
+    fn from(error: unsigned_varint::decode::Error) -> Self {
+        Self::VarintDecodingError(error.into())
+    }
+}
--- a/did_peer/src/key/key.rs
+++ b/did_peer/src/key/key.rs
@@ -1,22 +1,22 @@
 use crate::error::DidPeerError;
 
-use super::SupportedKeyType;
+use super::KeyType;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Key {
-    key_type: SupportedKeyType,
+    key_type: KeyType,
     key: Vec<u8>,
 }
 
 impl Key {
-    pub fn new(key: Vec<u8>, key_type: SupportedKeyType) -> Result<Self, DidPeerError> {
+    pub fn new(key: Vec<u8>, key_type: KeyType) -> Result<Self, DidPeerError> {
         // If the key is a multibase key coming from a verification method, for some reason it is also
         // multicodec encoded, so we need to strip that. But should it be?
         let key = Self::strip_multicodec_prefix_if_present(key, &key_type);
         Ok(Self { key_type, key })
     }
 
-    pub fn key_type(&self) -> &SupportedKeyType {
+    pub fn key_type(&self) -> &KeyType {
         &self.key_type
     }
 
@@ -24,7 +24,7 @@ impl Key {
         self.key.as_ref()
     }
 
-    pub fn prefixed_key(&self) -> Vec<u8> {
+    pub fn multicodec_prefixed_key(&self) -> Vec<u8> {
         let code = self.key_type().into();
         let mut buffer = [0u8; 10];
         let bytes = unsigned_varint::encode::u64(code, &mut buffer);
@@ -34,18 +34,14 @@ impl Key {
     }
 
     pub fn fingerprint(&self) -> String {
-        multibase::encode(multibase::Base::Base58Btc, &self.prefixed_key())
-    }
-
-    pub fn prefixless_fingerprint(&self) -> String {
-        self.fingerprint().trim_start_matches('z').to_string()
+        multibase::encode(multibase::Base::Base58Btc, &self.multicodec_prefixed_key())
     }
 
     pub fn base58(&self) -> String {
         bs58::encode(&self.key).into_string()
     }
 
-    pub fn multibase(&self) -> String {
+    pub fn multibase58(&self) -> String {
         multibase::encode(multibase::Base::Base58Btc, &self.key)
     }
 
@@ -58,7 +54,7 @@ impl Key {
         })
     }
 
-    fn strip_multicodec_prefix_if_present(key: Vec<u8>, key_type: &SupportedKeyType) -> Vec<u8> {
+    fn strip_multicodec_prefix_if_present(key: Vec<u8>, key_type: &KeyType) -> Vec<u8> {
         if let Ok((value, remaining)) = unsigned_varint::decode::u64(&key) {
             if value == Into::<u64>::into(key_type) {
                 remaining.to_vec()
@@ -75,7 +71,7 @@ impl Key {
 mod tests {
     use super::*;
 
-    fn new_key_test(key_bytes: Vec<u8>, key_type: SupportedKeyType) {
+    fn new_key_test(key_bytes: Vec<u8>, key_type: KeyType) {
         let key = Key::new(key_bytes.clone(), key_type.clone());
         assert!(key.is_ok());
         let key = key.unwrap();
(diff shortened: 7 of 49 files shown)


swcurran:
FYI — We just did a quick clarification to the DID Peer Spec. to clarify that the padding characters of the Base64URL encoding of the service item MUST have the padding characters removed. The padding characters (“=“) are not permitted in a DID, per the peer DID spec.

did_parser/src/did.rs
```diff
@@ -91,3 +93,9 @@ impl<'de> Deserialize<'de> for Did {
         Self::parse(did).map_err(serde::de::Error::custom)
     }
 }
+
+impl From<Did> for DidUrl {
+    fn from(did: Did) -> Self {
+        Self::from_did_parts(did.did().to_string(), 0..did.did.len(), did.method, did.id)
```
bobozaur, line 99: I think the `did` field can be passed directly, even if the field is private, as this is in the same module with the struct.

did_peer/src/error.rs
```diff
@@ -0,0 +1,69 @@
+use std::convert::Infallible;
+
+use did_doc::schema::verification_method::VerificationMethodType;
+use thiserror::Error;
```
bobozaur, line 4: This is a very very very small thing, but IMO it's a good practice to alias stuff that are common in the `std`. Usually `Error` refers to the `std::error::Error` trait. Disregard if it's unimportant to you.

did_peer/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+    JsonError(#[from] serde_json::Error),
+    #[error("Regex error: {0}")]
+    RegexError(#[from] regex::Error),
+}
+
+impl From<Infallible> for DidPeerError {
+    fn from(_: Infallible) -> Self {
+        panic!("Attempted to convert an Infallible error")
+    }
+}
```
bobozaur, line 48: `Infallible` is usually handled by `unwrap` or `expect`. I'm curious as to why you implemented this.
If some poor soul unknowingly has this conversion somewhere, they will get a runtime panic. If this is removed, it's caught at compile time since the conversion cannot take place and *must* be handled.

I personally think it'd be best to remove it.

did_peer/src/key/key_type.rs
```diff
@@ -0,0 +1,46 @@
...
+            236 => Ok(SupportedKeyType::X25519),
+            237 => Ok(SupportedKeyType::Ed25519),
+            238 => Ok(SupportedKeyType::Bls12381g1g2),
+            4608 => Ok(SupportedKeyType::P256),
+            4609 => Ok(SupportedKeyType::P384),
+            4610 => Ok(SupportedKeyType::P521),
+            p @ _ => Err(DidPeerError::UnsupportedMulticodecDescriptor(p)),
+        }
+    }
+}
```
bobozaur, line 46: Might be worth considering making some constants for these `u64` ids. Maybe even just associated to the type.

did_peer/src/key/verification_method.rs
```diff
@@ -0,0 +1,228 @@
...
+        }
+        VerificationMethodType::Bls12381G1Key2020 => SupportedKeyType::Bls12381g1,
+        VerificationMethodType::Bls12381G2Key2020 => SupportedKeyType::Bls12381g2,
+        VerificationMethodType::X25519KeyAgreementKey2019 | VerificationMethodType::X25519KeyAgreementKey2020 => {
+            SupportedKeyType::X25519
+        }
+        t @ _ => return Err(DidPeerError::UnsupportedVerificationMethodType(t.to_owned())),
+    };
+    Key::new(vm.public_key().key_decoded()?, key_type)
+}
```
bobozaur, line 58: What do you think of making this a method on the `VerificationMethod` type?

did_peer/src/key/verification_method.rs
```diff
@@ -0,0 +1,228 @@
...
+    vm_type: VerificationMethodType,
+    key: &Key,
+    id: DidUrl,
+    did: Did,
+    public_key_encoding: &PublicKeyEncoding,
+) -> Vec<VerificationMethod> {
+    vec![VerificationMethod::builder(id, did, vm_type)]
+        .iter()
+        .map(|builder| add_public_key_to_builder(builder.to_owned(), key, public_key_encoding))
+        .collect::<Vec<VerificationMethod>>()
```
bobozaur, line 70: I'm curious as to why return a `Vec` here? Unless I'm wrong, this constructs a single verification method. Why not return that?

did_peer/src/key/verification_method.rs
```diff
@@ -0,0 +1,228 @@
...
+    .enumerate()
+    .map(|(index, builder)| {
+        let key = match index {
+            0 => &g1_key,
+            1 => &g2_key,
+            _ => unreachable!(),
+        };
+        add_public_key_to_builder(builder.to_owned(), key, public_key_encoding)
+    })
+    .collect::<Vec<VerificationMethod>>()
```
bobozaur, line 94: What do you think about first creating the individual verification methods and constructing the `Vec` *after*? This avoids constructing a 2 elements long `Vec` to just iterate over and consume, to create another `Vec`.

did_peer/src/peer_did/peer_did.rs
```diff
@@ -0,0 +1,146 @@
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct PeerDid {
+    did: Did,
+    numalgo: Numalgo,
+    transform: Option<Transform>,
+}
+
+impl PeerDid {
+    pub fn parse<T>(did: T) -> Result<Self, DidPeerError>
```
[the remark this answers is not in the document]
Patrik-Stas: But I've realized that if you'd want to call resolve method right here, the resolve method right now is parametrized with options, which would make things slightly trickier, especially if we would want to store the resolved DidDoc into `PeerDid` itself. Hmm.

Well I suppose the status quo is better after all.

did_peer/src/peer_did/mod.rs
```diff
@@ -0,0 +1,6 @@
+mod numalgo;
+mod peer_did;
+mod transform;
+
+pub use numalgo::Numalgo;
+pub use peer_did::PeerDid;
```
Patrik-Stas, line 6: keeping imports pretty 👍

did_peer/src/numalgos/numalgo2/generate/helpers.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    for cd in did_document.capability_delegation() {
+        did = append_encoded_key_segment(did, &did_document, cd, ElementPurpose::CapabilityDelegation)?;
+    }
+
+    Ok(did)
+}
+
+pub fn append_encoded_service_segment(
+    mut did: String,
```
bobozaur, line 48: Since this only appends, why not take a `&mut String`?

did_peer/src/numalgos/numalgo2/generate/helpers.rs
```diff
@@ -0,0 +1,256 @@
...
+    if let Some(service_encoded) = service_encoded {
+        let encoded = format!(".{}{}", ElementPurpose::Service, service_encoded);
+        did.push_str(&encoded);
+    }
+
+    Ok(did)
+}
+
+fn append_encoded_key_segment(
+    mut did: String,
```
bobozaur, line 76: Just like above, maybe it's worth considering taking a `&mut String`.

did_peer/src/numalgos/numalgo2/generate/helpers.rs
```diff
@@ -0,0 +1,256 @@
...
+use did_doc::schema::{did_doc::DidDocument, verification_method::VerificationMethodKind};
+use did_doc_sov::extra_fields::ExtraFieldsSov;
+
+use crate::{
+    error::DidPeerError,
+    key::get_key_by_verification_method,
+    numalgos::numalgo2::{purpose::ElementPurpose, service_abbreviated::ServiceAbbreviated},
+};
+
+pub fn append_encoded_key_segments(
```
Patrik-Stas, line 11: `pub(crate)` should suffice

did_peer/src/numalgos/numalgo2/generate/helpers.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    for cd in did_document.capability_delegation() {
+        did = append_encoded_key_segment(did, &did_document, cd, ElementPurpose::CapabilityDelegation)?;
+    }
+
+    Ok(did)
+}
+
+pub fn append_encoded_service_segment(
```
Patrik-Stas, line 47: dtto

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
```diff
@@ -0,0 +1,413 @@
...
+pub fn process_elements(
+    mut did_doc_builder: DidDocumentBuilder<ExtraFieldsSov>,
+    did: &Did,
+    public_key_encoding: &PublicKeyEncoding,
+) -> Result<DidDocumentBuilder<ExtraFieldsSov>, DidPeerError> {
+    let numalgoless_id = did.id().chars().skip(1).collect::<String>();
+    let mut service_index: usize = 0;
+
+    // Skipping one here because the first element is empty string
+    for element in numalgoless_id.split('.').skip(1) {
```
bobozaur, line 22: How about something like this? Avoids allocation, skips the first char like before *and* the first `.` that I assume is causing the empty string for the first element.

```suggestion
    let numalgoless_id = &did.id()[2..];
    let mut service_index: usize = 0;
    
    for element in numalgoless_id.split('.') {
```

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
```diff
@@ -0,0 +1,413 @@
...
+) -> Result<DidDocumentBuilder<ExtraFieldsSov>, DidPeerError> {
+    let purpose: ElementPurpose = element
+        .chars()
+        .nth(0)
+        .ok_or(DidPeerError::DidValidationError(format!(
+            "No purpose code following element separator in '{}'",
+            element
+        )))?
+        .try_into()?;
+    let purposeless_element = element.chars().skip(1).collect::<String>();
```
bobozaur, line 44:
```suggestion
    let purposeless_element = &element[1..];
```

did_peer/src/numalgos/numalgo2/generate/helpers.rs
Patrik-Stas: especially since this is a bit dangerous api - did is being mutated, but function can fail, leaving `did` in inconsistent state. But I think it's reasonable optimization as far as these functions stay local to the `generate` module

did_peer/src/numalgos/numalgo2/generate/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use self::helpers::{append_encoded_key_segments, append_encoded_service_segment};
+
+pub fn generate_numalgo2(did_document: DidDocument<ExtraFieldsSov>) -> Result<PeerDid, DidPeerError> {
+    let mut did = String::from("did:peer:2");
+
+    did = append_encoded_key_segments(did, &did_document)?;
+    did = append_encoded_service_segment(did, &did_document)?;
+
+    PeerDid::parse(did)
```
bobozaur, line 16: Perhaps this could be made more efficiently if the `PeerDid::parse_numalgo()` and `PeerDid::parse_transform()` would process the unprefixed strings (or just the chars, actually) instead of processing the entire `Did` for which the prefix is known to be correct since it's being added here.

did_peer/src/peer_did/numalgo.rs
```diff
@@ -0,0 +1,36 @@
+use std::fmt::Display;
+
+use crate::error::DidPeerError;
+
+#[derive(Clone, Debug, PartialEq)]
```
bobozaur, line 5: This can be `Copy` and returning it would be cheap instead of passing around references.

did_peer/src/peer_did/transform.rs
```diff
@@ -0,0 +1,20 @@
+use crate::error::DidPeerError;
+
+#[derive(Clone, Debug, PartialEq)]
```
bobozaur, line 3: Same as for `Numalgo`, this can be `Copy` and returning it would be cheap instead of passing around references.

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
```diff
@@ -0,0 +1,413 @@
...
+        did_doc_builder = process_service_element(&purposeless_element, did_doc_builder, service_index)?;
+    } else {
+        did_doc_builder =
+            process_key_element(&purposeless_element, did_doc_builder, did, public_key_encoding, purpose)?;
+    }
+
+    Ok(did_doc_builder)
+}
+
+fn process_service_element(
```
Patrik-Stas, line 56: same note about naming, i'd suggest "mapping" or "encoding" vocabulary rather than "processing" across the board in this file

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
```diff
@@ -0,0 +1,413 @@
...
+    let id = format!("#{}-{}", service_type.to_lowercase(), index).parse()?;
+
+    if service.routing_keys().is_empty() {
+        build_service_aip1(service, id, service_type)
+    } else {
+        build_service_didcommv2(service, id, service_type)
+    }
+}
+
+fn build_service_aip1(
```
Patrik-Stas, line 125: i'd suggest to keep the "deabbreviate" vocabulary, eg `deabbreviate_service_aip1`

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
```diff
@@ -0,0 +1,413 @@
...
+    service: ServiceAbbreviated,
+    id: Uri,
+    service_type: String,
+) -> Result<Service<ExtraFieldsSov>, DidPeerError> {
+    Ok(Service::<ExtraFieldsSov>::builder(
+        id,
+        service.service_endpoint().parse()?,
```
Patrik-Stas, line 132: why not to return `Url` from `service_endpoint()`? Then we wouldn't have re-parse what we already are sure is valid url

did_peer/src/numalgos/numalgo2/service_abbreviated.rs
Patrik-Stas, line 1: the process of abbreviation is in in `generate/helpers.rs`, de-abbreviation is here. I'd prefer having them side by side.

Secondly de-abbreviation is using custom methods, abbreviation is implemented as `impl TryFrom<&Service<ExtraFieldsSov>> for ServiceAbbreviated`I'd suggest to use the same approach for both

Patrik-Stas (changes requested):

did_resolver/src/traits/resolvable/mod.rs
```diff
@@ -11,11 +11,12 @@ use self::{resolution_options::DidResolutionOptions, resolution_output::DidResol
 
 #[async_trait]
 pub trait DidResolvable {
-    type ExtraFields: Default;
+    type ExtraFieldsService: Default;
+    type ExtraFieldsOptions: Default;
```
bobozaur, line 15: Why is `Default` required here? You can already pass dummy ZST types like `()` or whatever else to "avoid" the additional fields. The `Default` seems like having a risk of having to implement it for the type you want to pass, which might create it in an unusable state, which you have to take care of or you end up with garbage fields.
Or am I missing something?

bobozaur (changes requested):

did_doc_sov/src/extra_fields/didcommv2.rs
```diff
@@ -34,10 +46,24 @@ impl ExtraFieldsDidCommV2Builder {
         self
     }
 
+    pub fn set_accept<I>(mut self, accept: I) -> Self
+    where
+        I: IntoIterator<Item = AcceptType>,
```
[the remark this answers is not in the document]
bobozaur: I for one am against this. There's a concrete type that's needed here, which consumers can convert to themselves if needed. Abstracting over that doesn't help with anything other than unnecessarily doing the conversion sometimes.

did_parser/src/did_url.rs
```diff
@@ -146,6 +146,48 @@ impl DidUrl {
...
+            method: None,
+            id: None,
+            path: None,
+            fragment: Some(1..len + 1),
+            queries: HashMap::new(),
+            params: HashMap::new(),
+        })
+    }
+
+    pub(crate) fn from_did_parts(
```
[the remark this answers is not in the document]
bobozaur: Or directly move this to the `From` impl as the ranges are known to be valid there.

did_doc_sov/src/extra_fields/didcommv1.rs
```diff
@@ -56,12 +68,25 @@ impl ExtraFieldsDidCommV1Builder {
         self
     }
 
+    pub fn set_accept<I>(mut self, accept: I) -> Self
+    where
+        I: IntoIterator<Item = AcceptType>,
+    {
+        self.accept = accept.into_iter().collect();
+        self
+    }
```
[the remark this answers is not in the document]
mirgee: When working on the PR I made `accept` a `HashMap` and this is a leftover. I will remove it. Not sure about the overhead, I would expect it would not be hard for the compiler to remove these redundant calls.

did_peer/src/key/verification_method.rs
mirgee: See the callsite.

did_peer/src/numalgos/numalgo2/generate/helpers.rs
mirgee: I strongly prefer immutability wherever possible, and following the principle of least surprise, i.e. in this case taking the old version by value and returning new version of the string instead of mutating the argument "silently" in place.

did_peer/src/key/verification_method.rs
bobozaur: There seems to be just one, the `get_verification_methods_by_key` function. Why not make it a `Vec` there? Or at least not construct a single element `Vec` just to consume and iterate over it, then collect it in a newly allocated `Vec`.

did_peer/src/numalgos/numalgo2/generate/helpers.rs
bobozaur: I'd argue that it is less surprising to pass in a `&mut String` and realize it will get mutated rather than passing in an owned `String` just to get it back after it gets mutated. Because whichever way it's done, the very same heap memory location *does* get mutated.

did_resolver/src/traits/resolvable/mod.rs
mirgee: I was also debating whether to get rid of it, thinking of the same argument. It's not there to "avoid" additional fields, but more for convenience. Speaking of `ExtraFieldsOptions`, I don't see the problem of unusable state as present for the current use-case, but in general it is there. I will remove the bound from `ExtraFieldsOptions`, removing it from `ExtraFieldsService` is out of scope of this PR.

did_peer/src/key/verification_method.rs
mirgee: Alright, I don't have a strong preference here.

did_resolver/src/traits/resolvable/mod.rs
bobozaur: Wasn't it added in `ExtraFieldsService` in this PR, though? May I ask why is it outside of scope?

did_peer/src/key/verification_method.rs
bobozaur: Please at least make the initial container an array...

did_peer/src/numalgos/numalgo2/generate/helpers.rs
mirgee: I don't see how the fact that the very same heap memory location gets mutated both ways influences the degree of surprise, or how is it argument either way. What I meant by surprise to avoid are the potential side-effects related to passing around mutable references (making it unclear where the actual value gets mutated). Again, I prefer passing ownership and returning new value - it avoids creating mutable references and side-effects, it clearly establishes ownership of the value and the places where it is modified.

did_resolver/src/traits/resolvable/mod.rs
mirgee: It wasn't, as you can see in the diff.

did_peer/src/error.rs
mirgee: No, I agree. Just missed this, thanks for pointing it out.

did_peer/src/key/verification_method.rs
mirgee: The method returns `Key`, which is defined currently in `did_peer`, and both are only used in `did_peer`. I am not convinced that `did_peer` is the ideal place for either of them, but put them here for now since they are not used anywhere else.

did_peer/src/key/verification_method.rs
mirgee: I have no issue with that, sounds reasonable.

did_peer/src/error.rs
mirgee: `TryFrom<Did>` (used in `PeerDid::parse`) is infallible for `Did`, but may not be infallible otherwise, so can't unwrap on the callsite.

did_resolver/src/traits/resolvable/mod.rs
bobozaur: Ah, my bad, the type was renamed but the `Default` was there.

did_peer/src/error.rs
bobozaur: Oh, I see. Then it does make sense. Thanks for clarifying this!

did_peer/src/key/verification_method.rs
bobozaur: Hm, fair.

did_peer/src/numalgos/numalgo2/generate/helpers.rs
bobozaur: Alright, your PR, your call.

did_doc/src/schema/verification_method/public_key.rs
```diff
@@ -0,0 +1,112 @@
...
+};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(untagged)]
+#[serde(deny_unknown_fields)]
+pub enum PublicKeyField {
+    #[serde(rename_all = "camelCase")]
+    Multibase { public_key_multibase: String },
+    #[serde(rename_all = "camelCase")]
+    Jwk { public_key_jwk: JsonWebKey },
```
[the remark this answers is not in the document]
mirgee: Do you mean JWK specifically? That one is mentioned [here](https://www.w3.org/TR/did-core/#verification-method-properties), some others are mentioned [here](https://www.w3.org/TR/did-spec-registries/#verification-method-properties).

did_peer/src/key/key_type.rs
```diff
@@ -0,0 +1,46 @@
...
+    Bls12381g1g2,
+    Bls12381g1,
+    Bls12381g2,
+    X25519,
+    P256,
+    P384,
+    P521,
+}
+
+impl From<&SupportedKeyType> for u64 {
```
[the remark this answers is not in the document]
mirgee: Yeah, I agree, I will include a link to the table of multicodec descriptors

did_doc_sov/src/extra_fields/didcommv2.rs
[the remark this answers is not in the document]
mirgee: As I said elsewhere, this is a remnant of originally trying to use `HashSet` only to realize that order may actually matter. I don't see an issue with leaving it abstracted like this as it gives flexibility and has no downsides, but I have no issue with removing it either.

did_parser/src/did_url.rs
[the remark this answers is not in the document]
mirgee: I would expect this trait to be implemented for `DidUrl`, as `Did` can conceptually be infallibly converted to `DidUrl`, as a valid pure `Did` IS also a valid `DidUrl`. The method needs access to the ranges of `Did` and needs to set the ranges of newly created `DidUrl`, both of which are private fields. Arguably adding crate-visible getters for ranges of `Did` is less "dangerous" than adding a crate-visible constructor with ranges for `DidUrl`, so this is what I'll do, moving this trait implementation to the `did_url` module.

did_peer/src/key/key.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Key {
+    key_type: SupportedKeyType,
+    key: Vec<u8>,
+}
+
+impl Key {
+    pub fn new(key: Vec<u8>, key_type: SupportedKeyType) -> Result<Self, DidPeerError> {
+        // If the key is a multibase key coming from a verification method, for some reason it is also
```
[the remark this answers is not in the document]
mirgee: This is a point of confusion for me as well, so it will be good to have another pair of eyes (with a brain attached) taking a look at it. 👍

did_peer/src/key/key_type.rs
```diff
@@ -0,0 +1,46 @@
+use crate::error::DidPeerError;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum SupportedKeyType {
```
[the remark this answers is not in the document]
mirgee: Yeah, I was thinking the same - I guess the `Supported` part is implied.

did_peer/src/key/key.rs
```diff
@@ -0,0 +1,244 @@
+use crate::error::DidPeerError;
+
+use super::SupportedKeyType;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Key {
```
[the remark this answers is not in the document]
mirgee: There is not such intention, but on the other hand, there is no restriction currently either. But `PublicKey` is more descriptive and correct for all current use-cases, so I have no issue with renaming it.

did_peer/src/key/key.rs
```diff
@@ -0,0 +1,244 @@
...
+    key_type: SupportedKeyType,
+    key: Vec<u8>,
```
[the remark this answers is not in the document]
mirgee: Hmm... If you don't like `key` (I can see why - it too generic), I would suggest `key_bytes`, for example?

did_peer/src/key/key.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    pub fn key_type(&self) -> &SupportedKeyType {
+        &self.key_type
+    }
+
+    pub fn key(&self) -> &[u8] {
+        self.key.as_ref()
+    }
+
+    pub fn prefixed_key(&self) -> Vec<u8> {
```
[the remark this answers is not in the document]
mirgee: This is a [multicodec](https://github.com/multiformats/multicodec) prefix. The multicodec codes are encoded as unsinged varints (variable-width integers). The [maximum practical size of a varint is 63 bits](https://github.com/multiformats/unsigned-varint#practical-maximum-of-9-bytes-for-security), which is 9 bytes (the first bit of each byte is a continuation bit), which about fits into `u64`. I don't know why the length of the buffer 10 instead of 9, but it's a constant defined by the library.

did_peer/src/key/key.rs
```diff
@@ -0,0 +1,244 @@
...
+        let code = self.key_type().into();
+        let mut buffer = [0u8; 10];
+        let bytes = unsigned_varint::encode::u64(code, &mut buffer);
+        let mut prefixed_key = bytes.to_vec();
+        prefixed_key.extend_from_slice(&self.key);
+        prefixed_key
+    }
+
+    pub fn fingerprint(&self) -> String {
```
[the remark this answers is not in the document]
mirgee: I agree the naming is dubious. I will remove the current `multibase` method (seems unused) and call this function multibase - what do you think? One slight issue with that is that multibase-encoded key will contain the multicodec prefix, whereas base58 (for example) will not. But then again, perhaps that's expected behavior (as per the values in verification methods)...

did_peer/src/key/key.rs
```diff
@@ -0,0 +1,244 @@
...
+        multibase::encode(multibase::Base::Base58Btc, &self.prefixed_key())
+    }
+
+    pub fn prefixless_fingerprint(&self) -> String {
```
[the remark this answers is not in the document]
mirgee: OK, see no problem with that

did_peer/src/key/key.rs
```diff
@@ -0,0 +1,244 @@
...
+        self.fingerprint().trim_start_matches('z').to_string()
+    }
+
+    pub fn base58(&self) -> String {
+        bs58::encode(&self.key).into_string()
+    }
+
+    pub fn multibase(&self) -> String {
```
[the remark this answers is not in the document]
mirgee: I see, you noticed 😄 I will resolve this to keep the discussion one place

did_peer/src/peer_did/peer_did.rs
```diff
@@ -0,0 +1,146 @@
...
+use did_doc_sov::extra_fields::ExtraFieldsSov;
+use did_parser::Did;
+use once_cell::sync::Lazy;
+use regex::Regex;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use super::{numalgo::Numalgo, transform::Transform};
+
+// TODO: This regex does not cover peer:did:3
+static PEER_DID_REGEX: Lazy<Regex> = Lazy::new(|| {
```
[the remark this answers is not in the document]
mirgee: OK

did_peer/src/peer_did/peer_did.rs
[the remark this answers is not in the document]
mirgee: Isn't the regex sufficient to validate the DID? Moreover, you might want to use the DID without resolving it, so resolving it when parsing seems redundant. Also, resolution is numalgo-specific and may not be possible (e.g. numalgo 3).

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
```diff
@@ -0,0 +1,149 @@
...
+use did_parser::Did;
+
+use crate::{
+    error::DidPeerError,
+    key::{get_verification_methods_by_key, Key},
+    numalgos::numalgo2::{purpose::DidPeerPurpose, service_abbreviated::ServiceAbbreviated},
+    peer_did_resolver::options::PublicKeyEncoding,
+};
+
+pub fn process_elements(
```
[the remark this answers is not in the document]
mirgee: There is no need, this entire module is not visible outside `numalgo2` module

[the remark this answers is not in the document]
mirgee: What do you mean? This just passes the ownership of `did_doc_builder` to the builder and declares it mutable (which is necessary to change its value). Returning the builder adds flexibility to the caller to adjust the DDO before building (e.g. adding `@context`, etc.).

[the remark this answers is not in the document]
mirgee: I am not sure about this suggestion honestly. The current name seems appropriate to me given the context - "element" is actually a term from the spec, so assuming good knowledge of the spec, "processing" of "elements" is an understandable concept following quite naturally.

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
```diff
@@ -0,0 +1,149 @@
...
+    mut did_doc_builder: DidDocumentBuilder<ExtraFieldsSov>,
+    did: &Did,
+    public_key_encoding: &PublicKeyEncoding,
+) -> Result<DidDocumentBuilder<ExtraFieldsSov>, DidPeerError> {
+    let numalgoless_id = did.id().chars().skip(1).collect::<String>();
```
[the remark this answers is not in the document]
mirgee: Good point!

did_peer/src/key/verification_method.rs
```diff
@@ -0,0 +1,228 @@
+use did_doc::schema::verification_method::{
+    IncompleteVerificationMethodBuilder, VerificationMethod, VerificationMethodType,
+};
+use did_parser::{Did, DidUrl};
+
+use crate::{error::DidPeerError, peer_did_resolver::options::PublicKeyEncoding};
+
+use super::{Key, SupportedKeyType};
+
+pub fn get_verification_methods_by_key(
```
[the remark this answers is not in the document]
mirgee: I agree it's not immediately obvious what this function does (from the name), but the alternatives seem quite equivalent to me?

did_peer/src/key/verification_method.rs
```diff
@@ -0,0 +1,228 @@
...
+    Ok(build_verification_methods_from_type_and_key(
+        vm_type,
+        key,
+        id,
+        did.to_owned(),
+        public_key_encoding,
+    ))
+}
+
+pub fn get_key_by_verification_method(vm: &VerificationMethod) -> Result<Key, DidPeerError> {
```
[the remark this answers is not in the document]
mirgee: I will resolve this to keep discussion one place.

did_peer/tests/fixtures/no_services.rs
```diff
@@ -0,0 +1,29 @@
+pub static PEER_DID_NUMALGO_2_NO_SERVICES: &str = "did:peer:2\
+.Ez6LSpSrLxbAhg2SHwKk7kwpsH7DM7QjFS5iK6qP87eViohud\
+.Vz6MkqRYqQiSgvZQdnBytw86Qbs2ZWUkGv22od935YF4s8M7V";
+
+pub static DID_DOC_NO_SERVICES: &str = r##"
+    {
+        "id": "did:peer:2.Ez6LSpSrLxbAhg2SHwKk7kwpsH7DM7QjFS5iK6qP87eViohud.Vz6MkqRYqQiSgvZQdnBytw86Qbs2ZWUkGv22od935YF4s8M7V",
+        "verificationMethod": [
+            {
+                "id": "#6MkqRYqQ",
```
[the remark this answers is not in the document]
mirgee: I think it should be a valid relative DID URL, will take a closer look at the spec.

did_peer/tests/resolve_negative.rs
```diff
@@ -0,0 +1,89 @@
+mod fixtures;
+
+use did_peer::peer_did_resolver::{
+    options::{ExtraFieldsOptions, PublicKeyEncoding},
+    resolver::PeerDidResolver,
+};
+use did_resolver::traits::resolvable::{resolution_options::DidResolutionOptions, DidResolvable};
+use tokio::test;
+
+macro_rules! resolve_negative_test {
```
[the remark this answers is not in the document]
mirgee: Hmm, I tend to agree macros should be used judiciously.

did_peer/tests/resolve_negative.rs
```diff
@@ -0,0 +1,89 @@
...
+            let resolver = PeerDidResolver::new();
+            let options = DidResolutionOptions::new()
+                .set_extra(ExtraFieldsOptions::new().set_public_key_encoding(PublicKeyEncoding::Multibase));
+            let ddo_result = resolver.resolve(&$peer_did.parse().unwrap(), &options).await;
+            assert!(ddo_result.is_err());
+        }
+    };
+}
+
+resolve_negative_test!(
```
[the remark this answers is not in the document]
mirgee: I agree. It was just faster to write the tests like this initially. Will adjust.

did_peer/src/numalgos/numalgo2/generate/helpers.rs
mirgee: Again, this module is not visible outside numalgo2.

did_peer/src/numalgos/numalgo2/generate/helpers.rs
mirgee: Closing to keep discussion one place.

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
mirgee: Hmm, as I said, I prefer the current naming of "processing" of "elements", but of course my position is biased by coming up and being very used to this naming. Again, the notion "processing" of "elements" seems obvious to me given what is expected to resolve a `did:peer:2`.

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
mirgee: Sounds reasonable.

did_doc/src/schema/verification_method/public_key.rs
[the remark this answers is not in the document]
Patrik-Stas: ah yeah, rookie mistake, I think I was searching with snake case. Thanks

did_peer/src/numalgos/numalgo2/service_abbreviated.rs
mirgee: Good catch, I originally intended to have deabbreviation implemented in this module.

did_peer/src/key/key.rs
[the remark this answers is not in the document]
Patrik-Stas: I was concerned that the accessing the Vec data would look awkward eg `key.key`, but just realized these are private, so that probably won't occur, as it will be rather accessed via `self`.
So that being said, I think am fine with `key`, or `key_bytes` as well

did_peer/src/key/key.rs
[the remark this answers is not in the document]
Patrik-Stas: Thank you for clarifying, then how about calling it explicitly `multicoded_prefixed_key`? Or else leaving a comment about what sort of prefix it is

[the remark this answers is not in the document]
mirgee: Sounds good

did_peer/src/key/key.rs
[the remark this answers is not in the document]
Patrik-Stas: sounds good, but with slight modification - I'd call the method `multibase_58`, as multibase can be in any base, but this function is producing multibase encoding specifically with base of 58.

did_peer/src/peer_did/peer_did.rs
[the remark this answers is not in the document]
Patrik-Stas: well, if some of the key fragments in peer-did did is nonsense, altough it is still a did, it is not a peer did. The only way to check that peer-did is valid peer-did is to actually resolve it (which is technically just decoding) (at least for numalgo 2)

Anyway, you can resolve this thread (no pun intended :-D)

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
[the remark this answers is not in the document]
Patrik-Stas: I see, if that so elements is probably good to keep, but I don't love "processing", it's quite generic, while really it's esentially mapping/transforming the elements onto diddoc. 

Just my opinion, but I am leaving up to your judgement to decide

did_peer/src/peer_did/peer_did.rs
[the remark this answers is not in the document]
mirgee: I guess one might distinguish valid unresolvable peer dids from invalid peer dids... But I see what you mean. I would leave the user to try resolving the DID in addition to parsing it instead of resolving it for him when parsing (which seems a little unexpected). Also, there is the issue with resolution options... Resolving - the thread for now :).

did_peer/src/key/verification_method.rs
```diff
@@ -0,0 +1,228 @@
...
+    did: Did,
+    public_key_encoding: &PublicKeyEncoding,
+) -> Vec<VerificationMethod> {
+    vec![VerificationMethod::builder(id, did, vm_type)]
+        .iter()
+        .map(|builder| add_public_key_to_builder(builder.to_owned(), key, public_key_encoding))
+        .collect::<Vec<VerificationMethod>>()
+}
+
+fn build_verification_methods_from_bls_multikey(
```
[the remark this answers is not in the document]
Patrik-Stas: Shall we remote the related code then?

did_peer/tests/fixtures/no_services.rs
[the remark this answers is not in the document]
Patrik-Stas: You are right, relative DID URL can be transformed to absolute DID URL https://www.w3.org/TR/did-core/#relative-did-urls. So I suppose the referred code 
```
"#6MkqRYqQ"
```
is technically the same as
```
"did:peer:2.Ez6LSpSrLxbAhg2SHwKk7kwpsH7DM7QjFS5iK6qP87eViohud.Vz6MkqRYqQiSgvZQdnBytw86Qbs2ZWUkGv22od935YF4s8M7V#6MkqRYqQ",
```

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
Patrik-Stas: leaving final decision to your judgement

did_peer/src/key/verification_method.rs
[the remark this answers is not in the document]
mirgee: The issue is other frameworks may use it. It would be nice to support as many key types as possible, and I reckon it shouldn't be too difficult to support `Bls12381g1g2` in particular. How about we leave / finish the implementation and add tests?

did_peer/src/key/verification_method.rs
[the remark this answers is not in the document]
mirgee: Hmm, I see...

did_peer/src/peer_did/numalgo.rs
mirgee: Can you please clarify for me - is the argument that cloning would be _cheaper_ than passing a reference?

did_peer/src/error.rs
mirgee: Actually, I realized you are not talking about an error type, but the thiserror derive macro implementing `std::error::Error` trait. I think this is fine, it's clear from the context that it is a macro, not a trait, so no confusion with `std::error::Error` should occur, and `thiserror` documentation imports the macro in the same way.

did_peer/src/numalgos/numalgo2/resolve/helpers.rs
mirgee: This function isn't  deabbreviating the fields of `ServiceAbbreviated`, so I think the current naming is more accurate.

did_peer/src/key/key.rs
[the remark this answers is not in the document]
mirgee: My concern with this is that to be consistent I should replace all mentions of the concept of `key` with `public_key`, e.g. 
* `append_encoded_key_segment` -> `append_encoded_public_key_segment`,
* `process_key_element` -> `process_public_key_element`,
* `build_verification_methods_from_type_and_key` -> `build_verification_methods_from_type_and_public_key`,
* `get_key_by_verification_method` -> `get_public_key_by_verification_method`,
* ...
thus making some names overly verbose.

did_parser/src/did_url.rs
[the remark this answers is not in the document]
mirgee: Actually, I don't think this is an issue. It's a crate internal method, so it's not enabling the client to shoot themselves in the foot. It's here only to simplify the construction of `DidUrl` from ranges which are only accessible from the `did` module.

did_peer/src/numalgos/numalgo2/service_abbreviated.rs
mirgee: Actually, since abbreviation is done only when generating a peer did and deabbreviation is done only when resolving, and thus `resolve` does not need to have deabbreviation accessible and `generate` does not need to have abbreviation accessible, I moved the conversion to their respective modules for what I think is a better separation of concerns. The module `service_abbreviated` now has responsibility of defining the fields of `ServiceAbbreviated`, defining the way of (de)serializing and constructing it.

did_peer/src/numalgos/numalgo2/generate/mod.rs
mirgee: I am sorry, I don't think this micro-optimization is worth it in the slightest. Also, from the perspective of `PeerDid`, the prefix is not known to be correct.

did_peer/src/key/key.rs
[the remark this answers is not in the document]
Patrik-Stas: okay, sounds reasonable

did_peer/src/key/verification_method.rs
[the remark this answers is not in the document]
Patrik-Stas: if our conterparty uses this type of keys, we will anyway won't be able to work with it right now. We don't enctry/decrypt/sign/verify for this type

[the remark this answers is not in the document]
mirgee: I know, but shouldn't we? That's what I am asking.

Patrik-Stas (approved):

did_peer/src/key/key.rs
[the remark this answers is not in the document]
mirgee: Actually, this implementation relies on the base of `fingerprint`, which is internal to `Key` (i.e. may change). Also, there are multiple places where it can be used.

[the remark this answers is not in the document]
Patrik-Stas: roger that!

bobozaur (approved):

Patrik-Stas (approved):

PR #887, opened 2023-06-27
Make PeerDid generic over numalgo

mirgee:
Makes the peer did submethod (numalgo) part of the `PeerDid` type, allowing for statically ensuring

* its properties (e.g. resolvable to a DDO, ability to be shortened to did:peer:3, etc.) or
* concrete value of the numalgo underlying the peer did.

`GenericPeerDid` enum was added to cover situations when there is no concrete numalgo expected.


Patrik-Stas (dismissed):
left minor comment, generally lgtm

did_peer/src/peer_did/peer_did/generic.rs
```diff
@@ -0,0 +1,144 @@
...
+        validate::validate,
+    },
+};
+use did_parser::Did;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use super::PeerDid;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum GenericPeerDid {
```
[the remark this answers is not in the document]
mirgee: Would it be useful? We are not supporting generation / resolution of `Numalgo1`, or `Numalgo2`, so the idea is to just disallow deserializing to them (they can't be subsequently resolved anyway). I guess one may argue that adding them later may require changing exhaustive matches, so perhaps I should add `#[non_exhaustive]` attribute?

did_peer/src/peer_did/peer_did/generic.rs
```diff
@@ -0,0 +1,144 @@
...
+    Numalgo2(PeerDid<Numalgo2>),
+    Numalgo3(PeerDid<Numalgo3>),
+}
+
+impl GenericPeerDid {
```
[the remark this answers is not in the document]
mirgee: I would hope that (the current) `PeerDid` will be used as a go to and `GenericPeerDid` will be used when `PeerDid<N>` can't be used for some reason, so that's why I gave them shorter and longer name respectively.

did_peer/src/peer_did/peer_did/generic.rs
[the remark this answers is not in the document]
Patrik-Stas: reasonable

did_peer/src/peer_did/peer_did/generic.rs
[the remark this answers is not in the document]
Patrik-Stas: I see, alright then

did_peer/src/peer_did/numalgos/traits.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn resolve(
+        &self,
+        did: &Did,
+        public_key_encoding: PublicKeyEncoding,
+    ) -> Result<DidDocument<ExtraFieldsSov>, DidPeerError>;
+}
+
+pub trait ToNumalgo3: Numalgo {
+    fn to_numalgo3(&self, did: &Did) -> Result<PeerDid<Numalgo3>, DidPeerError>;
+}
```
bobozaur, line 50: Could you provide some info about this trait? From what I see it's implemented for `Numalgo2` and `Numalgo3`, taking `&self` by reference but never really using it and always returning `PeerDid<Numalgo3>`.

I'm curious because I personally find that:
1) The name is a bit misleading -> You're not transforming between `Numalgo` variants, you actually return a `PeerDid`.
2) The signature seems weird given the implementations of the trait.
3) The purpose of the trait is unclear. Given that all of these are unit types, this could just as easily be a marker trait used for the trait bound in `impl<N: ToNumalgo3> PeerDid<N>` and have the actual implementation happen there.

did_peer/src/peer_did/peer_did/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+                match N::parse(value.to_string()) {
+                    Ok(peer_did) => Ok(peer_did),
+                    Err(err) => Err(E::custom(format!("Failed to parse numalgo: {err}"))),
+                }
+            }
+        }
+
+        deserializer.deserialize_str(PeerDidVisitor(PhantomData))
+    }
+}
```
bobozaur, line 84: The `CowStr` type from `shared_vcx` could be used here for less boilerplate deserialization. If the input can be borrowed from, it will borrow it and `.as_ref()` could be used to get a `&str`. Otherwise it will make an owned variant of the data, which again can be turned into a `&str` through the same method.

bobozaur (commented):
Generally LGTM, but I left a few comments with my opinions.

did_peer/src/peer_did/numalgos/traits.rs
mirgee: Implementors of the trait can be converted to peer:did:3. This is true for peer:did:2 since it can be shortened to peer:did:3 and for peer:did:3 because it already is peer:did:3. In some cases you don't care which numalgo you have as long as it can be converted to 3, e.g. querying a database for a record stored by peer:did:3.

bobozaur: Yeah, on a second look I got mixed up between the implementations and they behave differently so a marker trait won't suffice. My bad, sorry. 

I'd still drop the `&self` in the signature though. 

Sealing it is definitely to be done, but I wouldn't honestly worry about that too much, at least not right now. I skipped sealing some stuff in `messages` and the new `Connection` myself as I was more focused on the actual implementation than a consumer randomly implementing a trait 😅 .

Definitely not proud of that now that I think about it, so I will revisit it soon 😄 .

did_peer/src/peer_did/peer_did/mod.rs
mirgee: We already had a discussion about bringing in dependencies unnecessarily, and you know my opinion on it.

did_peer/src/peer_did/numalgos/traits.rs
mirgee: Removed `&self` from signature of `to_numalgo3`.

bobozaur (approved):

Patrik-Stas (approved):

PR #888, opened 2023-06-28
ci: run libvcx tests with modular features

Patrik-Stas:
### `libvcx_core`
- Modify global state handling and init related functions to enable running both with legacy vdrtools and credx/indy-vdr setup

### `aries-vcx`
- tweak indyvdr error mapping

### New cargo features in `libvcx_core`, `libvcx`, `vcx-napi-rs`
- Added features `anoncreds_vdrtools` `anoncreds_credx` to select anoncreds implementation 
- Added features `ledger_vdrtools`, `ledger_indyvdr` to select ledger client implementation

### CI
- Leave only 1 code coverage job which runs both integration and unit tests in `aries-vcx`. The coverage jobs runs with modular libs profile


bobozaur (approved):
LGTM, even though I'm not a fan of the separate features for every primitive thing, they will make transitioning to `anoncreds-rs` only easier.

PR #894, opened 2023-07-07
Update dependencies and AriesMessage Deserialize impl

### LINKED ISSUE #893: Recent versions of serde (e.g. 1.0.167) cause messages crate to not compile
```
impl<'de> Deserialize<'de> for AriesMessage {
```

within `messages/src/lib.rs` uses internal structures from `serde`, which exposes use to breaking internal changes from serde.

it seems that in serde 1.0.167 has had a change in their return value for `deserializer.deserialize_any`, from a `TaggedContent` struct, to a tuple `(CowStr<'_>, Content<'_>)`. so previous calls to `tagged.content` and `tagged.tag` are no longer valid. 

Quick fix should be to change the impl to `tagged.1` and `tagged.0` i believe.. however maybe it's a good opportunity to investigate if the internal serde APIs really need to be used, as this is an issue that may keep occurring on serde minor version bumps.

Found this issue after `cargo update`ing a crate that consumes `messages` (it auto updated `serde` to 167 since it does not expect breaking changes from minor bumps).

bobozaur:
Fixes #893.

Also updates `transitive` dependency and simplifies the `Deserialize` impl for `AriesMessage` to require less of `serde`'s internal components. 

Now only `Content` and `ContentDeserializer` are used merely at a type level, and these have been around for a very long time and are pretty much stable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -445,7 +445,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.16",
+ "syn 2.0.23",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


gmulhearn (approved):
CI is happy, I'm happy

PR #895, opened 2023-07-07
Added capability of migrating wallet through the node.js wrapper

bobozaur:
Exposes the wallet migration functionality in the Node.js wrapper


Patrik-Stas (changes requested):
This is not yet usable in practice - libvcx, unlike aries-vcx, is singletonish about everything, there's concept of "main wallet", "main pool", "main agency client" etc.

While technically you have access to wallet handle (`open_as_main_wallet` returns it), there's no place we need nodejs consumers to actually work with (any) handles. So to stay consistent, I would rather prefer to have something like "migrate_main_wallet(destWalletHandle)" with expectation that the user must call `open_as_main_wallet` first, and that's gonna be the source wallet.

The problem with `migrate_main_wallet(destWalletHandle)` however is that you don't have any way to get the `destWalletHandle`, because of singletonish nature of `libvcx`.
So a solution for that would be to have interface `migrate_main_wallet(wallet_config, wallet_credentials)` on `vcx-napi-rs` level, and also `libvcx_core` level, then proceed and within that function open the destination wallet and provide the handle to the migrator itself.

It would be good to include simple unit test - create wallet and migrate it (even if there's no records) just to make sure that the basic flow works and the API can be worked with.

Patrik-Stas (approved):

PR #897, opened 2023-07-13
Refactor dealing with tails_dir in tests

Patrik-Stas:
- Refactor tests (way of creating tails directory)
- Clear up confusion calling variables `tails_file` what actually is `tails_directory`
- Remove duplicated method for creating tmp directories and files

CI: 
- Small adjustment, when building android docker image, in case of cache miss, do not try to load up `main` branch revision of that image (this is happening to optimize CI in order to reuse some docker layers)


aries_vcx_core/src/indy/ledger/pool.rs
```diff
@@ -69,7 +69,11 @@ pub mod test_utils {
 
     use super::*;
 
-    pub fn get_temp_dir_path(filename: &str) -> PathBuf {
+    pub fn get_temp_dir_path() -> PathBuf {
```
mirgee, line 72: It seems like we have this function defined twice, it would be nice (and should be easy) to remove the duplication.

aries_vcx/tests/test_creds_proofs.rs
```diff
@@ -313,7 +312,7 @@ mod integration_tests {
     #[ignore]
     async fn test_agency_pool_generate_proof() {
         SetupProfile::run(|mut setup| async move {
-            create_and_store_credential(
+            let (_, _, _, _, _, _, _, _, _, _, tails_dir, _) = create_and_store_credential(
```
mirgee, line 315: Oh, good ol' functions returning 12 values in a tuple, always make me smile (and / or cry) :)

.github/actions/build-image/action.yml
```diff
@@ -54,11 +57,15 @@ runs:
...
         if ! docker pull "$LAST_BRANCH_BUILD"; then
-          LAST_MASTER_BUILD="$REMOTE_DOCKER_REPOSITORY:${{ inputs.branch-main }}"
-          echo "Could not pull $LAST_MASTER_BUILD, will try last master build $LAST_MASTER_BUILD"
-          if ! docker pull "$LAST_MASTER_BUILD"; then
-            echo "Could not pull $LAST_MASTER_BUILD either ¯\_(ツ)_/¯. Will just have to build the image from scratch."
-          fi;
+          if [ "${{ inputs.skip-build-optimization }}" == "false" ]; then
+            LAST_MASTER_BUILD="$REMOTE_DOCKER_REPOSITORY:${{ inputs.branch-main }}"
+            echo "Could not pull $LAST_BRANCH_BUILD, will try last master build $LAST_MASTER_BUILD"
+            if ! docker pull "$LAST_MASTER_BUILD"; then
```
mirgee, line 63: Not actionable, just thinking that this is quite inefficient, likely pulling some layers unnecessarily, while it should be easy to check whether a specific layer is present in the registry / remote cache at build time...

libvcx_core/src/api_vcx/utils/devsetup.rs
```diff
@@ -45,7 +44,7 @@ impl SetupGlobalsWalletPoolAgency {
         let init = Self::init().await;
 
         let pool_name = Uuid::new_v4().to_string();
-        let genesis_path = get_temp_dir_path(DEFAULT_GENESIS_PATH).to_str().unwrap().to_string();
+        let genesis_path = get_temp_file_path(DEFAULT_GENESIS_PATH).to_str().unwrap().to_string();
```
mirgee, line 47: Perhaps doing `.into_os_string().into_string().unwrap()` might give very slightly more informative errors (the original `OsString`) than unwrapping an option, but I don't think it matters.

mirgee (approved):

aries_vcx/tests/test_creds_proofs.rs
Patrik-Stas: yeah I know ^_^ 

A task for another day to sort these out

aries_vcx_core/src/indy/ledger/pool.rs
Patrik-Stas: it was duplicated before, but actually this PR removed the duplication

.github/actions/build-image/action.yml
Patrik-Stas: Maybe possible if the entire approach is re-thought, as the cached images are now compressed, so the only way right now to reuse their layers is to load them up locally first. But I get your point

libvcx_core/src/api_vcx/utils/devsetup.rs
Patrik-Stas: I acknowledge the suggestion, I think there might be a lot more places to use `OsString` with more benefit. Since this is just testing, I'll skip it in this pr, but could be nice to do in bulk separately

PR #901, opened 2023-07-18
Rename tails variables in prover code

Patrik-Stas:
- This is breaking change for provers constructing "selected credentials" datastructure when creating proof. While previously they had option to specify field `tails_file`, now they have to use `tails_dir`. It's essentially forced renamed, semantics didn't change. Previously `tails_file` attribute was actually expected to point to a directory (containing tails file)
(eg. see https://github.com/hyperledger/aries-vcx/pull/901/files#diff-7542b0e9db422e41b81fc4f3bf5da974a9e5ec73a1a6a2d08d4698d4a9eb9b89R26 as example of prover accommodating this breaking change)


gmulhearn (approved):
nice, has been a smell for me for awhile

gmulhearn:
out of sync with main, but LGTM

PR #915, opened 2023-07-26
Handler layer for SM

bobozaur:
TODO


aries_vcx/src/new_protocols/connection/invitee/handlers.rs
```diff
@@ -0,0 +1,217 @@
...
+    };
+
+    let request = Request::with_decorators(msg_id, content, decorators);
+    let sm = AriesSM::Connection(ConnectionSM::InviteeRequested(sm));
+    sm_storage.put_new_state(sm_id, sm).await?;
+
+    Ok(request)
+}
+
+pub async fn handle_response<S, W>(
```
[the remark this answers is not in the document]
Patrik-Stas: ### Opinion:

It tries to shield away user from concept of states and state machines, and worry about the result because the data just gets update in the storage. But how will user handle following cases then: 
- the transitioning function fail with error which might disappear upon retry - typically any IO - for example resolving DID Document. The FSM in the storage stay the same, an `Error` is returned from the function. Is every `Error` supposed to signal retryable operation?
- Is `Ok` variant of the function always the "happy case"? What would be approach for entering "hard failed" states? It seems like there 2 options:
  - If you return Err, we'll need to distinguish retryable errs and protocols erros (not retryable)
  - if you return Ok, the user won't know what actually happened in transition - whether the state machine update in the storage behind the scenes - in our transition function - ended up in "happy case" state or "non-happy case" state.

aries_vcx/src/new_protocols/connection/inviter/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+// If this flies with a single state then
+// we can remove the generic
+#[derive(Clone, Debug)]
+pub struct InviterConnection<S> {
+    did: String,
+    verkey: String,
+    state: S,
+}
+
+impl InviterConnection<InviterComplete> {
```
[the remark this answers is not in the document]
Patrik-Stas: ### ELI5 observatory comment for other reviewers:

Bogdan's approach reduced number of state significantly - one reason for that is elimination of states such as `Invited` stated, which used to be a state where a "invitation message has been received and somewhat validated", and the state held the invitation.

This was somewhat convenient for the consumers, because upon receiving invitation, they would construct "connection state machine", 1 type of object they would deal the entire time, anytime they need to do something / drive protocol further. You could have a table or a collection of Connection objects, track their state and try to drive them forward asynchronously over time, if you wanted.

With Bogdan's approach such "message holder" state are gone - if you receive invitation, and you want to drive the protocol later on (for example after mobile app restart, server restart, etc.) it's your responsibility to have a place for message store. 
For example, you will need to store Invitation, and then pick it up later on if you want to proceed. It's simple with connection, but can get more complex when you are already in the middle of protocol. For example you are trying to receive a credential. You have sent `credential-request` message, so you already have a state machine in some "Requested" state. Next up you receive `issuer-credential`, but you only want to proceed with human supervision. So if you want to enable that being asynchronous, or span multiple servers, you need to store the `issuer-credential` message along with the `requested` state machine. You either link up the 2 objects, or simply anytime you pick-up unprocessed message from your store, you resolve the appropriate state machine (probably a lot better approach).

### Opinion
Although some places we need to imo keep some granularity in number of state, this approach forces consumers to work bit more with message, but actually this plugs in kind of nicely in mobile setting with mediators, which literally provide you the "store of unprocessed messages" where many message types trigger some UI user prompt. After the transition is completed, typically the message would be marked in the mediator as processed / marked for deletion.

In server setting there might likely be some queue middleware delivering the messages the processors and it's kind of the same story.

The only thing you need it store of **unprocessed** messages, and from a message you then always resolve relevant state machine, do some processing, trigger a prompt, after transition is done, delete the message (or keep it for some message audit-log purposes)

aries_vcx/src/new_protocols/connection/invitee/handlers.rs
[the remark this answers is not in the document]
Patrik-Stas: ### Opinion

Since these - essentially transition functions / processing functions - are divorced from state machines themselves, it makes state machines a bit more universal - technically someone could implement their own way of, for example - processing connection request.
In old paradigm the code existed directly on the state machines, and you either take it or not.

Patrik-Stas (commented):
left a review

Patrik-Stas:
@bobozaur to be more concrete on the next steps, how about these:

- As we agreed on call with @gmulhearn lets do some in-mem test of 2 parties
- Can you also incorporate draft how to treat IO errors in the "handlers" which result in "retryable errors"?  I think this is important piece of puzzle which can imply pro/against arguments to the approach with the storage interface. 

## Questions
We touched on a call, but perhaps I forgot. I was making an argument that in order to find out ID of a state machine, you likely already made IO operation to load the state machine from a storage. Then you select ID out of it, pass it in here, just for the handler to do another IO to load the same state machine again. 
Do you have argument why you want to keep loading state machine in the handler, or how to avoid duplicating the IO?

bobozaur:
> @bobozaur to be more concrete on the next steps, how about these:
> 
>     * As we agreed on call with @gmulhearn lets do some in-mem test of 2 parties
> 
>     * Can you also incorporate draft how to treat IO errors in the "handlers" which result in "retryable errors"?  I think this is important piece of puzzle which can imply pro/against arguments to the approach with the storage interface.
> 
> 
> ### Questions
> 
>     1. We touched on a call, but perhaps I forgot. I was making an argument that in order to find out ID of a state machine, you likely already made IO operation to load the state machine from a storage. Then you select ID out of it, pass it in here, just for the handler to do another IO to load the same state machine again.
>        Do you have argument why you want to keep loading state machine in the handler, or how to avoid duplicating the IO?

Maybe I'm missing something, but are you asking "To find the ID of the state machine, people have to look up the state machine to find it's ID. How to avoid that?".

`Id` could mean whatever you want it to mean in the context of identifying a state machine. Nobody said it has to be a single `String` or something like that. If you look up state machines by `thread_id` and whatever else, then use that in your lookup when you implement the `StateMachineStorage` trait. 

Idk if this answers your question as I don't think I understood it properly.

Patrik-Stas:
@bobozaur 
> Maybe I'm missing something, but are you asking "To find the ID of the state machine, people have to look up the state machine to find it's ID. How to avoid that?"?

Not quite. I am saying following will be happening:
- App code: You get a message, check message `thid`
- App code: You lookup state machine based on `thid` to find out its `id` (IO fetching the state machine)
- App code: call some handling function, passing in storage impl and `id`
- aries-vcx: fetch state machine based on `id` (IO fetching the same data again)

So my point that in order to find out `id`, you will likely already have the state machine.

bobozaur:
## Idea

I won't include problem reports in the POC because I think a good way to deal with them is to refactor the `AriesVcxError` and have some variant like `ProtocolError` which could either contain the `ProblemReport` or the pieces needed to construct it.

I'm basing this idea on the fact that a `ProblemReport` really just represents an error, so we might as well treat it as such. Users could match on the error in places where they want to send a problem report and send it if present / can be generated.

Having a completely new return type that would contain the problem report would just make things more convoluted (but indeed slightly more visible) where as just encapsulating problem reports as errors would make things more uniform, with the slight downside that a consumer of the code would have to match on the error to get the problem report (which I don't think it's that big of a deal, really).

The refactor of `AriesVcxError` could also aid in giving more info to consumers in regards to what happened with the state machine. The three possible outcomes are `Success -> New state`, `NonFatalError -> OldState` and `FatalError -> FailedState`. The last two could be error variants that people could through for more info, while obviously an `Ok(AriesMessage)` simply represents the happy case of a transition.

@Patrik-Stas @gmulhearn @mirgee Thoughts?

Patrik-Stas:
> But why wouldn't you use the thread_id (and any other needed details) to get the state machine directly? Is there anything preventing that in the interface that I don't see?
Perhaps I am biased from commonly used approach currently and I don't see better viable alternatives but...

IMO you will likely store state machines along with other information about them together, in some sort of record such as (assuming json document database, but could be a row in database, a value under key in redis, etc..):
```
{ 
  "id": "foo",
  "creation_time": "2000-10-31T01:30:00.000-05:00"
  "thread_id": "bar",
  "state": "complete",
  "something_else": "foobar",
  "fsm": "<serialized aries-vcx fsm>"
}
```
Now again - if I receive an aries message, I look at its `thread_id` (say with value `bar`) and need to find ou `id` of the respective state machine to proceed. So I need to make IO to match one of these records against `thread_id: bar`, in order to find out that the state machine. So already at this point, you can simply load up entire state machine instead of just taking out `id` and throwing away the rest. Anyway, one way or the other, you made IO to match `thread_id --> fsm_id`.

But the the handling function itself will have to again perform similar IO, but this time matching the record by `id` - which seem unnecessary, but current design will IMO lead to these flows.

Ofc `id` and `thread_id` can be indexed, so this IO can be relatively fast, but that doesn't seem as a strong argument.

bobozaur:
I'm not particularly versed in No-SQL databases, but in SQL ones a fairly simple `JOIN` could take the `thread_id` and output the state machine directly. I doubt there's no similar mechanism in No-SQL databases.

The `StateMachineStorage::Id` associated type **does not** have to be the actual database ID of the state machine, just the information needed to *identify* the state machine in the database. It could even be an enum where each variant is handled differently, etc.

I still fail to see how this wouldn't be suitable for virtually any use case. Is there something I'm still missing?

Patrik-Stas:
@bobozaur it doesn't really matter what database you use - sql or nosql or file - my concern is that there will **have to** be 2 IO "get" operations (fetch `id` by `thid`, fetch `fsm` by `id`) in storage design where normally 1 IO op would suffice (fetch `fsm` by `thid`)

Do you see what I mean now?

And there's 3 options how to go about it:
- say it's okay price to pay for the given interface design
- accept entire `fsm` instead of `fsm_id` and loading the `fsm` (this give programmer the **option** to load entire fsm based on `thid`
- introduce the original idea of "resolver" which would be capable of resolving the fsm based on the message itself 

I am not sure what's the best approach, but I want to make sure you understand my concern and I am would to know your opinion - or whether you even see that as an issue.

Patrik-Stas:
> The refactor of AriesVcxError could also aid in giving more info to consumers in regards to what happened with the state machine. The three possible outcomes are Success -> New state, NonFatalError -> OldState and FatalError -> FailedState. The last two could be error variants that people could through for more info, while obviously an Ok(AriesMessage) simply represents the happy case of a transition.

That seems reasonable, thanks. Wanted to make sure we enable user to know what happened without explicitly re-loading the state machine and inspecting it. The proposed idea should be fine imo

Patrik-Stas:
> @bobozaur it doesn't really matter what database you use - sql or nosql or file - my concern is that there will have to be 2 IO "get" operations ( [op 1] fetch id by thid, [op 2] fetch fsm by id ) in storage design where normally 1 IO op would suffice ( [op 1] fetch fsm by thid)

### brainstorming
Btw I had same concern for post-transitions modifications to database. If user wants to track FSM state, or last update time, I was concerned the current design would force him to "re-load" the FSM after the transition is complete, inspecting it, then making additional update IO operation - hence under many storage design, 2 unnecessary IO operations.

But then I thought the "post-transition" db updates can be part of `put_new_state` implementation - so if serialized FSM is stored in the same record along with other metadata, custom fields, these can be updated all together. 

So I think the storage part is 🆗 , this just a bit of a braindump - just an invitation to comment on this though process ^_^

bobozaur:
@Patrik-Stas 

I still don't get it, so I'll try to provide an example.

Let's assume you have these two tables in MySQL:
```sql
CREATE TABLE thread_id_mapping (
    thread_id VARCHAR(100),
    state_machine_id VARCHAR(100)
);

CREATE TABLE state_machines (
    state_machine_id VARCHAR(100),
    state_machine_json VARCHAR(5000)
);
```

And a draft of the `StateMachineStorage` impl would be:
```rust
struct SmStorage;

impl StateMachineStorage for SmStorage {
    type Id = String;

    async fn get(&self, id: &Self::Id) -> Result<AriesSM, AriesVcxError> { 
        let sm = query("
             SELECT state_machine_json FROM state_machines S
             INNER JOIN thread_id_mapping T
                   ON T.state_machine_id = S.state_machine_id
             WHERE T.thread_id = ?;").fetch_one().deserialize()?;
    }

    async fn put_new_state(&self, id: Self::Id, sm: AriesSM) -> Result<(), AriesVcxError>;

    async fn put_same_state(&self, id: Self::Id, sm: AriesSM) -> Result<(), AriesVcxError>;
}
```

You're saying that users start with a `thread_id` and they **need** to get a `state_machine_id` so they can call `StateMachineStorage::get()` with it to get the `state_machine_json` to deserialize, but that's not true. They can define the retrieval and storage however they want, and they can use whatever identifier Rust type they choose to represent it.

In my example, using the `thread_id` directly and joining on the `state_machine_id` achieves the exact same thing in only one database call.

bobozaur:
The `Id` associated type represents "the information needed to *identify* the state machine in the database". That doesn't mean it has to be the actual ID used as a key in the database. In the scenario we've been talking about it is indeed the `thread_id`.

Perhaps the name is misleading so if you have a better idea I'm open to suggestions.

Patrik-Stas:
@bobozaur, okay I see what you mean - the meaning of id in the storage trait depends on what is being passed into the "handler" functions from the app code. However, I am not fully on board with the versatility.

You've described id as "the information needed to identify the state machine in the database." As a developer integrating this POC, I would interpret `id` as the state machine's primary identifier. But in reality, "id is the piece of information that you choose to pass into transition/handler methods." You could argue that we just rename `id` to something else, but I find more issues with this design:

- Generality: This interpretation is a bit abstract for an implementor and requires anticipation of how you're going to call the transition/handler methods.
- Indirection: The storage trait isn't invoked directly by the developer; it's called by the transition/handler methods. So, in explaining id, you need to refer to a different part of the codebase.
- Inconsistency: Developers might have multiple implementations of these traits, each with a different get strategy. Different definitions of id across implementations could lead to confusion.
- Efficiency Concerns: Once a developer decides what id represents, they must keep that consistent across the storage implementation, potentially leading to inefficient update implementations. For example, if developer decides that the "id" argument represents aries thread_id, an UPDATE SQL query would then also have to involve JOINs, just like the `get` implementation.

Therefore, I suggest following changes as remedy:
- Remove the get method from the trait, letting users load the state machine on their own.
- Redefine id in `put_*` methods to represent the actual "id" of the state machine, a primary identifier in the storage

These changes would make the definition of id in the trait clearer, allow users to load state machines as needed, and prevent forcing users into potentially inefficient update implementations.

aries_vcx/src/new_protocols/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+pub struct BootstrapInfo {
+    service_endpoint: Url,
+    recipient_keys: Vec<String>,
+    routing_keys: Vec<String>,
+    did: Option<String>,
+    service_endpoint_did: Option<String>,
+}
+
+#[async_trait]
+pub trait MessageHandler {
```
Patrik-Stas, line 39: This MessageHandler approach, while it makes it more convenient to write the implementations of the functions inside, suffers from similar weaknesses as our (recently remove) approach of passing in `Arc<Profile>` in function arguments.

- Breaking law of Demeter.  In order to process a message, I need to construct handler which needs to have set up `LedgerRead`, `LedgerWrite`, ... etc. while in practice, the function I am about to call might be only using some or none of the components. 
Then that leads to passing in mock implementations "when you know ahead of time the function you are planning to call will not internally leverage that particular component".
- It's also remove hits about how certain functions work. For example with old style functions, you would have something like `process_request(read: Arc<dyn LedgerRead>, msg: Message)` so you can tell at glance - "ah, this function will be reading ledger"

aries_vcx/src/new_protocols/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+            ledger_read,
+            ledger_write,
+            wallet,
+            anoncreds,
+        };
+
+        (did, msg_handler)
+    }
+
+    async fn _test_connection_handler(pool_handle: i32) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
```
Patrik-Stas, line 302: this is really helpful for review 👍

aries_vcx/src/new_protocols/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+
+        // Build request
+        let (alice_did, alice_verkey) = alice.wallet().create_and_store_my_did(None, None).await?;
+        let mut did_doc = AriesDidDoc {
+            id: alice_did.clone(),
+            ..Default::default()
+        };
+        did_doc.set_service_endpoint(bootstrap_info.service_endpoint);
+        did_doc.set_routing_keys(bootstrap_info.routing_keys);
+        did_doc.set_recipient_keys(bootstrap_info.recipient_keys);
```
Patrik-Stas, line 344: this is essentially transformation of invitation into diddoc, could be a function ( i believe something like that exists in the codebase already, thought not sure if it's reusable here)

aries_vcx/src/new_protocols/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+            ..Default::default()
+        };
+
+        let response_decorators = ResponseDecorators {
+            thread: Thread::new(request.decorators.thread.map(|t| t.thid).unwrap_or(request.id)),
+            please_ack: None,
+            timing: Some(timing),
+        };
+
+        let response = Response::with_decorators(Uuid::new_v4().to_string(), response_content, response_decorators);
```
Patrik-Stas, line 383: This lower level approach - removing state machines - is exposing the lack and the need for message builder.

Patrik-Stas (commented):
while I don't like everything about it, I think the demonstration test could be fairly convincing if refactored a bit more. 

I like in this demo, threading is really detached from protocol, as we've been thinking about. I've left remarks in similar note on Miro's PR over here https://github.com/hyperledger/aries-vcx/pull/928#discussion_r1293799425
And I am wondering if Miro goes ahead with the changes, whether you two have intersections in your conclusions.

I think if you:
- get rid of `MessageHandler` 
and Miro 
- eliminates thids from transition functions 
- perhaps extract functions so they are not defined directly on state machine structures

The code of both PRs will start to fruther converge to similar direction.

aries_vcx/src/new_protocols/mod.rs
bobozaur: I don't have a strong opinion on this. It was indeed mainly a left over that was adjusted to make function signatures shorter.

aries_vcx/src/new_protocols/mod.rs
bobozaur: It's more than that.

I added the `BootstrapInfo` in the view of also maintaining those `DID` info (like `service_endpoint_did`) in case you might wanna re-read it from the ledger. 

No idea if there are use cases for that and, if there aren't, we can simply resolve it to a DidDoc directly and call it a day.

aries_vcx/src/new_protocols/mod.rs
bobozaur: It does, yes.

aries_vcx/src/new_protocols/mod.rs
bobozaur: Also, this kind of helped with the test and creating the two entities, but you already know I'm not in favor of testing dictating production code 😅 .

Another possible use case would be uniform decorator handling, but idk if this really should reside in here or not. The argument about having to pass a lot of components which might not even be used is good though, so I'll refactor this to only be used for entities in the test.

bobozaur:
The implementation was simplified by all together removing decorators. This has two reasons:
- better putting into spotlight the protocol specific content (as that's what matters from a protocol standpoint)
- removing the decorators boilerplate (as they should be processed differently anyway, perhaps at a layer above this)

Additionally, @Patrik-Stas is right about the `MessageHandler` trait being to excessive with the components it needs, so modified the function signatures to take whatever components they need to process a specific message. `MessageHandler` is only there for testing purposes and was moved to the `test` module.

aries_vcx/src/new_protocols/mod.rs
bobozaur: `MessageHandler` was officially demoted 😄

aries_vcx/src/new_protocols/mod.rs
```diff
@@ -0,0 +1,355 @@
...
+}
+
+async fn parse_pw_did_invitation<T>(
+    ledger_read: &T,
+    mut invitation: PairwiseDidInvitationContent,
+) -> VcxResult<BootstrapInfo>
+where
+    T: IndyLedgerRead + AnoncredsLedgerRead,
+{
+    let service = match get_service(ledger_read, &invitation.service_endpoint).await {
```
Patrik-Stas, line 73: just noticed - it should be more like 
```
&invitation.content.services[0]
```
instead of 
```
&invitation.service_endpoint
``
(correct, although not ideal)

Patrik-Stas:
Closing this, will be still available among closed PRs for future reference. One of the things that came out of this was also additional proposal branch which can be found here https://github.com/hyperledger/aries-vcx/pull/944

PR #918, opened 2023-07-28
Refactor features, enable build with --no-default-features

Patrik-Stas:
At high level, this re-enables building aries-vcx, aries-vcx-core without any features enabled. Comes along with some refactoring to make sense of things and prepare ground for gradual removal of vdrtools based components.

- `aries-vcx-core`: define more granular features - this will come handy in subsequent PR's removing parts of vdrtools `vdrtools_anoncreds`, `vdrtools_wallet`, `vdrtools_ledger`
- move non-indy specific files out of indy directories (eg `ToBaseAgencyClientWallet` moved from `aries_vcx_core::wallet::indy::agency_client_wallet` to `aries_vcx_core::wallet::agency_client_wallet`
- `aries-vcx-core`: include `indy-api-types` as dependency to access `indy_api_types::errors` directly to simplify error mapping (`aries_vcx_core/src/errors/mapping_vdrtools.rs` )
- remove support for `vdrtools` based ledger client from `did_resolver_sov` - it was causing some complications which won't be relevant as soon as this https://github.com/hyperledger/aries-vcx/pull/914 get merged

### DIFF
--- a/aries_vcx_core/Cargo.toml
+++ b/aries_vcx_core/Cargo.toml
@@ -6,7 +6,10 @@ edition = "2021"
 [features]
 ########################## DEP FLAGS ################################
 # Feature flag to include the libvdrtools dependency
-vdrtools = ["dep:libvdrtools"]
+vdrtools = ["vdrtools_anoncreds", "vdrtools_wallet", "vdrtools_ledger"]
+vdrtools_anoncreds = ["dep:libvdrtools"]
+vdrtools_wallet = ["dep:libvdrtools"]
+vdrtools_ledger = ["dep:libvdrtools"]
 # Feature flag to include the 'modular library' dependencies (vdrtools alternatives; indy-vdr, indy-credx)
 modular_libs = ["dep:indy-vdr", "dep:indy-credx", "dep:indy-ledger-response-parser", "dep:lru"]
 vdr_proxy_ledger = ["modular_libs", "dep:indy-vdr-proxy-client"]
--- a/aries_vcx_core/Cargo.toml
+++ b/aries_vcx_core/Cargo.toml
@@ -6,7 +6,6 @@ edition = "2021"
 [features]
 ########################## DEP FLAGS ################################
 # Feature flag to include the libvdrtools dependency
-vdrtools = ["vdrtools_anoncreds", "vdrtools_wallet", "vdrtools_ledger"]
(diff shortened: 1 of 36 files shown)


aries_vcx_core/Cargo.toml
```diff
@@ -28,6 +30,7 @@ indy-vdr = { version = "0.3.4", default-features = false, features = ["ffi", "lo
 # - update libvdrtools to use =0.3.6 ursa
 indy-credx = { git = "https://github.com/anonyome/indy-shared-rs.git", rev = "7342bc624d23ece8845d1a701cd2cdc9cd401bb0", optional = true }
 libvdrtools = { path = "../libvdrtools", optional = true }
+indy-api-types = { path = "../libvdrtools/indy-api-types" }
```
gmulhearn, line 33: As discussed privately; it's not ideal IMO that this is a forced dependencies, as IMO we really want to be moving away from anything vdrtools related, or rather avoid modular_libs consumers having to include any vdrtools deps. 

I see now that you're importing this here such that you can update `errors/mod.rs` from:
```
#[cfg(any(feature = "vdrtools", feature = "modular_libs", feature = "vdr_proxy_ledger"))]
mod mapping_vdrtools;
```
to
```
mod mapping_indy_api_types;
```

Where `mapping_indy_api_types` is necessary for non-vdrtools traits (indy_vdr_ledger), since indy_vdr_ledger relies on IndyError from indy-api-types.

gmulhearn: Anyway, i know you already know all this, I'm just catching up on it... 

this import hasn't made it worse, just highlighted to me that the modular_libs feature is leaking some vdrtools stuff

gmulhearn: perhaps as a note, you could add a comment explaining that this is temporarily whilst we are de-vdrtooling the codebase

gmulhearn (approved):

aries_vcx_core/Cargo.toml
Patrik-Stas: Thanks George, yes, we are on the same page here~ 

I'll omit the note as addressing the problem you highlighted should be relatively imminent and simple right after https://github.com/hyperledger/aries-vcx/pull/914 is in.

Issue #922, opened 2023-08-04
Presenting same credential with different revocation timestamps (or lack thereof)


gmulhearn-anonyome:
Possibly relates to; https://github.com/hyperledger/aries-vcx/issues/834. Or atleast it's in the same code area to be reworked.

# Problem
Currently the implementation of `build_rev_states_json` in `prover_internal.rs` forces that the same cred used on multiple referents must present the exact same timestamp in it's NRP. E.g. if "referent1" is being presented for interval  `to: 10` and "referent2" for `to: 20`, then build_rev_states_json will force both to the same timestamp for presentation (10 or 20, depending on ordering). 

# Bigger Problem?
This problem is possibly bigger than the main problem, but heavily related. In `build_rev_states_json` it looks like any credential that is revocable, is forced into creating a NRP, even if the proof request did not request it... Testing this against ACA-py 0.8.2 results in a verification failure, with the following error: `"VALUE_ERROR::Timestamp on sub-proof #0 is superfluous vs. requested attribute {referent-here}"`. We are creating a NRP that is not necessary, and ACA-py sees this as superfluous.

# Expected Behaviour
* referents that do not need to present a NRP (no referent interval nor global interval) should not be forced to present one 
* credentials with the same rev-reg-id should not be forced to present the same timestamp (https://github.com/hyperledger/aries-vcx/issues/834)

gmulhearn-anonyome:
Happy to take this one on if we agree it is a problem; i think `prover_internal` may just need a fresh lookover

gmulhearn:
Update: i still get the error with VCX<->ACApy (1.1+)

PR #923, opened 2023-08-04
draft: issue 922 Prover revocation states fix up

gmulhearn:
#922 

TODO


gmulhearn:
@JamesKEbert , yeah can close this, was only half way thru, and the changes are quite dated now. I think the issue is still relevant though and should remain, i still get the error with VCX<->ACApy in very particular setups

PR #924, opened 2023-08-04
fix(aries-vcx): fixed dependency listing in README for `Cargo.toml`

arminveres:
Using path does not work and crates from git sources need be specified through `git=...`.

### DIFF
--- a/aries_vcx/README.md
+++ b/aries_vcx/README.md
@@ -1,12 +1,12 @@
 # aries-vcx
-`aries-vcx` is more of a "library" rather than "framework". We strive to be not too 
-opinionated and simply provide building blocks for whatever you want to build. 
+`aries-vcx` is more of a "library" rather than "framework". We strive to be not too
+opinionated and simply provide building blocks for whatever you want to build.
 
-You can use `aries-vcx` to build both end-user products (mobile wallets in role of credential 
-holder and prover) or server (typically in role of issuer, verifier). 
+You can use `aries-vcx` to build both end-user products (mobile wallets in role of credential
+holder and prover) or server (typically in role of issuer, verifier).
 
 Generally, the crate allows you to:
-- create encrypted wallet, 
+- create encrypted wallet,
 - read/write from/to Indy ledger,
 - establish didcomm connections and exchange messages,
 - create and process Aries messages to drive Aries protocols.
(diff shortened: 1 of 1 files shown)


Patrik-Stas:
hi @arminveres thanks for the fix! 

By Hyperledger policy, all commits must be signed off (DCO) - which basically every commit message must contain following bottom line (example):
```
    Add integration test to nodejs wrapper
    
```

there's a switch in `git` to help you do this `--signoff`. You can fix your PR like this:
```
git commit --signoff --amend
```
and then push force the branch.

Thank you!

Patrik-Stas:
@arminveres by the way, what we should probably also mention in the readme, it's highly recommended to fix the revision of the dependency, so you are not simply fetching the latest `main` branch (eg. adding `ver` field specification for the dependency in `cargo.toml`)

arminveres:
Thanks @Patrik-Stas, I amended the commit and included a version tag. Should I be more verbose about it?

Patrik-Stas:
@arminveres thanks for prompt DCO fix, but note that the example is currently not quite valid. `version` is not valid attribute when dealing with `git` dependencies - have a look at this https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

>  You can combine the git key with the `rev`, `tag`, or `branch` keys

arminveres:
I see, you are right, it built on my system, but it ignored the version tag and proceeded to use the latest tag available, so I didn't notice it. I adjusted the commit. Thanks!

Patrik-Stas:
thanks @arminveres looks good, merging

By the way, I am curious to ask, are you looking to build some using `aries-vcx`, or are you yet just looking around? Let me know if you'd need some support; also we have discord and weekly community calls

arminveres:
> Thanks @arminveres, looks good, merging.
> 
> By the way, I am curious to ask, are you looking to build some using `aries-vcx`, or are you yet just looking around? Let me know if you'd need some support; also we have discord and weekly community calls

Thanks!
I am working on my thesis and I am planning on using Aries, also possibly other parts of Hyperledger. I have already joined the discord. I will surely need support, its a tad overwhelming with so many repos and frameworks to go through! Thanks again!

PR #928, opened 2023-08-10
DID Exchange Protocol

mirgee:
Initial version of [DID Exchange protocol](https://github.com/hyperledger/aries-rfcs/blob/main/features/0023-did-exchange/README.md#request-message-attributes) implementation with support for peer DIDs (peer:did:2 for now), new `DidDocument` crate, and JWS signing / verification.

The primary aim is not to achieve perfection at this stage but to roll out a first iteration prototype.

Correctness has been tested (so far only) using AATH tests running against AcaPy peer:did branch (or against aries-vcx itself), where all didx tests are passing whether the requester or responder role is assumed by either side.


aries_vcx/src/protocols/did_exchange/service/responder/response_sent/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+            ),
+            output: response,
+        })
+    }
+
+    pub fn receive_complete(
+        self,
+        complete: Complete,
+    ) -> Result<DidExchangeServiceResponder<Completed>, TransitionError<Self>> {
+        if complete.decorators.thread.thid != self.state.request_id {
```
Patrik-Stas, line 62: I know this mirror mostly current approach and we've been tracking ids/threads in state machines, it's been questioned whether this should really be happening, and whether the state machines shall do a bit less - only care for the protocol part, but less so for decorators.
I am not sure if that's always the case, and whether it's possible to assume a state machine would *never* need access to some decorator. 

I do tend to lean toward the idea that we should deal with thids, ids, pthids on protcol level. Either way, to look up and attempt to update a state machine, upper layers of code will definitely have to do it the same thing - inspect thid/pthid/id of a arriving message and use that information to lookup the proper state machine in storage in order to perform update. So checking thid, pthids etc. in transitions ultimately leads to duplication which already must been performed by consumer of this code.
I guess this could make again make SMs further bit more lightweight in terms of logic and content of the states.

(And I wonder if doing so could by a chance lead to further to rethinking any other high level design decisions in terms of protocols/state machines)

Patrik-Stas (commented):
had a chance to have a short look only so far, but left some comment and thouhts

aries_vcx/src/protocols/did_exchange/service/responder/response_sent/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+};
+
+use super::DidExchangeServiceResponder;
+
+pub mod config;
+mod helpers;
+
+impl DidExchangeServiceResponder<ResponseSent> {
+    pub async fn receive_request(
+        ReceiveRequestConfig {
```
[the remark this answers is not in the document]
mirgee: > I find it harder to read function when I have to hop into structure definition to see what data type what is.

I think this argument does not apply when using an IDE, but can even so be easily mitigated by defining the config object near the function definition and/or destructuring.

Passing in a single config object with 10 fields, unlike passing in 10 arguments:
* allow adding new optional arguments without changing function's signature - a breaking change (data locality),
* are order-independent (more important the larger the number of arguments),
* are more explicit in terms of argument-value assignment and thus safer,
* make function calls more readable (when done well + more important the larger the number of arguments).

aries_vcx/src/protocols/did_exchange/service/responder/response_sent/mod.rs
Patrik-Stas: Fair enough, there's not reason to make contract, on the other hand, from all protocols we deal with today, at their core, each of them can work without using decorators whatso-ever. So asking for full message, including decorators, when you actually only need the "protocol piece" to make things work (eg. doing something what'd be considered driving state/protocol forwrad/generate reply). 

But that links to the other point - currently we indeed access decorators. So the second point I argue for, is that we probably should remove `thid / pthid / id` verifications on state machine level. What we are actually doing is runtime testing of the code consuming this. 
While it's nice to have and - "why not to make assert if you can" - it doesn't quite feel inlined with single resposibility principle. 

Btw I know some 2 years ago I claimed the opposite - that state machine should have awareness of its "identity", by knowing thread it belongs to - as it's essential piece of information. Back than we had philosophy of keeping serialization model backwards compatible, and the idea was that you will ALWAYS end up wanting linking a FSM with thread id, and since we provide stable serialization format, we can do that for the user. 
Obviously we change our thoughts on that, and serialized state machines are "black boxes" now - so users of aries-vcx are definitely (and we know from our experience) tracking threads ids on their own. Hence that explains why I changed position on this. I now think state machines shouldn't know/care "who they are", and should be burden of the upper layer to assure passing in correct messages.

Patrik-Stas: One thing I have missed is consideration that without dealing with `thread_id`, `id` etc. we can't build complete "ready-to-send" response msgs.
But I can see that - along with other decorators like `~timing`, localization etc. - being often part of an upper layer, some message processor in the future. Otherwise, if keep inflating state machine / transition layer with additional logic (set up timing decorator, set up locatization decorator and what not, ... ) either end up being very opinionated, or we further expand function argument with config which will direct the logic how the decorators shall be set up (and the function arguments are already quite massive, albeit wrapped into a struct)

But it can be gradual process. We can start building full blown messages now, then pull some logic (attaching decorators) on an upper "framework" layer later. 
All that being said, I still think we should probably start in a point where state machine doesn't remember it's `thid`.

mirgee: As mentioned in private, I have many reservations to what you said there (but perhaps am aligned with the way you are going). I don't think imposing assumptions left and right is not a good approach (even when it might lead to a good solution). I would much rather think in terms of components, responsibilities and data, but for that, we need to first use common language, and you are using terms for which we might have different definitions in confusing way (using term "state machine" while commenting on `service` module, etc.).

Ultimately, I think that GitHub comments are not the best medium for this conversation.

aries_vcx/tests/test_did_exchange.rs
```diff
@@ -0,0 +1,99 @@
...
+use did_doc_sov::service::didcommv2::ServiceDidCommV2;
+use did_doc_sov::service::ServiceSov;
+use messages::msg_fields::protocols::out_of_band::invitation::Invitation;
+use url::Url;
+
+use crate::utils::devsetup_alice::create_alice;
+use crate::utils::devsetup_faber::create_faber;
+
+#[tokio::test]
+async fn did_exchange_test() {
```
Patrik-Stas, line 25: currently failing on line `47` (even before the `todo!`)

did_doc_sov/src/service/mod.rs
Patrik-Stas, line 1: hmmm, strange that this still shows up delta 👀  - we got these changes merged in main recently

aries_vcx/src/protocols/did_exchange/service/mod.rs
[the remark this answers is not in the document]
mirgee: The intention behind the naming was to get rid of transitions and the typestate pattern...

aries_vcx/src/protocols/did_exchange/service/generic/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+    Completed(DidExchangeServiceRequester<Completed>),
+}
+
+#[derive(Debug, Clone)]
+pub enum ResponderState {
+    ResponseSent(DidExchangeServiceResponder<ResponseSent>),
+    Completed(DidExchangeServiceResponder<Completed>),
+}
+
+impl GenericDidExchange {
```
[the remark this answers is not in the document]
mirgee: I agree in principle with it not making sense in terms of methods shared between the roles, but this is a quick-and-dirty implementation and the current approach allows for braindead simple handling on the `aries-vcx-agent` level which was the entire point of the enum to begin with.

aries_vcx/src/protocols/did_exchange/service/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+use did_doc_sov::DidDocumentSov;
+use public_key::Key;
+
+use super::{
+    states::{abandoned::Abandoned, traits::ThreadId},
+    transition::transition_result::TransitionResult,
+};
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct DidExchangeService<I, S> {
```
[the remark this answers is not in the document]
mirgee: There is very little shared between requester and responder as well, but they both share need to store their own and counterparty's DDO / key, hence the shared struct. But I don't have strong opinion either.

aries_vcx/tests/test_did_exchange.rs
mirgee: I know...

did_doc_sov/src/service/mod.rs
mirgee: Rebased

aries_vcx/src/protocols/did_exchange/service/generic/mod.rs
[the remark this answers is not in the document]
Patrik-Stas: Got it, I'd like to go ahead with that

aries_vcx/src/protocols/did_exchange/service/mod.rs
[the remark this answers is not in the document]
Patrik-Stas: I see, but given the current state of thing, are you okay with the rename?

aries_vcx/src/protocols/did_exchange/service/requester/request_sent/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        let key = Key::new(
+            our_did_document
+                .verification_method()
+                .first()
+                .unwrap()
+                .public_key()
+                .key_decoded()?,
+            KeyType::Ed25519,
+        )?;
```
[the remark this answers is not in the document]
mirgee: I disagree, the whole point of this function is to compose a message from easy-to-provide config, so I don't think putting more burden on the caller outweighs the benefit of simplifying internal code at this point.

mirgee:
@gmulhearn @bobozaur I don't know why @Patrik-Stas requested a review, this PR is not ready for review yet.

mirgee:
We should be able to group the messages crates after this is merged

Patrik-Stas (approved):

PR #936, opened 2023-08-14
Allow parsing unqualified DIDs

mirgee:
Enables parsing of unqualified DIDs in `did_parser`, a temporary measure for the purposes of gracefully transitioning to using fully qualified DIDs, and will be reverted once the transition is complete.

### DIFF
--- a/did_parser/tests/did_url/positive.rs
+++ b/did_parser/tests/did_url/positive.rs
@@ -312,4 +312,14 @@ test_cases_positive! {
         None,
         HashMap::new(),
         HashMap::new()
+
+    test_case22:
+        "2ZHFFhzA2XtTD6hJqzL7ux#1",
+        Some("2ZHFFhzA2XtTD6hJqzL7ux"),
+        None,
+        Some("2ZHFFhzA2XtTD6hJqzL7ux"),
+        None,
+        Some("1"),
+        HashMap::new(),
+        HashMap::new()
 }
--- a/did_resolver_registry/src/error.rs
+++ b/did_resolver_registry/src/error.rs
@@ -3,12 +3,16 @@ use std::error::Error;
 #[derive(Debug)]
 pub enum DidResolverRegistryError {
     UnsupportedMethod,
+    UnqualifiedDid,
 }
 
 impl std::fmt::Display for DidResolverRegistryError {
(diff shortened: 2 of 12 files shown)


did_parser/tests/did_url/positive.rs
```diff
@@ -312,4 +312,14 @@ test_cases_positive! {
         None,
         HashMap::new(),
         HashMap::new()
+
+    test_case22:
```
Patrik-Stas, line 316: while did is technically a did url, I would find this test more fitting under `tests/did` or perhaps append some additional path to test more didurl code paths

did_resolver_registry/src/error.rs
```diff
@@ -3,12 +3,16 @@ use std::error::Error;
 #[derive(Debug)]
 pub enum DidResolverRegistryError {
     UnsupportedMethod,
+    UnqualifiedDid,
```
Patrik-Stas, line 6: in context of did resolver, I think it would be safe to assume that unqualified did are `did:sov:<unqialified>`?

Patrik-Stas (approved):
looks good but left some comments - consider them, otherwise feel free to merge

did_resolver_registry/src/error.rs
mirgee: I don't think so. If you know you have unqualified did, you should qualify it before attempting to resolve it (with the method only you know you want to use, it is not implicit). If you don't know you have unqualified DID, you should get a clear error indicating such fact instead of failing later with potentially confusing error unrelated to the root cause.

did_parser/tests/did_url/positive.rs
mirgee: Agreed, adding more tests is warranted 👍

mirgee: However, the test you are commenting on is testing a strict DID URL (see the fragment at the end).

mirgee: Added a few more tests.

did_resolver_registry/src/error.rs
Patrik-Stas: fair enough

Patrik-Stas (approved):

PR #938, opened 2023-08-16
Attempt to convert legacy DID documents to new during deserialization

mirgee:
Introduces a lightweight, easy-to-remove legacy DDO data structure to the `did_document_sov` crate. During the deserialization of `DidDocumentSov`, if the incoming data matches the legacy structure, it is converted to `DidDocumentSov` with `IndyAgent` service type.

Received legacy DDO is first converted to did:peer:2 following [this](https://github.com/TimoGlastra/legacy-did-transformation) document, and from it then back-resolved to the new DDO format following the did:peer:2 spec.

PR #939, opened 2023-08-16
Extract subset of changes made in #928

mirgee:
Extracts changes made to `did_doc`, `did_doc_sov`, `did_doc_sov`, `did_peer`, and `public_key` crates in #928. Most of these changes are either consequences of #913 or additions of features found to be useful during #928.

The reason for extracting these changes is that
* doing so will minimize conflicts encountered when rebasing #928 after changes made to the modified crates (e.g. in  #938),
* they are independent and useful outside of #928.

TODO: Prune


did_doc_sov/src/lib.rs
```diff
@@ -139,3 +154,87 @@ impl<'de> Deserialize<'de> for DidDocumentSov {
         })
     }
 }
+
+impl Serialize for DidDocumentSov {
```
Patrik-Stas, line 158: add tests for `serialization -> deserialization`, assure we don't duplicate services

did_doc_sov/src/service/aip1.rs
```diff
@@ -59,6 +62,15 @@ impl TryFrom<Service<ExtraFieldsSov>> for ServiceAIP1 {
     }
 }
 
+impl TryFrom<Service<HashMap<String, Value>>> for ServiceAIP1 {
+    type Error = DidDocumentSovError;
+
+    fn try_from(service: Service<HashMap<String, Value>>) -> Result<Self, Self::Error> {
+        let extra = serde_json::from_value::<ExtraFieldsAIP1>(serde_json::to_value(service.extra())?)?;
```
Patrik-Stas, line 69: looks like will never fail currently, since `ExtraFieldsAIP1 {}` doesn't have any fields

mirgee: Added `deny_unknown_fields` (seems to work somehow).

did_doc_sov/src/lib.rs
```diff
@@ -164,7 +164,12 @@ impl Serialize for DidDocumentSov {
 
         for service_sov in &self.services {
             let service: Service<ExtraFieldsSov> = service_sov.clone().try_into().map_err(serde::ser::Error::custom)?;
-            builder = builder.add_service(service);
+            // Not very efficient, but
```
Patrik-Stas, line 167: perhaps one more way to consider:
Change `DidDocumentSov` such instead of 
```
services: Vec<ServiceSov>
```
it would use HashMap, using service id as key. That way:
- you have to worry about duplicates at serialization
- you could give user option to update a service, though this might come off as untransparent API, as builder's `add_service` would be possibly not appending but overwriting (if service with same `id` exists) (so then `add_service` should probably be renamed)
- guarantee that serialized diddoc does not have 2 services with same `id`

Some possible drawbacks:
- if you read a diddoc which has service with 2 duplicate services, incorrectly with the same ID (perhaps we should fail at deserialization in a first place), then you would end up "squashing them together" and produced serialized diddoc would not be the same as the invalid original.

I think the current fix is sufficient, just something to consider - perhaps for separate PR if you think such a change make sense.

Patrik-Stas (approved):

did_doc_sov/src/lib.rs
mirgee: You have to realize why `DidDocumentSov` stores `Vec<ServiceSov>`. It has a very good reason, which is that the infallible getter `services`, must return `&[ServiceSov]` (as I was explaining to you).

mirgee: Another issue is that from `HashMap` you get values only in arbitrary order (which is the reason I got rid of it in the builder), with O(capacity) complexity. Then the deserialization -> serialization -> deserialization test would be failing due to arbitrariness in the ordering.

PR #934, opened 2023-08-11
Eliminate dependence of indy-ledger-response-parser on indy-api-types

mirgee:
Eliminates dependence of `indy-ledger-response-parser`  on `indy-api-types`, which now leak out of `libvdrtools` only through `indy-wallet`.

PR #940, opened 2023-08-17
Try Hide indy-api-types from modular_libs consumers

gmulhearn:
Testing...

i _believe_ that modular_libs should not need anything to do with indy-api-types after this great PR: https://github.com/hyperledger/aries-vcx/pull/934, therefore we should make it optional in aries_vcx_core..

this takes the dep count from `554` to `300` for:
```
cargo check --features modular_libs --no-default-features -p aries-vcx
```


mirgee (approved):
Trimming down on dependencies is always a win 👍

PR #941, opened 2023-08-18
Expose public key getter on verification method

mirgee:
Introduces a `public_key::Key` getter in `VerificationMethod` to facilitate easy retrieval of the key representation derived from `VerificationMethod`.

### DIFF
--- a/did_doc/src/schema/verification_method/verification_method.rs
+++ b/did_doc/src/schema/verification_method/verification_method.rs
@@ -1,7 +1,8 @@
 use did_parser::{Did, DidUrl};
+use public_key::Key;
 use serde::{Deserialize, Serialize};
 
-use crate::schema::types::jsonwebkey::JsonWebKey;
+use crate::{error::DidDocumentBuilderError, schema::types::jsonwebkey::JsonWebKey};
 
 use super::{public_key::PublicKeyField, VerificationMethodType};
 
@@ -37,8 +38,15 @@ impl VerificationMethod {
         &self.verification_method_type
     }
 
-    pub fn public_key(&self) -> &PublicKeyField {
+    pub fn public_key_field(&self) -> &PublicKeyField {
         &self.public_key
+    }
+
+    pub fn public_key(&self) -> Result<Key, DidDocumentBuilderError> {
+        Ok(Key::new(
+            self.public_key.key_decoded()?,
+            self.verification_method_type.try_into()?,
+        )?)
     }
 }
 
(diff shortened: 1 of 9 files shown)


did_doc/src/schema/verification_method/verification_method.rs
```diff
@@ -37,9 +38,16 @@ impl VerificationMethod {
         &self.verification_method_type
     }
 
-    pub fn public_key(&self) -> &PublicKeyField {
+    pub fn public_key_field(&self) -> &PublicKeyField {
```
Patrik-Stas, line 41: To stay consistent with other methods (even across different structs/impls) I'd suggest to keep the original name and come up with alternative name for the new method

did_doc/src/schema/verification_method/verification_method.rs
```diff
@@ -37,9 +38,16 @@ impl VerificationMethod {
...
         &self.public_key
     }
+
+    pub fn public_key(&self) -> Result<Key, DidDocumentBuilderError> {
```
Patrik-Stas, line 45: Find it kind of code-smellish that this method return DidDocumentBuilderError - as consumer I would be surprised to get this type of error from this method

Patrik-Stas (changes requested):

did_doc/src/schema/verification_method/verification_method.rs
mirgee: Remember that `public_key` is not a field of `VerificationMethod`. This is a function which returns a representation of the value of `public_key_*`, and that representation has a type called `PublicKeyField`. Hence, I think this naming is not breaking consistency (which is getter names and the corresponding fields having the same name).

did_doc/src/schema/verification_method/verification_method.rs
mirgee: If you are suggesting it should return `PublicKeyError`, that would likely require the client crate implement a conversion to its error type from both `DidDocumentBuilderError` and `PublicKeyError`. This may be usually necessary, as the client crate will likely be using `Key` extensively anyways, but other use-cases may come up where the client crate extracts the `public_key`, does not use it, and just passes it to another crate. One may argue whether that is an anti-pattern, but not unimaginable, so why complicate the API?

If you are suggesting to rename the error type, then I agree, as the crate is called `did_doc` instead of the original `did_doc_builder`.

did_doc/src/schema/verification_method/verification_method.rs
Patrik-Stas: Ah I see it now. Make sense

did_doc/src/schema/verification_method/verification_method.rs
Patrik-Stas: I see... fair enough, I think rename will be nice though. You can do it separately or here, I am hitting approve button. Feel free to merge

Patrik-Stas (approved):

Issue #786, opened 2023-03-28
Remove IO between the parties in protocol integration tests


Patrik-Stas:
Rewrite integration tests such that tests focused on non-connection protocols do not exchange messages via didcomm, instead juggle the messages between 2 instance of state machines - simply assume a trusted communication channel exist.

Eg where previously we would have code
```
prover.build_presentation()
send_presentation(prover, connection_prover_to_verifier) // generate message, send it over didcomm connection
try_verify_presentation(verifier, connection_verifier_to_prover) // lookup messages in agency, pick presentation req., try to use it to drive protocol
```
while the code sample looks fairly simply here, the illustrated methods would typically do bunch of stuff behind the scenes as indicated by comment.

So instead, we should do:
```
prover.build_presentation()
let msg = prover.get_presentation_msg();
verifier.process_presentation(msg)
```
and not concern ourselves with the message delivery at all. Just pass the message around the memory for test purpose.

### Note:
This also ties with refactoring around state machines which has been discussed but yet not formalized as Github issue - state machines themselves should not be concerned with sending any data in the first place. That means we need to eliminate states like `<MsgType>Sent` (for example `PresentationRequestSentState` in proof verifier)

styvane:
I want to work on this one.

PR #945, opened 2023-08-19
Declutter issuance&presentation protocols

Patrik-Stas:
### Changes
- No breaking changes
- Remove concept of "Action" as way to trigger actions/process messages from both Holder and Issuer handlers & state machines
- Externalize interactions with mediated connections and related code to files `mediated_holder.rs`, `mediated_issuer.rs` etc. out of state machine/handlers code itself
- Removed disabled unit tests - majority of tests was testing validity of sequence of FSM transitions. These will be irellevant with typestate pattern

### Next steps:
- We can follow this up with changes to states to remove IO - that itself has higher value than transformation to typestate pattern. 
  - After that, we can in parallel boost up testing performance significantly https://github.com/hyperledger/aries-vcx/issues/786
- In last stage we could rewrite thing to typestate + abstracting out & unifying approach for message processing as drafted in https://github.com/hyperledger/aries-vcx/pull/944


.github/workflows/main.yml
```diff
@@ -2,11 +2,11 @@ name: CI
...
 on:
   push:
-    branches:
-      - main
-  pull_request:
     branches:
       - "**"
+#  pull_request:
+#    branches:
+#      - "**"
```
mirgee, line 9: Can you re-enable the CI, please?

aries_vcx/src/handlers/issuance/mediated_holder.rs
```diff
@@ -0,0 +1,99 @@
...
+    let messages = connection.get_messages(agency_client).await?;
+    if let Some((uid, msg)) = holder_find_message_to_handle(sm, messages) {
+        sm.process_aries_msg(ledger, anoncreds, msg.into(), Some(send_message))
+            .await?;
+        connection.update_message_status(&uid, agency_client).await?;
+    }
+    Ok(sm.get_state())
+}
+
+// todo: returns specific type
```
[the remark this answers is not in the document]
gmulhearn: oh right, this is just to match existing impl, mb

aries_vcx/src/protocols/issuance/mod.rs
```diff
@@ -2,25 +2,23 @@ use aries_vcx_core::ledger::base_ledger::AnoncredsLedgerRead;
...
-            AriesVcxErrorKind::InvalidJson,
-            format!(
-                "Cannot handle message {:?}: thread id does not match, expected {:?}",
-                message, thread_id
-            ),
-        ));
-    };
-    Ok(())
-}
+// todo: restore, how to call this on arbitrary Aries message?
```
gmulhearn, line 9: did you not resolve this TODO? it looks like you're using src/handler/utils::verify_thread_id now for this

libvcx_core/src/api_vcx/api_handle/issuer_credential.rs
```diff
@@ -30,31 +31,25 @@ pub fn issuer_credential_create(source_id: String) -> LibvcxResult<u32> {
...
-            .step(&get_main_anoncreds()?, message.into(), Some(send_message))
-            .await?;
+        credential.process_aries_msg(msg.into()).await?;
     } else {
         let messages = mediated_connection::get_messages(connection_handle).await?;
-        if let Some((uid, msg)) = credential.find_message_to_handle(messages) {
-            credential
-                .step(&get_main_anoncreds()?, msg.into(), Some(send_message))
-                .await?;
+        // if let Some((uid, msg)) = credential.find_message_to_handle(messages) {
```
gmulhearn, line 50: leftover?

gmulhearn:
i'm not too familar with the usage of the actions flows to fully review... so consider this as a _half-review_. but lgtm other than above comments

Patrik-Stas:
@gmulhearn okay, just for extra context, those "actions" were an obscure way to interact with state machines. For both processing incoming messages and also triggering "local actions" - for example sending response to counterparty.

So where previous you would have
`enum Action { 
   AcceptOffer(offer_message)
   SendRequest()
}`
and universal method
```
fn update(action: Action) {
   match (action) { 
      AcceptOffer(msg) => self.accept_msg_offer(msg)
      SendRequest() => self.send_request()
   }
} 
```
(in fact `self.accept_msg_offer(msg)` didn't really exist on top layer, but usually these matches were delegated to the most bottom layer - the state machine structure itself.

now we simply have every "action" replaced by a appropriate method on both handler and state machine level. There's no more universal `update` method which might-or-might-not do 10 different things, based on what the action variant is.

aries_vcx/src/protocols/issuance/mod.rs
Patrik-Stas: ah yeah, it's thread_id verification is indeed still in place, as it was before

Patrik-Stas: removing this commented fragment

aries_vcx/src/handlers/issuance/mediated_issuer.rs
```diff
@@ -0,0 +1,112 @@
...
+use aries_vcx_core::anoncreds::base_anoncreds::BaseAnonCreds;
+use aries_vcx_core::wallet::base_wallet::BaseWallet;
+use messages::msg_fields::protocols::cred_issuance::propose_credential::ProposeCredential;
+use messages::msg_fields::protocols::cred_issuance::CredentialIssuance;
+use messages::msg_fields::protocols::notification::Notification;
+use messages::AriesMessage;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+// todo: remove unused args
```
mirgee, line 16: This should be very straightforward, let's do it in this PR

aries_vcx/src/handlers/issuance/mediated_holder.rs
```diff
@@ -0,0 +1,99 @@
...
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use crate::errors::error::prelude::*;
+use crate::handlers::connection::mediated_connection::MediatedConnection;
+use crate::handlers::issuance::holder::Holder;
+use crate::handlers::util::{matches_opt_thread_id, matches_thread_id};
+use crate::protocols::issuance::holder::state_machine::HolderState;
+
+pub async fn holder_update_with_mediator(
```
mirgee, line 17: This function is only used in tests (and same for issuer, verifier, prover), so perhaps it should be placed near the test code (under `#[cfg(test)]`)?

aries_vcx/src/handlers/proof_presentation/mediated_verifier.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+    Ok(sm.get_state())
+}
+
+#[allow(clippy::unwrap_used)]
+pub fn verifier_find_message_to_handle(
+    sm: &Verifier,
+    messages: HashMap<String, AriesMessage>,
+) -> Option<(String, AriesMessage)> {
+    trace!("VerifierSM::find_message_to_handle >>> messages: {:?}", messages);
```
mirgee, line 44: These traces are a little a misleading (but no biggie, hopefully this code won't last long).

aries_vcx/src/handlers/issuance/mediated_issuer.rs
```diff
@@ -0,0 +1,112 @@
...
+        trace!("Issuer::update_state >>> found no msg to handle");
+    }
+    Ok(sm.get_state())
+}
+
+#[allow(clippy::unwrap_used)]
+pub fn issuer_find_messages_to_handle(
+    sm: &Issuer,
+    messages: HashMap<String, AriesMessage>,
+) -> Option<(String, AriesMessage)> {
```
bobozaur, line 40: Are there any plans regarding removing stuff like this? IMO the protocol should have nothing to do with finding messages and even if we don't remove the mediated connection stuff, perhaps we can separate concerns a bit better?

Patrik-Stas: Definitely, it's only exists because:
-  `update_state` functions in `libvcx_core`, and these are also relied on in NodeJS tests
- integration tests for aries-vcx itself - as soon as IO removal is completed, and tests are not using mediated connection anymore (which will be possible with https://github.com/hyperledger/aries-vcx/pull/946 merged) we can get back to this, and at very least lift these `*_find_messages_to_handle` code-smells a level up to `libvcx_core`, where it can be (or not) dealt with separately.

aries_vcx/src/handlers/issuance/mediated_holder.rs
Patrik-Stas: Also libvcx though, so not yet

https://github.com/hyperledger/aries-vcx/pull/945#discussion_r1302203423

Patrik-Stas:
Thanks for reviews, everything has been addressed, please approve if all looks good

aries_vcx/src/handlers/issuance/mediated_holder.rs
mirgee: These `_update_with_mediator` are positively not used in libvcx

bobozaur (approved):

aries_vcx/src/handlers/issuance/mediated_holder.rs
Patrik-Stas: Ahh yeah! You are right! I got it mixed up with `*_find_message_to_handle` which are.

So you are right, I'll put these functions under `#[cfg(test)]`

Patrik-Stas: Decent improvement indeed, applied

mirgee (approved):

PR #946, opened 2023-08-20
Remove msg-sending IO for issuer and holder

Patrik-Stas:
### Strategy
- This PR modifies Issuer/Holder handlers (and underlying ode) to remove any message sending. Instead, this layer only produces aries messages, but caller is responsibly for making sure the messages are sent to the respective protocol counterparty.
- This unlocks possibility to rapidly simplify test setup for issuance/verification protocol tests, as we will be no longer forced to interact with mediator and setup didcomm connection prior to testing other protocols.
- State machines will be generally well position for further transition to type-state pattern following  https://github.com/hyperledger/aries-vcx/pull/928 pattern where transitions function include "msg to be sent out" in return type.

### Changes
- For any user of aries-vcx or components built on top, there's breaking changes in terms of intermediate states. Some state were removed. See below.
- For any direct `aries-vcx` user, there's breaking changes in terms of sending messages. You have to send replies to counterparty yourself. 
   - This doesn't apply to `libvcx` and `rust-agent` built on top of `aries-vcx`, which preserve original behaviour and assure message sending themselves.

##### Issuer: Removed state `OfferSent`
- Remove `OfferSent` state, keep only `OfferSet`; treat `OfferSet` as `OfferSent`
- To keep Issuer dependant works, Issuer still has its original method `send_credential_offer` which unlike before, does not modify state. Hence existing code should still keep working.

##### Issuer: Removed state `CredentialSent`
- Removed state `CredentialSent`, introduced state `CredentialSet` instead
- The semantic reason between them is self-explaining, "set" doesn't imply the message has been sent
- Structural reason between the 2 is `CredentialSet` has additional field `msg_issue_credential: IssueCredential` to decouple construction of credential from sending `issue-credential` msg

#### Holder: Removed state `RequestSent`,  `ProposalSent`
- Removed state `RequestSent`, introduced state `RequestSet` instead
- Removed state `ProposalSent`, introduced state `ProposalSet` instead
- Generally changes and reasoning are analogous to what's described above for Issuer 

##### Holder: credential-ack
- Upon processing `issue-credential`, sender would previously automatically send `credential-ack` message, if ack was requested by the issuer. This no longer happens. Instead, after processing `issue-credential` msg, you should call `get_final_message` which can give you back aries message `Option<AriesMessage>` you shall send back to counterparty. The message might be `credential-ack` or `problem-report` (if processing `issue-credential` message failed)

##### Issuer & Holder: problem reports
- For both Issuer and Holder, if a transition fails, the state machine moves to failed state like before, BUT: doesn't automatically sends the problem report. You have to check the state and send problem report yourself. You have to do that yourself.


aries_vcx/src/handlers/issuance/holder.rs
```diff
@@ -52,21 +52,25 @@ impl Holder {
         Ok(())
     }
 
-    pub async fn send_request(
+    // todo: is the my_pw_did really necessary? is it used under the hood?
```
gmulhearn, line 55: this is something that has bugged me before about the holder API and older versions of credx/indy-sdk in general... i believe this gets attached in the credential request for it to act as a form of nonce for the issuer to sign over... however i don't believe it technically HAS to be a DID. 

In the past when i looked into this, deeper ursa code for issuer logic just refers to this value as a string called "prover id", rather than "prover did". unfortunately i believe libs like credx and vdrtools will enforce that this value MUST be a DID (when really it could be any arbitrary DID).

anoncreds-rs does a better job IMO by allowing an "entropy" string to be used instead when constructing a cred req; https://github.com/hyperledger/anoncreds-rs/blob/main/src/services/prover.rs#L107C34-L107C34

Patrik-Stas: Seems like we should create issue for this :-) It's really annoying indeed

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
```diff
@@ -96,20 +92,18 @@ pub struct IssuerSM {
...
 
-    let mut decorators = IssueCredentialDecorators::new(Thread::new(id.clone())); // this needs a Thread per RFC...
-    let mut timing = Timing::default();
-    timing.out_time = Some(Utc::now());
-    decorators.timing = Some(timing);
-
-    Ok(IssueCredential::with_decorators(id, content, decorators))
+    let mut decorators = IssueCredentialDecorators::new(Thread::new(id.clone()));
+    decorators.thread.thid = thread_id.clone();
+    decorators.please_ack = Some(PleaseAck::new(vec![]));
```
bobozaur, line 105: Why even set this?

aries_vcx/src/protocols/issuance/issuer/states/offer_set.rs
```diff
@@ -28,14 +34,45 @@ impl OfferSetState {
...
+    fn from((state, err): (OfferSetState, ProblemReport)) -> Self {
+        trace!("SM is now in Finished state");
+        FinishedState {
+            cred_id: None,
+            revocation_info_v1: Some(RevocationInfoV1 {
+                cred_rev_id: None,
+                rev_reg_id: state.rev_reg_id,
+                tails_file: state.tails_file,
+            }),
+            status: Status::Failed(err),
```
bobozaur, line 75: Any chance of getting rid of conversions like these, or are they outside the scope of this PR? I think a constructor like `FinishedState::new()` is much clearer than something like `(offer_state, problem_report).into()`

bobozaur (commented):
Seems OK to me so far. Left some comments outside of the review.

aries_vcx/src/protocols/issuance/issuer/states/offer_set.rs
Patrik-Stas: oh yeah. I hate those from heart :-D Yeap I'll do it here

aries_vcx/src/protocols/issuance/holder/state_machine.rs
```diff
@@ -64,13 +64,13 @@ impl fmt::Display for HolderFullState {
             HolderFullState::Initial(_) => f.write_str("Initial"),
             HolderFullState::ProposalSent(_) => f.write_str("ProposalSent"),
             HolderFullState::OfferReceived(_) => f.write_str("OfferReceived"),
-            HolderFullState::RequestSent(_) => f.write_str("RequestSent"),
+            HolderFullState::RequestSet(_) => f.write_str("RequestSent"),
```
mirgee, line 67: "RequestSet"

aries_vcx/src/protocols/issuance/holder/state_machine.rs
```diff
@@ -223,6 +226,23 @@ impl HolderSM {
         Ok(Self { state, ..self })
     }
 
+    #[deprecated]
```
mirgee, line 229: It would be nice to the future reader leave a deprecation note

aries_vcx/src/handlers/mod.rs
```diff
@@ -55,9 +55,8 @@ impl From<IssuerState> for u32 {
             IssuerState::Initial => 0,
             IssuerState::ProposalReceived => 1,
             IssuerState::OfferSet => 2,
-            IssuerState::OfferSent => 3,
             IssuerState::RequestReceived => 4,
-            IssuerState::CredentialSent => 5,
+            IssuerState::CredentialSet => 5,
```
mirgee, line 59: Probably just to state the obvious - all credentials in `OfferSent` and `CredentialSent` states will become invalid and will have to be discarded before updating to this version (and all code relying on the old states updated).

mirgee (commented):
> Issuer: If constructing anoncreds credential / issue-credential message fails, the state machine transitions to failed Finished state, but doesn't sends the problem report. You have to check the state and send problem report yourself.

> Holder: if construction anoncreds credential-request / request-credential message fails, the state machine transitions to failed Finished state, but doesn't sends the problem report. You have to check the state and send problem report yourself.

How and where is the potential problem report to be constructed? It is currently just not being sent (unless I am missing something)?

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
Patrik-Stas: just preserved the original behaviour

aries_vcx/src/handlers/mod.rs
Patrik-Stas: Yes indeed. We could potentially keep CredentialSent (just keep it, but do not transition into it anymore through out the protocol) temporarily ease the transition and remove the state later.

aries_vcx/src/protocols/issuance/issuer/states/offer_set.rs
Patrik-Stas: Removed

Patrik-Stas:
> How and where is the potential problem report to be constructed? It is currently just not being sent (unless I am missing something)?

Yes, it's constructed the same way as before, in the same places. In PR description I am saying that unlike before, now these problems reports are not sent automatically.

mirgee:
> > How and where is the potential problem report to be constructed? It is currently just not being sent (unless I am missing something)?
> 
> Yes, it's constructed the same way as before, in the same places. In PR description I am saying that unlike before, now these problems reports are not sent automatically.

That is what I am trying to pointing out. They are just not being sent anymore, and there is no way of sending them (AFAIK?). Can we include problem report sending as part of this PR? I tend to think it is quite important.

aries_vcx/src/protocols/issuance/holder/state_machine.rs
```diff
@@ -223,12 +231,32 @@ impl HolderSM {
...
+    // convenience function for sending the credential request. This will be removed in the future
+    // and some form of replacement will be provided instead outside of state machines.
+    pub async fn send_credential_request(&self, send_message: SendClosure) -> VcxResult<()> {
+        match self.state {
+            HolderFullState::RequestSet(ref state) => {
+                let mut msg: RequestCredential = state.msg_credential_request.clone().into();
+                let mut timing = Timing::default();
+                timing.out_time = Some(Utc::now());
+                msg.decorators.timing = Some(timing);
+                error!("Sending msg {msg:?}"); // TODO: it looks like received message has thread "test", but doesnt look like thats what we sending
```
mirgee, line 244: Just a reminder to address the TODO and remove the debug log

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
```diff
@@ -405,7 +405,8 @@ impl IssuerSM {
                             }),
                         })
                     }
-                    // todo: dont' transition, throw error, add to_failed transition() api which SM consumer can call
+                    // todo: 1. Don't transition, throw error, add to_failed transition() api which SM consumer can call
+                    //       2. Also create separate "Failed" state
                     Err(err) => {
                         let problem_report = build_problem_report_msg(Some(err.to_string()), &self.thread_id);
                         error!(
```
mirgee, line 412: This error log is misleading, but more importantly, the problem reports are still not being sent in case an error occurred, and nodejs consumer has no easy way to obtain and send them.

Patrik-Stas: Restored problem report sending on `libvcx_core` and `agent` level

aries_vcx/src/handlers/issuance/holder.rs
mirgee: Perhaps remove this comment and add a link to the issue somewhere near the definition of `prover_create_credential_req`?

aries_vcx/src/handlers/issuance/holder.rs
```diff
@@ -39,54 +67,55 @@ impl Holder {
...
     }
 
-    pub async fn decline_offer<'a>(&'a mut self, comment: Option<&'a str>, send_message: SendClosure) -> VcxResult<()> {
-        self.holder_sm = self
-            .holder_sm
-            .clone()
-            .decline_offer(comment.map(String::from), send_message)
-            .await?;
-        Ok(())
+    // ultimately this will be eliminated, as with state pattern, state transition will yield reply message
```
mirgee, line 90: Why not return the message from `prepare_credential_request` now?

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
mirgee: Good 👍

aries_vcx/src/handlers/issuance/holder.rs
Patrik-Stas: ✅

aries_vcx/src/handlers/mod.rs
mirgee: Let's just keep this in mind and make the next release a "major" one...

aries_vcx/src/protocols/issuance/issuer/state_machine.rs
mirgee, line 1: Good, but `SendClosure` is still imported (same for holder). In general, it wouldn't hurt to do something about all those compiler warnings...

mirgee (approved):

Issue #951, opened 2023-08-22
Minimize imports of agency_client


Patrik-Stas:
Following code
```
use aries_vcx::{agency_client::httpclient::post_message, errors::error::VcxResult, transport::Transport};
use async_trait::async_trait;
use url::Url;

pub struct HttpClient;

#[async_trait]
impl Transport for HttpClient {
    async fn send_message(&self, msg: Vec<u8>, service_endpoint: Url) -> VcxResult<()> {
        post_message(msg, service_endpoint).await?;
        Ok(())
    }
}
```
is in several crates - causing to bring in entire agency_client as dependency.

Issue #949, opened 2023-08-22
Oob builder for messages crate


Patrik-Stas:
Building OOB invitation is fundamental way for anyone to get started with aries-vcx, in its current state. The first thing you ever need if you want to simulate a basic use case with aries-vcx is to create the oob invitation.
This need came up for example in:
- https://github.com/hyperledger/aries-vcx/issues/870
- mentorship mediator project https://github.com/nain-F49FF806/axum-test-server/tree/main/src

Currently in `aries_vcx/src/handlers/out_of_band` there's `OutOfBandSender` which is position on same level like other protocol handlers, but in fact, this is just a builder for OOB messages. 
More-over, working with the `OutOfBandSender` is currently difficult, because it's method
```
pub fn append_service(mut self, service: &OobService) -> Self {
```
ultimately requires you to import `diddoc_legacy` into your project, which is not obvious.

Proposal:
- There's already established consensus that builder for messages crate is necessary, and constructing message by hand is currently cumbersome.
- So let's use this as opportunity to create first message builder, specifically for oob invites; also get rid of `OutOfBandSender` in the process

Additional note:
There's also `OutOfBandReceiver` - we'll get rid of that in separate step - it contains method. `nonmediated_connection_exists` which is being used in some NodeJS tests.

Patrik-Stas:
We now have builders for Messages crate, but we should yet do something about `OutOfBandSender` and `OutOfBandReceiver`

Issue #870, opened 2023-06-05
Create Alice Faber CLI Demo for aries_vcx (Rust)


gmulhearn:
For new rust consumer of the aries_vcx crate, it would be nice if we had a simple CLI demo application/s of the classic Faber (issuer/verifier) and Alice (holder/prover) interactions. 

This demo application would help new rust consumers see the expected flows of interactions for the different DI roles.

Some existing examples:
* ACA-py: https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/README.md
* Cheqd/AFJ: https://github.com/cheqd/cheqd-aries

# Conditions
* 2 different rust CLI applications (Alice & Faber)
* Uses aries_vcx rust crate directly
* Walks thru: Connection establishment, Credential Issuance/Receiving, Proof presentaiton/verification
* Reference the demo from the repo's documentation (README), for new comers to look at

# Open Questions
* Should the demo support revocable credentials? or keep it simpler with non-revocable?
* Indy profile or modular_libs Profiles? or support both?
* Application assumes we are running the localhost ledger? or support for any ledger (e.g. some `.env` config)?
* What message transport mechanism for agents?
    * Could just be a simple actix localhost server for posting and picking-up messages (no special mediator protocols)
    * In the future we would likely switch this to use the mediator (+ mediator client) implementation, once it's ready

gmulhearn:
We do have some existing "demos" in the `tests` directory i think. But IMO they are not very accessible. open to suggestions

Patrik-Stas:
Yeah it would be great for adoption and onboarding

Some thoughts:
> Uses aries_vcx rust crate directly
might lead to duplication with `ariese-vcx-agent` code. Depends what's the main purpose of such demo 
1. show `aries-vcx` code usage, as code example,
2. or demo of the overall workflows, lib capabilities, utility tool for testing

I think the 2nd is more important than the 1st point, as to understand aries-vcx API, that should be achieved by better docs, looking at integration tests using aries-vcx directly
So assuming we would agree the main goal is 2nd point, then I think it makes sense to build such demo on top of `aries-vcx-agent` https://github.com/hyperledger/aries-vcx/tree/main/agents/rust/aries-vcx-agent (which is also used to in Aries AATH). But then obviously we wouldn't want to build the CLI stuff into it directly, but rather in another layer on top of `aries-vcx-agent`  which we kind of started https://github.com/hyperledger/aries-vcx/pull/692

[...]

PR #952, opened 2023-08-22
draft: Demo for aries vcx

Patrik-Stas:
Partially addressing https://github.com/hyperledger/aries-vcx/issues/870
- Added Alice & Faber demo to establish connection
- Modified `simple_message_relay` to enable its consumption also as library, rather than executable only
- Extended `simple_message_relay` with mpsc channels which forward any incoming message to mpsc receiver. This is nice because in the demo, we don't need to poll HTTP API for new messages - but rather instantly get notified in-memory.

Having to interact with aries-vcx from scratch from new application, some improvements I've identified:
- https://github.com/hyperledger/aries-vcx/issues/949
- https://github.com/hyperledger/aries-vcx/issues/951

## Note
This can be thought of as temporary solution to demonstrate the APIs and usage. And while far from ideal yet (todos left, issues created), it's a lot easier to understand than pointing someone to integration tests today. Main issue of integration tests is that they are all couploed with vcxagency-node mediator, but also other issues.
After integration tests are in better shape after:
- https://github.com/hyperledger/aries-vcx/pull/945
- https://github.com/hyperledger/aries-vcx/pull/946
- and subsequent removal of IO, and therefore coupling with vcxagency-node mediator
- and some refactoring of "Alice" "Faber" devsetup agents

I believe that integration tests could be in condition to demonstrate the usage (and surface outstanding API issues) equally good as this demo.
But for the time being, this demo is a lot simpler than the existing integration tests.

## Demo
Running the demo produces following logs
image


Patrik-Stas:
@gmulhearn as a creator of original issue, do you agree this being in a single file? I think your vision to have 2 files, 1 for alice, one for faber.
I am fine with that too. Just started in one file to make this simple for me, but actually might be a lot more readable for someone if we split the roles.

mirgee:
> After integration tests are in better shape after:
> 
> * https://github.com/hyperledger/aries-vcx/pull/945
> * https://github.com/hyperledger/aries-vcx/pull/946
> * and subsequent removal of IO, and therefore coupling with vcxagency-node mediator
> * and some refactoring of "Alice" "Faber" devsetup agents
> 
> I believe that integration tests could be in condition to demonstrate the usage (and surface outstanding API issues) equally good as this demo.
> But for the time being, this demo is a lot simpler than the existing integration tests.

Will we remove the demo after the work mentioned is completed, i.e. integration tests are in such a condition that they fulfill the current purpose of the demo, i.e. to demonstrate aries-vcx usage, as discussed in speaking?

Patrik-Stas:
> > After integration tests are in better shape after:
> > 
> > * [Declutter issuance&presentation protocols #945](https://github.com/hyperledger/aries-vcx/pull/945)
> > * [IO for issuer and holder as opt-in #946](https://github.com/hyperledger/aries-vcx/pull/946)
> > * and subsequent removal of IO, and therefore coupling with vcxagency-node mediator
> > * and some refactoring of "Alice" "Faber" devsetup agents
> > 
> > I believe that integration tests could be in condition to demonstrate the usage (and surface outstanding API issues) equally good as this demo.
> > But for the time being, this demo is a lot simpler than the existing integration tests.
> 
> Will we remove the demo after the work mentioned is completed, i.e. integration tests are in such a condition that they fulfill the current purpose of the demo, i.e. to demonstrate aries-vcx usage, as discussed in speaking?

I'd be in favor, there surely will be a lot of duplication between this and integration tests after refactoring them

tools/simple_message_relay/src/main.rs
```diff
@@ -95,14 +46,15 @@ mod tests {
             .set_payload(msg)
     }
 
-    #[actix_web::test]
+    // #[actix_web::test]
```
gmulhearn, line 49: i dont think this test will run now

demo/src/http_client.rs
```diff
@@ -0,0 +1,13 @@
+use aries_vcx::{agency_client::httpclient::post_message, errors::error::VcxResult, transport::Transport};
+use async_trait::async_trait;
+use url::Url;
+
+pub struct HttpClient;
+
+#[async_trait]
+impl Transport for HttpClient {
+    async fn send_message(&self, msg: Vec<u8>, service_endpoint: Url) -> VcxResult<()> {
+        post_message(msg, service_endpoint).await?;
```
gmulhearn, line 10: Q) should we still be using stuff from the `agency_client` crate? I wonder if the demo would be cleaner if you just redefined `post_message` in this file rather than pulling in agency_client; the `agency_client::post_message` impl is full of stuff that consumers might not care about (e.g. mocking and layers of error handling, etc).

demo/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+use std::process::exit;
+
+use env_logger;
+use serde::Deserialize;
+use tokio::sync::mpsc;
+use tokio::sync::mpsc::{Receiver, Sender};
+use url::Url;
+
+use aries_vcx::transport::Transport;
+use aries_vcx_core::ledger::base_ledger::IndyLedgerRead;
```
gmulhearn, line 17: nit: could we try get rid of warnings in this crate pls (e.g. unused imports)

demo/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+    tx_invitation.send(msg_oob_invitation).await.unwrap();
+
+    info!("****** Faber received connection invitation via trusted means, will reply sending 'connection-request' message ******");
+    let msg_request = faber.wait_for_message::<Request>().await.unwrap();
+    // todo: make the decryption part of agent's receive function
+    let inviter_requested = faber.reply_connection_request(msg_request, faber_invite_info).await;
+
+    let msg_any = faber.wait_for_message::<AriesMessage>().await.unwrap();
+    info!("Faber received message {msg_any:?}");
+
```
gmulhearn, line 105: should we show that this is where faber would call to `acknowledge_connection` in order to transition into `InviterConnection<Completed>`?

looks a bit weird that `send_message_2` takes in an inviter who is still in the `Requested` state

gmulhearn:
nit; log colors are printing a little funky for me, is this expected? 

(ran with `RUST_LOG="info" cargo run`)

image

demo/src/demo_agent.rs
```diff
@@ -0,0 +1,275 @@
...
+        let (aries_msg, sender_key) = decrypt_as_msg::<T>(self.wallet.clone(), didcomm_msg.as_slice()).await;
+        info!(
+            "{}Agent[{}] has received a message sender verkey: {sender_key}",
+            self.name, self.log_color
+        );
+        Ok(aries_msg)
+    }
+
+    pub async fn prepare_invitation(&self) -> (Invitation, PairwiseInfo) {
+        let (faber_pw_did, faber_invite_key) = self.wallet.create_and_store_my_did(None, None).await.unwrap();
```
gmulhearn, line 138: nit; not sure if you were looking for it , but you could also call `PairwiseInfo::create(wallet)` to avoid construction of `PairwiseInfo` at the end

gmulhearn (approved):
This is really awesome, nice work. preemptive approval from me; just some comments to consider. (assuming others are happy too ocourse)

gmulhearn:
`send_message_2` bugs me a bit too.. maybe something like `send_message_as_inviter` `send_message_as_invitee`? not sure

demo/src/http_client.rs
bobozaur: There was an initiative to get the message sending part to `shared_vcx` at some point, which would've maybe helped with this. I believe it was some guy Andy who tried working on it and made a PR, but it just dried out I think.

demo/src/demo_agent.rs
```diff
@@ -0,0 +1,275 @@
...
+    };
+    MsgParts::with_decorators(uuid::Uuid::new_v4().to_string(), bm, bmd)
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct MessageData {
+    message: String,
+    recipient_verkey: String,
+    sender_verkey: String,
+}
```
bobozaur, line 64: Isn't there some `EncryptionEnvelope` that handles this same thing? I'm in favor of rewriting that, don't get me wrong 😅 , but I just wanted to point out that this might be useful in more places.

demo/src/demo_agent.rs
```diff
@@ -0,0 +1,275 @@
...
+            self.log_color,
+            serde_json::to_string(&basic_message).unwrap()
+        );
+        connection
+            .send_message(&self.wallet, &AriesMessage::BasicMessage(basic_message), &HttpClient)
+            .await
+            .unwrap();
+    }
+
+    // todo: this is ugly, Connection in reasonable states should implement send_message()
```
bobozaur, line 261: It does implement `send_message()`. The method is present for states that implement the `TheirDidDoc` trait. You could use a generic to avoid the method duplication here. 

Although we might really want to consider taking the message sending outside of the state machine altogether (especially since there's also `DidExchange`) and rather have the state machine provide the info necessary to send the message (like some common trait we discussed about a while ago).

tools/simple_message_relay/src/mpsc_registry.rs
```diff
@@ -0,0 +1,69 @@
...
+    // todo: don't unwrap, return result
+    pub async fn send_msg(&self, user_id: &str, message: T) {
+        let inner = self.inner.read().await;
+        if let Some((sender, _)) = inner.get(user_id) {
+            sender.send(message).await.unwrap()
+        } else {
+            warn!("User {} not found.", user_id);
+        }
+    }
+}
```
bobozaur, line 40: I admit that I didn't checkout the PR to look through the code in an IDE but I feel like this is not a good way to go about messages.

The channels are fine. The problem I see is the web server and this shared ownership registry with interior mutability.

Unless I'm mistaken, the web server is there to allow the reuse of the HttpClient for sending messages. My curiosity is: why not use a different implementor of the `Transport` type that relies entirely on channels, thus removing the need for a web server?

Secondly, channels could be distributed as needed to their agents based on the connections between them. This would remove the need to share the whole set of senders/receivers with all the agents, but rather allow each agent to have a channel per connection which it exclusively owns.

It's a big change and I didn't think of all the implementation details so feel free to disregard it, but that might speed things up even more and make them more idiomatic.

bobozaur (changes requested):
Mostly left remarks.

I think the message sending could be improved, but that would require quite a bit of a refactor. The duplicate methods should be easy to remove though, which is why I'd request some changes.

demo/src/demo_agent.rs
Patrik-Stas: Good point, technically we could instead of `send_message` on connection have just `encrypt` which would encrypt the message using the counteparty's ddo

Then we would be truly consistent on this front with all the other protocol handlers/state machines which had their msg sending IO taken out.

tools/simple_message_relay/src/mpsc_registry.rs
Patrik-Stas: It's a fair point. Arguably for demo purposes you might want to be closer to real world, eg. using http mediator; but on the other hand, the existence of the http mediator was largely abstracted away in this demo anyway.

I'll loo into this

demo/src/demo_agent.rs
Patrik-Stas: Yeah it's similar, I think something like this along with `decrypt_as_msg` could be perhaps adopted in `aries-vcx` (with some changes, surely)

demo/src/http_client.rs
Patrik-Stas: While I agree, connection has been freed from message sending IO and demo has been freed of HTTP communications. So resolving this

demo/src/main.rs
Patrik-Stas: After refactoring, things are bit different now; message sending happens over the memory only and that been pointed out doesn't exist anymore.

demo/src/demo_agent.rs
Patrik-Stas: I would like us to actually get rid of the whole `PairwiseInfo` thing, which why I am not using it, rather creating it as necessity at the point where it's needed.
The "did" portion of `PairwiseInfo` is generally useless

demo/src/mpsc_registry.rs
```diff
@@ -0,0 +1,80 @@
+use std::{collections::HashMap, fmt::Debug, sync::Arc};
+
+use tokio::sync::{mpsc, RwLock};
+
+#[derive(Clone)]
+pub struct MpscRegistry<T: Debug> {
+    sending: Arc<RwLock<HashMap<String, mpsc::Sender<T>>>>,
+    receiving: Arc<RwLock<HashMap<String, mpsc::Receiver<T>>>>,
+}
```
bobozaur, line 9: Wasn't the previous implementation similar to this in terms of communication? Maybe I'm missing something but I don't get why the shared ownership and interior mutability is needed. 

Each agent should be in charge of its own connections (senders to other agents and receivers from other agents) and practically own them exclusively. Or is there some reason why this can't be achieved?

bobozaur (changes requested):

Patrik-Stas:
Archiving this. Instead, a new demo will be created soon on top of `aries-vcx-agent` crate

Issue #948, opened 2023-08-20
Add support for `names` in proof requests in `credx` anoncreds impl


Patrik-Stas:
https://hyperledger.github.io/anoncreds-spec/#create-presentation-request 
> 
> attr_info has the following format:
> 
> {
>     "name": <string>,
>     "names": <[string, string]>,
>     "restrictions": <restrictions>,
>     "non_revoked": <non_revoc_interval>,
> }
> All of the items are optional, but one of name or names MUST be included, and not both.
> 
> name is a string, the name of an attribute from a source credential.
> The name is case insensitive with spaces ignored.
> names is a array of strings, the names of attributes from a source credential
> The names are case insensitive with spaces ignored.
> The attribute names MUST be sourced from a single credential.

Credx implementation is failing nodejs integration tests 
https://github.com/hyperledger/aries-vcx/actions/runs/5913722928/job/16038394663?pr=927
which are using `names` in proof request such as:

```json
{
  "name": "proof-test-standard",
  "non_revoked": {
    "from": null,
    "to": 1692483689381
  },
  "nonce": "868755284748768464035775",
  "requested_attributes": {
    "attr_basic_identity": {
      "names": [
        "name",
        "last_name",
        "sex"
      ],
      "restrictions": [
        {
          "issuer_did": "V4SGRU86Z58d6TV7PBUe6f"
        }
      ]
    },
    "attr_date": {
      "name": "date",
      "restrictions": {
        "issuer_did": "V4SGRU86Z58d6TV7PBUe6f"
      }
    },
    "attr_education": {
      "name": "degree",
      "restrictions": {
        "attr::degree::value": "maths"
      }
    },
    "attr_nickname": {
      "name": "nickname",
      "self_attest_allowed": true
    }
  },
  "requested_predicates": {
    "predicate_is_adult": {
      "name": "age",
      "p_type": ">=",
      "p_value": 18,
      "restrictions": [
        {
          "issuer_did": "V4SGRU86Z58d6TV7PBUe6f"
        }
      ]
    }
  },
  "version": "1.0"
}
```

Add support for this is missing in the current credx implementation.

PR #955, opened 2023-08-24
Added 'names' attribute field parsing in credx

bobozaur:
Addresses #948 .

### DIFF
--- a/aries_vcx/src/common/anoncreds.rs
+++ b/aries_vcx/src/common/anoncreds.rs
@@ -56,6 +56,43 @@ pub mod integration_tests {
         .await;
     }
 
+    // #[cfg(feature = "modular_libs")]
+    #[tokio::test]
+    #[ignore]
+    async fn test_pool_proof_req_attribute_names() {
+        SetupProfile::run(|setup| async move {
+            let proof_req = json!({
+               "nonce":"123432421212",
+               "name":"proof_req_1",
+               "version":"0.1",
+               "requested_attributes": json!({
+                   "multiple_attrs": {
+                       "names": ["name_1", "name_2"]
+                   },
+                   "address1_1": json!({
+                       "name":"address1",
+                       "restrictions": [json!({ "issuer_did": "some_did" })]
+                   }),
+                   "self_attest_3": json!({
+                       "name":"self_attest",
+                   }),
+               }),
+               "requested_predicates": json!({
+                   "zip_3": {"name":"zip", "p_type":">=", "p_value":18}
+               }),
+            })
+            .to_string();
+
+            let anoncreds = Arc::clone(&setup.profile).inject_anoncreds();
+            let _result = anoncreds
+                .prover_get_credentials_for_proof_req(&proof_req)
+                .await
+                .unwrap();
+            println!("{_result}");
+        })
+        .await;
+    }
+
     #[tokio::test]
     #[ignore]
     async fn test_pool_revoke_credential() {
--- a/aries_vcx/src/common/anoncreds.rs
+++ b/aries_vcx/src/common/anoncreds.rs
@@ -88,7 +88,6 @@ pub mod integration_tests {
                 .prover_get_credentials_for_proof_req(&proof_req)
                 .await
                 .unwrap();
-            println!("{_result}");
         })
         .await;
     }
--- a/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
+++ b/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
@@ -739,6 +739,8 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
                     )
                 })?;
 
+                let mut credentials_json = vec![];
+
                 for _attr_name in _attr_names {
                     let _attr_name = _attr_name.try_as_str()?;
                     let attr_name = _normalize_attr_name(_attr_name);
@@ -750,17 +752,15 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
                         ._get_credentials_for_proof_req_for_attr_name(restrictions, &attr_name)
                         .await?;
 
-                    let mut credentials_json = vec![];
-
                     for (cred_id, credx_cred) in credx_creds {
                         credentials_json.push(json!({
                             "cred_info": _make_cred_info(&cred_id, &credx_cred)?,
                             "interval": non_revoked
                         }))
                     }
-
-                    cred_by_attr[ATTRS][reft.clone()] = Value::Array(credentials_json);
                 }
+
+                cred_by_attr[ATTRS][reft] = Value::Array(credentials_json);
             } else {
                 let _attr_name = requested_val.try_get("name")?;
                 let _attr_name = _attr_name.try_as_str()?;
--- a/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
+++ b/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
@@ -142,27 +142,36 @@ impl IndyCredxAnonCreds {
     async fn _get_credentials_for_proof_req_for_attr_name(
         &self,
         restrictions: Option<&Value>,
-        attr_name: &str,
+        attr_names: Vec<String>,
     ) -> VcxCoreResult<Vec<(String, CredxCredential)>> {
-        let attr_marker_tag_name = _format_attribute_as_marker_tag_name(attr_name);
+        let mut attrs = Vec::new();
 
-        let wql_attr_query = json!({
-            attr_marker_tag_name: "1"
-        });
+        for name in attr_names {
+            let attr_marker_tag_name = _format_attribute_as_marker_tag_name(&name);
+
+            let wql_attr_query = json!({
+                attr_marker_tag_name: "1"
+            });
+
+            attrs.push(wql_attr_query);
+        }
 
         let restrictions = restrictions.map(|x| x.to_owned());
 
         let wql_query = if let Some(restrictions) = restrictions {
             match restrictions {
                 Value::Array(mut arr) => {
-                    arr.push(wql_attr_query);
+                    arr.extend(attrs);
                     json!({ "$and": arr })
                 }
-                Value::Object(obj) => json!({ "$and": vec![wql_attr_query, Value::Object(obj)] }),
-                _ => wql_attr_query,
+                Value::Object(obj) => {
+                    attrs.push(Value::Object(obj));
+                    json!({ "$and": attrs })
+                }
+                _ => json!(attrs),
             }
         } else {
-            wql_attr_query
+            json!(attrs)
         };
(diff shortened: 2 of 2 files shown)


aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -731,27 +731,59 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
                     format!("Unknown referent: {}", reft),
                 ))?;
 
-            let _attr_name = requested_val.try_get("name")?;
-            let _attr_name = _attr_name.try_as_str()?;
-            let attr_name = _normalize_attr_name(_attr_name);
+            if let Some(_attr_names) = requested_val.get("names").and_then(|v| v.as_array()) {
```
Patrik-Stas, line 734: I'd suggest to also make sanity check on name/names first - if none, or both are specific, it's invalid proof request (according to spec). Currently if:
-  both are specified, we don't error but only prioritize `names`
- if none is specified, we fail on line 765 with some error, but I imagine we could do a nicer one (signalling an invalid proof request for reason X).

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -731,27 +731,59 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
 
-            let credx_creds = self
-                ._get_credentials_for_proof_req_for_attr_name(restrictions, &attr_name)
-                .await?;
+                for _attr_name in _attr_names {
+                    let _attr_name = _attr_name.try_as_str()?;
+                    let attr_name = _normalize_attr_name(_attr_name);
 
-            let mut credentials_json = vec![];
+                    let non_revoked = requested_val.get("non_revoked"); // note that aca-py askar fetches from proof_req json
```
Patrik-Stas, line 748: Right, it should use the top level `non_revoked` value by default, but get override if specified on specific <requested attribute> level. And it seems like right now it's only looking on the <requested attribute> level, is that right?

aries_vcx/src/common/anoncreds.rs
```diff
@@ -56,6 +56,43 @@ pub mod integration_tests {
...
+               }),
+            })
+            .to_string();
+
+            let anoncreds = Arc::clone(&setup.profile).inject_anoncreds();
+            let _result = anoncreds
+                .prover_get_credentials_for_proof_req(&proof_req)
+                .await
+                .unwrap();
+            println!("{_result}");
```
Patrik-Stas, line 91: forgotten print

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -731,27 +731,59 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
                     format!("Unknown referent: {}", reft),
                 ))?;
 
-            let _attr_name = requested_val.try_get("name")?;
-            let _attr_name = _attr_name.try_as_str()?;
-            let attr_name = _normalize_attr_name(_attr_name);
+            if let Some(_attr_names) = requested_val.get("names").and_then(|v| v.as_array()) {
+                let _attr_names = requested_val.try_get("names")?.as_array().ok_or_else(|| {
```
Patrik-Stas, line 735: It feels like a code smell because the `ok_or_else` err handling should be never reached. Perhaps would be cleaner to do a match on tuple of `(requested_val.get("names"), requested_val.get("name"))` and that would also sort out the invalid cases referred to in the comment above.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -731,27 +731,59 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
-            cred_by_attr[ATTRS][reft] = Value::Array(credentials_json);
+                    for (cred_id, credx_cred) in credx_creds {
+                        credentials_json.push(json!({
+                            "cred_info": _make_cred_info(&cred_id, &credx_cred)?,
+                            "interval": non_revoked
+                        }))
+                    }
+                }
+
+                cred_by_attr[ATTRS][reft] = Value::Array(credentials_json);
```
Patrik-Stas, line 763: Subjective and not directly relevant to this PR, just noticed - the `reft` variable name, would prefer full wording `referent`

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -731,27 +731,59 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
+                        .await?;
 
-            cred_by_attr[ATTRS][reft] = Value::Array(credentials_json);
+                    for (cred_id, credx_cred) in credx_creds {
+                        credentials_json.push(json!({
+                            "cred_info": _make_cred_info(&cred_id, &credx_cred)?,
+                            "interval": non_revoked
+                        }))
+                    }
+                }
```
Patrik-Stas, line 761: Hmm, I haven't checked how this is done in other implementations, but it seems like something might be missing. If verifier specified `names: ["name", "surname"]` it means that fields from which these fields are selected from are originating from a single credential https://hyperledger.github.io/anoncreds-spec/#presentation-request-example
eg. not mixing "name" from one credential and "surname" from another.

While this code could coincidentally (and in practice currently even likely) work if there's not more alternative credentials in the wallet, it seems like prover is not trying to assure that property.

We don't necessarily have to address this right now, but lets at least create issue to track this.

Patrik-Stas (changes requested):
Left some comments, but happy to see this
https://github.com/hyperledger/aries-vcx/pull/927

now fully passing on top of this PR!

aries_vcx/src/common/anoncreds.rs
bobozaur: Fixed

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: Most of this was done by @gmulhearn . I only did the issuer.

bobozaur:
@Patrik-Stas Thanks for taking the time to review this. I want to highlight some things:
- This was "hastily" done to fix the issue. There are far better ways to handle this.
- Most of the implementation, apart from the issuer (which was done by me) was done by @gmulhearn . I'm not all that familiar with all the business logic.
- There's a lot of code smell throughout the entire credx implementation. I would not tweak certain things but rather rewrite it all (it's my motto, after all 😆 ).

Given these things, my approach would be:
- Get this into a suitable working state (if it's not already) and merge it.
- Remove vdrtools anoncreds implementation in favor of credx anoncreds.
- Rewrite ledger trait to add strong typing
- Rewrite credx anoncreds so that:
    - It uses strong typing introduced in the ledger trait refactor and brings its own
    - It is readable and maintainable
    - It is more performant, mainly due to stronger typing and not needing to serialize/deserialize stuff in every single function call.

For instance, the `name` or `names` situation could be handled in an enum type, thus enforcing that a single field of the two are provided. But I'd rather understand all the business logic, get all the prerequisites regarding strong typing done, and then rewrite the credx anoncreds impl rather than patching this single spot in the code pile we have now.

I'd like to hear what you, or the others, think about this roadmap.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
gmulhearn: ye IMO this is the whole point of `names`; requiring that multiple attributes are presented from the same credential. `credentials_json` should only be populated with credentials which have ALL the attributes (names) present

gmulhearn: Hmm no that doesn't fix it. Consider a request for `names: ["a", "b", "c"]`, and a wallet with `cred1 { a, d }`, `cred2 { b, e }`, and `cred3 { c, f }`.

The tweaked implementation with `all_found` would result in `cred1`, `cred2` and `cred3` being added to `credentials_json` and included. Where really, none of the credentials satisfy the requirement of having all 3 attribute names (a, b, and c), so the credentials for that referent should be empty.

A credential should only be added to the list if it has ALL the `name`/`names` as attributes on the credential, AND it satisfies any restrictions applied to it. 

The ideal way to do that IMO is to rewrite `_get_credentials_for_proof_req_for_attr_name` a bit, such that it only needs to be called once per referent, and it takes in an array of `attr_names`, and the WQL query will search for credentials which have ALL those attributes (currently it just checks for 1 attr_name passed in).

gmulhearn: Anyway, will leave it up to you guys whether that is scope for this PR or for sometime else; as you've stated, the impl is far from perfect and a rewrite is probs best... the original impl was meant to act as a 'base' for the sake of getting it in there, so it's great that you guys are readdress the impl and improving

Patrik-Stas:
@bobozaur that's a great plan! Sounds good to me 👍

Patrik-Stas (approved):

bobozaur:
> @bobozaur sounds good, although I am not sure what would you rewrite about the ledger trait (other than adding typing (which would ideally be the case for all of the `aries_vcx_core` components) )
> 
> Anyway, in scope of this PR I am fine with merge

We could still add the associated types. It looks like @gmulhearn has the same concern so that we don't end up with a very rigid interface that only works with credx.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
bobozaur: Okay I had another attempt at fixing this. Your insights were very valuable, thank you! I hope the query composition is fine, but I assume we'll see in the CI.

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -142,27 +142,36 @@ impl IndyCredxAnonCreds {
...
-                _ => wql_attr_query,
+                Value::Object(obj) => {
+                    attrs.push(Value::Object(obj));
+                    json!({ "$and": attrs })
+                }
+                _ => json!(attrs),
             }
         } else {
-            wql_attr_query
+            json!(attrs)
```
gmulhearn, line 174: I think in these cases, we'd still want things to be ANDed together;

I.e. the query is for where 'attr1=="1" && attr2=="2" && attr3=="3" etc'. So I think it's still json({"$and": attrs}).

Either way, these branches are a bit of an edge case. I might make ticket about the wallet queries in this class, we really should be using the Indy-wql crate, or create our own utility for query construction

Patrik-Stas (approved):

gmulhearn (approved):
lgtm assuming we make another ticket to address a revision to the wql structures

mirgee (approved):

PR #960, opened 2023-08-29
Fix profile building in tests, considering vdrtools is default feature

Patrik-Stas:
- When profile is selected dynamically based on feature flags (in testing), it would previously always select always `vdrtools` profile, regardless of `modular_libs` feature being enabled.
- This bug was probably brought in some of the relatively recent test setup / feature refactoring.
- Nevertheless the modular libs (which were apparently not being tested in CI since introduction of the bug) are passing fine ✅ 

Proper solution though, should be removing `vdrtools` as default feature
- had a quick shot at it, but it's a bit trickier, as tests require vdrtools wallet regardless of whether we are running with credx/vdrtools anoncreds
- it's possible to address, but figured it'll be easier just to wait out for vdrtools anoncreds to die out

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -336,6 +336,7 @@ jobs:
         run: |
           cargo test --manifest-path="wallet_migrator/Cargo.toml";
           RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs -- --include-ignored;
+          RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs_revocations -- --include-ignored;
 
   test-integration-libvcx:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -336,7 +336,6 @@ jobs:
         run: |
           cargo test --manifest-path="wallet_migrator/Cargo.toml";
           RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs -- --include-ignored;
-          RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs_revocations -- --include-ignored;
```
gmulhearn-anonyome, line 339: how come we're removing this? is it now failing?

Patrik-Stas: good catch, reverted. 

I mistakenly thought it's working on prefix basis, thinking I've removed duplicated test execution, while obviously it's specifying specific test files.

gmulhearn (approved):
LGTM assuming happy CI 👌

PR #961, opened 2023-08-29
Prover: fix credx credential retrieval

Patrik-Stas:
- Fixes retrieval of credentials for prover. When multiple restriction objects are specified, they should be treated as `OR` instead of `AND`. Verifier is asking prover to satisfy at least 1 restriction object, not all of them.

- Adds test `test_agency_pool_it_should_select_credentials_for_satisfiable_restriction` which was failing without modifications to WQL made in the PR. Passes with the modifications. See anoncreds spec https://hyperledger.github.io/anoncreds-spec/#restrictions

> A boolean expression is formed by ORing and ANDing the source credential properties. The following JSON is an example. Any of the source credential properties listed above can be used in the expression components:
> 
>       "restrictions": [
>         {
>           "issuer_did": "<did>",
>           "schema_id": "id"
>         },
>         {
>           "cred_def_id" : "<id>",
>           "attr::color::marker": "1",
>           "attr::color::value" : "red"
>         }
>       ]
> The properties in each list item are AND’d together, and the array elements are OR’d together. As such, the example above defines the logical expression:
> 
> The attributes must come from a source verifiable credential such that:
>    issuer_did = <did> AND
>      schema_id = <id>
>    OR
>    cred_def_id = <id> AND
>       the credential must contain an attribute name "color" AND
>       the credential must contain an attribute name "color" with the attribute value "red"

- Test refactoring also uncovered that test `test_agency_pool_it_should_fail_to_select_credentials_for_predicate` is not working with credx implementation; see: https://github.com/hyperledger/aries-vcx/issues/962

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -336,6 +336,7 @@ jobs:
         run: |
           cargo test --manifest-path="wallet_migrator/Cargo.toml";
           RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs -- --include-ignored;
+          RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs_revocations -- --include-ignored;
 
   test-integration-libvcx:
     needs: workflow-setup
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -336,7 +336,6 @@ jobs:
         run: |
           cargo test --manifest-path="wallet_migrator/Cargo.toml";
           RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs -- --include-ignored;
-          RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs_revocations -- --include-ignored;
 
   test-integration-libvcx:
     needs: workflow-setup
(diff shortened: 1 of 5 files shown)


aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -160,12 +160,13 @@ impl IndyCredxAnonCreds {
...
+                    attrs.push(restrictions_wql);
+                    json!({ "$and": attrs })
                 }
-                Value::Object(obj) => {
-                    attrs.push(Value::Object(obj));
+                Value::Object(restriction) => {
+                    attrs.push(Value::Object(restriction));
                     json!({ "$and": attrs })
                 }
                 _ => json!(attrs),
```
gmulhearn, line 172: I suppose this case is us saying "ignore the restrictions if they aren't a valid restriction json type (obj or array)".

I wonder more generally if we should just throw an error here; "unsupported restriction". 

If not, I'm not sure json!(attrs) is correct for ignoring restrictions.

I think it should be json!({$and: attrs})

To indicate that ALL attr in the attrs array must be satisfied

gmulhearn: Also a couple lines below (175) has the same issue

gmulhearn (changes requested):
LGTM; but would be good to fix up my comment while you're in this area.

Good to merge after that

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
Patrik-Stas: should be addressed, please check

gmulhearn: Sorry for delay, your fox matches my understanding, thanks

gmulhearn (approved):
Fix from my last comment looks good

.github/workflows/main.yml
```diff
@@ -336,7 +336,6 @@ jobs:
         run: |
           cargo test --manifest-path="wallet_migrator/Cargo.toml";
           RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs -- --include-ignored;
-          RUST_TEST_THREADS=1 CARGO_INCREMENTAL=0 TEST_POOL_IP=127.0.0.1 cargo test --manifest-path="aries_vcx/Cargo.toml" -F migration --test test_creds_proofs_revocations -- --include-ignored;
```
mirgee, line 339: I am sorry, why was this removed?

mirgee (approved):

PR #967, opened 2023-09-06
Eradicate mediated connection from issuance and presentation integration tests, part 1

mirgee:
Reduces the number of issuance and presentation integration tests relying on mediated connection to exchange messages instead of exchanging them directly between the SM handlers, with the goal of removing their dependency on agency client and MediatedConnection and eventually making the integration tests runnable without a mediator.

This change results in a minor test speedup, however the main bottleneck is schema / cred. def. creation, which can be mediated by sharing schemas / credential definitions between tests wherever possible.


aries_vcx/tests/test_creds_proofs.rs
```diff
@@ -628,32 +622,30 @@ mod tests {
...
-                &mut verifier,
-                &institution.profile.inject_wallet(),
-                &institution.profile.inject_anoncreds_ledger_read(),
-                &institution.profile.inject_anoncreds(),
-                &institution.agency_client,
-                &institution_to_consumer,
-            )
-            .await
-            .unwrap();
+            verifier
```
Patrik-Stas, line 641: ❤️

aries_vcx/tests/utils/scenarios.rs
```diff
@@ -233,68 +236,40 @@ pub mod test_utils {
...
             .unwrap();
-        send_closure(msg_response).await.unwrap();
-        tokio::time::sleep(Duration::from_millis(1000)).await;
-        holder
+        (holder, cred_request)
     }
 
-    pub async fn send_cred_proposal(
+    pub async fn create_credential_proposal(
         alice: &mut Alice,
```
Patrik-Stas, line 268: unused argument

Patrik-Stas (dismissed):
this is amazing 🚀 Significant improvement in readability and mental overhead reading the tests, now that connections are gone.

Left some comments but generally ✅

aries_vcx/tests/test_creds_proofs.rs
```diff
@@ -599,16 +595,8 @@ mod tests {
...
-                create_connected_connections(&mut consumer, &mut institution).await;
             let (schema_id, cred_def_id, _rev_reg_id, _cred_def, _rev_reg, _credential_handle) =
-                issue_address_credential(
-                    &mut consumer,
-                    &mut institution,
-                    &consumer_to_institution,
-                    &institution_to_consumer,
-                )
-                .await;
+                issue_address_credential(&mut consumer, &mut institution).await;
```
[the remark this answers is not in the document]
mirgee: 😁

Patrik-Stas:
Goddamn Jimmy, this is some serious gourmet shiiiit! 
image

I am even looking at `integration-tests-cntd` branch which seems to be trimmed down even a lot more than this. I might be getting ahead of myself, because I don't know what's your working strategy, but noticed on that branch, mysql tests are also removed - I'd keep that; do you want to remove them?

One more intermediate remark, and again maybe you already plan doing something like this, just sharing my immediate thoughts - relating to our private discussion yesterday about `scenario.rs`:
- I would suggest to move methods which related to a single agent as an Agent's impl method (and ideally merge Alice & Faber into one singular Agent abstraction); such as 
```
pub async fn accept_proof_proposal(
    faber: &mut Faber,
    verifier: &mut Verifier,
    presentation_proposal: AriesMessage,
) -> AriesMessage {
```

- While keeping in workflows methods which contain 2-agent interaction, such as
```
pub async fn _exchange_credential(
    consumer: &mut Alice,
    institution: &mut Faber,
    credential_data: String,
    cred_def: &CredentialDef,
    rev_reg: &RevocationRegistry,
    comment: Option<&str>,
) -> Issuer {
```

Patrik-Stas (approved):

PR #954, opened 2023-08-24
Uniffi demo qr scan

swaptr:
Merge only after #896 
This PR elaborates the Uniffi demo and API.

PR #968, opened 2023-09-06
Uniffi remove autogenerated bindings

swaptr:
Autogenerated uniffi bindings can be untracked by default. Merge after #954.


gmulhearn (approved):
nice thanks. FYI @Patrik-Stas

PR #970, opened 2023-09-07
Uniffi simple message relay

swaptr:


uniffi_aries_vcx/core/Cargo.toml
```diff
@@ -19,6 +19,7 @@ path = "uniffi-bindgen.rs"
 uniffi = "0.23.0"
 aries-vcx = { path = "../../aries_vcx" }
 tokio = { version = "1.24.1" }
+libvdrtools = { path = "../../libvdrtools", features=["default"] }
```
gmulhearn, line 22: why is this needed??

uniffi_aries_vcx/core/src/core/unpack_message.rs
```diff
@@ -0,0 +1,34 @@
...
+        let wallet = profile_holder.inner.inject_wallet();
+        let unpacked_bytes = wallet.unpack_message(&packed_bytes).await?;
+        let unpacked_string = match String::from_utf8(unpacked_bytes) {
+            Ok(str) => str,
+            Err(err) => todo!(),
+        };
+        let value = match serde_json::from_str::<UnpackMessage>(&unpacked_string) {
+            Ok(str) => str,
+            Err(err) => todo!(),
+        };
```
gmulhearn, line 31: can just escape `serde` errors with `?`. as we have already defined a serde error -> VcxUniFFI error transformer

uniffi_aries_vcx/core/src/core/unpack_message.rs
```diff
@@ -0,0 +1,34 @@
+use std::sync::Arc;
+
+use aries_vcx::errors::error::{AriesVcxError, AriesVcxErrorKind};
+use vdrtools::domain::crypto::pack::UnpackMessage;
```
gmulhearn, line 4: preferably create your own struct for this rather than importing from libvdrtools. we should try to avoid direct libvdrtools dependencies

gmulhearn: you could just copy paste libvdrtool's definition for the struct;
```
#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
pub struct UnpackMessage {
    pub message: String,
    pub recipient_verkey: String,
    pub sender_verkey: Option<String>,
}
```

uniffi_aries_vcx/core/src/core/unpack_message.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use super::profile::ProfileHolder;
+use crate::{errors::error::VcxUniFFIResult, runtime::block_on};
+
+pub fn unpack_message(profile_holder: Arc<ProfileHolder>, packed_msg: Vec<u8>) -> VcxUniFFIResult<Vec<u8>> {
```
gmulhearn, line 9: if you wanted, could do it this way:
```
pub fn unpack_message(profile_holder: Arc<ProfileHolder>, packed_msg: String) -> VcxUniFFIResult<UnpackMessage> {
...
}
```

in kotlin will look like:
```
data class Unpackmessage(message: String, recip: String, sender: Option<String>)
```

uniffi_aries_vcx/core/Cargo.toml
swaptr: Was used for accessing the `UnpackMessage` trait for deserialization. Now fixed with an implementation of it inside the uniffi crate.

uniffi_aries_vcx/demo/app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,4 @@
 <resources>
     <string name="app_name">demo</string>
-</resources>
\ No newline at end of file
+    <string name="serviceEndpoint">https://b199-27-57-116-96.ngrok-free.app</string>
```
swaptr, line 3: Does this need to be set as an env var? Or any other secure and flexible way?

gmulhearn: to be honest, i'd just make a `Constants.kt` file, and put this value in as a `BASE_RELAY_ENDPOINT` const val, and add a comment above it explaining what it means

uniffi_aries_vcx/demo/app/src/main/java/org/hyperledger/ariesvcx/ScanScreen.kt
```diff
@@ -66,9 +74,13 @@ fun ScanScreen(connection: Connection, profileHolder: ProfileHolder) {
                         )
                         connection.sendRequest(
                             profileHolder,
-                            "https://google.com",
+                            "$BASE_RELAY_ENDPOINT/send_user_message/${walletConfig.walletKey}",
```
gmulhearn, line 77: using walletKey as the unique ID is very suspicious; wallet key is meant to be a unique password for your device wallet. I'd recommend we just use something else hardcoded as the ID. e.g. `demo-user-1`

swaptr: Oh no I dont know what I was thinking, This is supposed to be a user id, right?

gmulhearn (dismissed):
Nice work. This will be good to merge after we stop using `walletKey` as the user id; for now we can just use a hardcoded string: `demo-user-1` for example.

And also as discussed, in a later PR we will look into cleaning up the app logic by putting a lot of the logic into a global "controller" class. this will let us clean up some of the HTTP polling logic currently within the HomeScreen

uniffi_aries_vcx/demo/app/src/main/java/org/hyperledger/ariesvcx/HomeScreen.kt
```diff
@@ -1,37 +1,76 @@
...
-        storageCredentials = null,
-        rekey = null,
-        rekeyDerivationMethod = null
-    )
+    var flagKeepFetching by remember {
+        mutableStateOf(true)
+    }
+
+    val request = Request.Builder()
+        .url("$BASE_RELAY_ENDPOINT/pop_user_message/${walletConfig.walletKey}")
```
gmulhearn, line 49: still an issue here with walletKey being used. should match the other string. you could also make it into some constant string;

`const val RELAY_USER_ID = "demo-user-1"

gmulhearn (approved):

PR #974, opened 2023-09-11
Eradicate mediated connection from issuance and presentation integration tests, part 2

mirgee:
Continuation of #967.

* Completely removes usage of `MediatedConnection` and `agency_client` from any and all integration tests.
* Removes legacy agency tests.
* MySQL tests are preserved, where keypair generation is used to test MySQL connection.
* Redundant `integration_tests` and `tests` modules were removed.
* Alice and Faber were significantly pruned, where the plan is to replace them with a singular test agent impl.


aries_vcx/tests/utils/devsetup_faber.rs
```diff
@@ -58,48 +49,23 @@ pub async fn create_faber_trustee(genesis_file_path: String) -> Faber {
...
-    profile
-        .inject_anoncreds()
-        .prover_create_link_secret(DEFAULT_LINK_SECRET_ALIAS)
-        .await
-        .unwrap();
-    Faber::setup(profile, genesis_file_path, public_did).await
+    create_faber_from_seed(&generate_random_seed(), genesis_file_path).await
 }
 
 impl Faber {
```
Patrik-Stas, line 55: The `create_schema` is only used in 2 tests,  `pairwise_info` seems to be unused (?) - seems like fairly easy win to remove these 2 as well.
Then the only difference between Alice and Faber would be `institution_did`, which is used quite heavily. 

But it could be another PR, just noting

mirgee: This is exactly what I am doing right now :)

Patrik-Stas (dismissed):
amazing again :-) Huge burden droped off once more

PR #976, opened 2023-09-13
Initial mediator work

nain-F49FF806:
Display an OOB invitation. Also presented as QR!


nain-F49FF806:
#977 does more on top of these changes.  Consider that PR instead.

PR #980, opened 2023-09-15
add cargo ndk build script for uniffi

swaptr:
Credits to @gmulhearn  for simplifying the build process by using cargo-ndk. This PR integrates those changes and adds relevant documentation. 
Note:
The "other" scripts will be untracked once work on CI (UniFFI) is finished.


uniffi_aries_vcx/README.md
```diff
@@ -1,24 +1,55 @@
...
+```
+
+eg. for Android ARM64 on host linux platform.
+
+```bash
+rustup target add aarch64-linux-android
+```
+
+More documentation can be found [here](https://rust-lang.github.io/rustup/cross-compilation.html). Information on supported platforms can be found [here](https://doc.rust-lang.org/nightly/rustc/platform-support.html).
+
```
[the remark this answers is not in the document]
gmulhearn: https://developer.android.com/studio/projects/install-ndk#default-version

uniffi_aries_vcx/README.md
```diff
@@ -1,24 +1,55 @@
...
+3. Make sure your ANDROID_SDK and ANDROID_NDK_HOME environment variables point to the SDK and NDK roots respectively.
```
gmulhearn, line 43: Maybe just change this to: 
> if cargo-ndk cannot find the NDK installed on your machine, you may have to configure environment variables as described here: https://github.com/bbqsrc/cargo-ndk#usage

gmulhearn: since most users won't have to configure environment variables if they installed with android studio (as we recommend)

gmulhearn (approved):

Issue #947, opened 2023-08-20
Remove `tokio` as aries-vcx dependency


Patrik-Stas:
- In `aries_vcx` crate, `tokio` is required dependency. However it is only used in tests and test_utils submodules. Make sure `tokio` is dev-dependency only

JamesKEbert:
Addressed in #981

PR #981, opened 2023-09-15
Minimize dependency on tokio

Patrik-Stas:
Addressing https://github.com/hyperledger/aries-vcx/issues/947

- In `aries_vcx`, swapped tokio async sleeps for std blocking sleeps. This is okay because the sleep are only in test setup code, and these we run tests in single thread. If we gain capability to run tests in parallel, with a bit of effort, this can easily be update back to tokio sleeps (just need to properly feature flag the test setup code, which isn't the case today)

- in `aries_vcx_core`, swapped Mutex around cache for std lock. There was no need for tokio mutex, as the locks are not held across an await point
- in `aris_vcx_core`, the tokio import dependency is still declared, but limited to `sync` feature (for `upshot` module use by indy client implementations (`indy_vdr` tools internals require passing callbacks, not sure how difficult would those be to convert to async functions instead ) )

There's no changes visible to lock file because we declare dependency on tokio in other workspace projects (like uniffi, libvcx), but in external projects consuming `aries-vcx`, this should trim dependency tree (and likelyhood of dependency conflicts)

### DIFF
--- a/aries_vcx/Cargo.toml
+++ b/aries_vcx/Cargo.toml
@@ -51,7 +51,6 @@ uuid = { version = "0.8", default-features = false, features = ["v4"] }
 strum = "0.16.0"
 strum_macros = "0.16.0"
 derive_builder = "0.10.2"
-tokio = { version = "1.20.4" }
 thiserror = "1.0.37"
 url = { version = "2.3", features = ["serde"] }
 
--- a/aries_vcx/src/common/proofs/verifier/verifier.rs
+++ b/aries_vcx/src/common/proofs/verifier/verifier.rs
@@ -277,7 +277,7 @@ pub mod integration_tests {
             false,
         )
         .await;
-        tokio::time::sleep(Duration::from_millis(500)).await;
+        std::thread::sleep(Duration::from_millis(500));
 
         let cred_id = create_and_write_credential(
             anoncreds_issuer,
--- a/aries_vcx/src/common/test_utils.rs
+++ b/aries_vcx/src/common/test_utils.rs
@@ -46,7 +46,7 @@ pub async fn create_and_write_test_schema(
         .publish_schema(&schema_json, submitter_did, None)
         .await
         .unwrap();
-    tokio::time::sleep(Duration::from_millis(1000)).await;
+    std::thread::sleep(Duration::from_millis(500));
     Schema::create_from_ledger_json(&schema_json, "", &schema_id).unwrap()
 }
 
--- a/aries_vcx_core/Cargo.toml
+++ b/aries_vcx_core/Cargo.toml
@@ -31,7 +31,7 @@ thiserror = "1.0.40"
 lazy_static = "1.4.0"
 derive_builder = "0.12.0"
 uuid = { version = "1.3.0", default-features = false, features = ["v4"] }
-tokio = { version = "1.20" }
+tokio = { version = "1.20.5", features = ["sync"], default-features = false }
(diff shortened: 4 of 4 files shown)


aries_vcx_core/src/ledger/response_cacher/in_memory/mod.rs
```diff
@@ -40,7 +42,7 @@ impl ResponseCacher for InMemoryResponseCacher {
         let id = id.to_string();
         let obj = serde_json::to_string(&obj)?;
 
-        let mut cache = self.cache.lock().await;
+        let mut cache = self.cache.lock()?;
```
bobozaur, line 45: There are not the same thing and, unlike other changes which really just target sequential single-threaded tests, this is production code.

The `std::sync::Mutex` is blocking and will block the entire runtime thread while waiting. The `tokio` one is designed to give back control to the runtime if the lock cannot be acquired, allowing other things to run while waiting.

bobozaur (changes requested):
I'm strongly against the in memory cacher getting changed to the sync `Mutex` as that will affect performance in production code.

I don't have a strong opinion about the tests though, although I think it might be better to simply use `tokio` as a `dev-dependency` instead of switching up the waiting tasks. 

While negligible, it should still be better to have the waiting orchestrated by the runtime as opposed to putting the thread to sleep and being at the kernel's mercy.

aries_vcx_core/src/ledger/response_cacher/in_memory/mod.rs
Patrik-Stas: Thanks for catching this! Would been a disaster having this merged

aries_vcx/Cargo.toml
```diff
@@ -51,7 +51,6 @@ uuid = { version = "0.8", default-features = false, features = ["v4"] }
 strum = "0.16.0"
 strum_macros = "0.16.0"
 derive_builder = "0.10.2"
-tokio = { version = "1.20.4" }
```
bobozaur, line 54: Why not keep it as a `dev-dependency` though, leaving everything else unchanged?

Patrik-Stas: not quite possible yet ... it's a part of bigger problem, and a dilemma @mirgee has run into while doing test refactoring

There's test setup methods in `aries_vcx/src/common/test_utils.rs` such as `create_and_write_test_cred_def` which are used by both:
- tests `#[cfg(test)]` under `aries-vcx/src` - these are often esentially integration tests, but do not yet reside in `aries-vcx/tests`
- integration tests under `aries-vcx/tests`
And for that reason `aries_vcx/src/common/test_utils.rs` cannot be flagged as `#[cfg(test)]` and therefore couldn't access tokio as dev dependency.

So the right solution is to do some further restructuring, move tests around etc. But that would be a different, bigger task than what I am doing here.

bobozaur: Okay, I see. 

But in that case I personally disagree with this PR. If everyone else agrees that this is good, feel free to merge it, but I really see this as shoving a problem under the rug. 

I think the proper way that you mentioned is indeed the right solution and should be what gets done.

mirgee: I agree with @bobozaur here, which is the reason why I waited with my review to see what others think about these changes.

Patrik-Stas: Fair enough. I'll close and create tracking issue

PR #982, opened 2023-09-15
Connection protocol: add encrypt_message, remove auto-problem-report sending

Patrik-Stas:
Addition:
- Add `encrypt_message` on `impl<I, S> Connection<I, S> where S: TheirDidDoc`

Breaking change:
- For `InviteeConnection<Requested>` method `handle_response`, remove automatic problem-report sending in case of error processing `connection-response` message. The caller should, like in other protocols, inspect Error result returned and decide to send (or not) problem report
- Same applies for `InviterConnection<Invited> ` method `handle_request` processing of `connection-request`. 
- Remove methods defined on `Connection` for to dispatch problem report messages


gmulhearn (approved):
lgtm

PR #986, opened 2023-09-19
Fix lints and improve formatting

bobozaur:
Adds more comprehensive formatting settings as well as tweaking `clippy` and `rustc` lints.

The smaller crates have been adjusted and their lints fixed, but bigger ones like `aries-vcx` or `libvcx-core` will be treated independently in different PRs since they are the main sources for conflicts.

The declared lints are fairly opinionated, but it's a list that I've been building and use with pretty much all of my projects. They are obviously up for discussion and modifications.


.rustfmt.toml
```diff
@@ -0,0 +1,14 @@
+# rust-analyzer can be configured with nightly rustfmt
+# i.e. for vscode in settings.json:
+#    "rust-analyzer.rustfmt.extraArgs": [
+#        "+nightly"
+#    ],
+unstable_features = true
```
mirgee, line 6: Do we really want to enable `unstable_features` in rustfmt? My worry is only that a compliant code might become noncompliant with another nightly release as unstable features are added and removed. But perhaps that's a misguided concern?

bobozaur: A lot of really useful things are unavailable in stable. Pretty much every setting is unstable, but I've used them for quite some time and can't really remember stumbling into false positives or deprecated features.

bobozaur: There's one macro in `aries-vcx` that's been bothering me with the formatting, but that was happening even before this PR (which points to a problem in `rustfmt` itself rather than the unstable feature). It's the `get_attach_as_string` which I decided to mark as `#[rustfmt::skip]` to avoid having to deal with it every time that file gets formatted/saved.

It's probably because `rustfmt` does not handle `let ... else {}` blocks (which are fairly new) that well right now, perhaps in particular when within a macro's body.

mirgee: Alright, let's try it, we can always easily change this setting in case we do run into issues.

Cargo.toml
```diff
@@ -38,3 +38,32 @@ edition = "2021"
...
+pedantic = "warn"
+struct_excessive_bools = "allow"
+trivially_copy_pass_by_ref = "allow"
+must_use_candidate = "allow"
+match_bool = "allow"
+no_effect_underscore_binding = "allow"
+module_name_repetitions = "allow"
+
+[lints.rust]
+meta_variable_misuse = "warn"
```
Patrik-Stas, line 53: can we add job which would consider every warning to be error instead? Just to make sure we won't accumulate warnings over time again.

Patrik-Stas (changes requested):
lgtm, but left 1 comments

Cargo.toml
bobozaur: I thought we already touched down on this.

This can be controlled through a compiler flag which I plan on adding in a future PR. If I add it now, other branches (including this one) will fail in CI because there still are lints around.

Patrik-Stas (approved):

gmulhearn (approved):

mirgee (approved):

PR #987, opened 2023-09-20
feat: Cred Issuance V2 APIs

gmulhearn-anonyome:
# Overview

v1: https://github.com/hyperledger/aries-rfcs/blob/main/features/0036-issue-credential/README.md

v2: https://github.com/hyperledger/aries-rfcs/blob/main/features/0453-issue-credential-v2/README.md

v2 of issue-credential makes the protocol flow work for 'any' credential format (anoncreds, json-ld w3c, etc), whereas v1 has a lot of fields which are specific to anoncreds.

they do this by hiding away credential-format-specific data into attachments, and then we have a `formats` field in the messages to specify the format being used.

the v2 specific has tables of 'format registries' which contain links to the expected attachment payloads for different formats (anoncreds, w3c etc).

To cater for this generic format concept in our state machines, I've tried to use generics `HolderCredentialIssuanceFormat` & `IssuerCredentialIssuanceFormat`, which are traits that are responsible for creating the attachment payloads (and returning metadata) for a particular format. I've created an Anoncreds impl of these traits for proof of concept purposes. These anoncreds impls reuse a lot of existing infrastructure from our v1 anoncreds state machines.

The idea is that in the future aries-vcx could include more formats, or consumers can go ahead and create their own format impls outside of aries-vcx. This way, aries-vcx does not have a hard requirement to keep up with the (potentially) growing list of formats in the registry, as consumers can create their own in the meantime.

This potential for growth and supporting generic formats is why I've gone with generics instead of defined enum variants for different formats; i think enum variants would lock consumers into only using a set of formats that we support.

~Note that HolderV2 and IssuerV2 assume that you are using a single format family (anoncreds, json-ld, etc) thru out the whole protocol, and they also assume that you are using single attachments per message. The spec implies that you can have multiple attachments, potentially with different formats, within single messages, however i am yet to see a use-case for this or much discussion about it. For comparison, aca-py seems to assume single attachments and of the same format thru out the protocol. All in all, this confusion around what is 'potentially possible' with this spec lead me just to implement HolderV2 and IssuerV2 under those assumptions. Perhaps if it does become a use case, we'd need to make a breaking change to these typestate machines, however for now i'd like to keep it simple; i can't even begin to imagine how you'd generically deal with all the possible permutations when using multi attachments and multi formats...~

_While the above is true, i believe we agreed that single format is acceptable for now until we have a real use case. discussion [here](https://github.com/hyperledger/aries-rfcs/issues/796)_

# Open Qs
* is the way I'm handling recoverable errors acceptable and what we were thinking? 
* are we happy with me bringing in `mockall` for unit test? IMO we should be leveraging it thru out our tests, especially tests involving Profile trait components

# TODO:
- [x] Manually test against ACApy issue v2
- [ ] Finish unit tests with mocks
- [ ] More system tests for more anoncreds variety
- [ ] Remove the mock jsonld format handler
- [ ] Remove acapy testing related assets
- [ ] more commenting
- [ ] couple of small in-line TODOs
- [ ] make lint more happy :)

# Future work
some items which i'd like to leave for the future;
* Serde of the state machines
* 'Generic' Holder / Issuers
* 'processing' simplified layer


Patrik-Stas:
> The idea is that in the future aries-vcx could include more formats, or consumers can go ahead and create their own format impls outside of aries-vcx. This way, aries-vcx does not have a hard requirement to keep up with the (potentially) growing list of formats in the registry, as consumers can create their own in the meantime.

I like this extensibility property a lot

gmulhearn:
`HolderV2` with `HyperledgerIndyHolderCredentialIssuanceFormat` works with aca-py for receiving anoncreds via issue-credential-v2; just manually tested it

aries_vcx/src/protocols/issuance_v2/holder/states/credential_received.rs
```diff
@@ -0,0 +1,29 @@
+use messages::msg_fields::protocols::cred_issuance::v2::issue_credential::IssueCredentialV2;
+
+use crate::protocols::issuance_v2::formats::holder::HolderCredentialIssuanceFormat;
+
+pub struct CredentialReceived<T: HolderCredentialIssuanceFormat> {
+    #[allow(dead_code)] // `credential` may become used in future
+    pub(crate) credential: IssueCredentialV2,
+    pub(crate) stored_credential_metadata: T::StoredCredentialMetadata,
```
Patrik-Stas, line 8: Perhaps should be private. We are actually not using `new` constructor now, as it's possible to construct the `struct` manually - which is what we do instead.

aries_vcx/src/protocols/issuance_v2/holder/states/credential_received.rs
```diff
@@ -0,0 +1,29 @@
+use messages::msg_fields::protocols::cred_issuance::v2::issue_credential::IssueCredentialV2;
+
+use crate::protocols::issuance_v2::formats::holder::HolderCredentialIssuanceFormat;
+
+pub struct CredentialReceived<T: HolderCredentialIssuanceFormat> {
+    #[allow(dead_code)] // `credential` may become used in future
+    pub(crate) credential: IssueCredentialV2,
```
Patrik-Stas, line 7: I am leaning toward removing this. You put a hard effort to eliminate redundancy and store minimal set of data inside `stored_credential_metadata`. But we store entire `issue-credential` message, it has privacy/security implications. 
In mobile setting, you would ideally want to store in the wallet not only all crypto data, but also all private data. If you store state machine in this state, you are leaking the private data in possibly un-encrypted storage.

aries_vcx/src/protocols/issuance_v2/holder/states/complete.rs
```diff
@@ -0,0 +1,23 @@
+use std::marker::PhantomData;
+
+use messages::msg_fields::protocols::cred_issuance::v2::ack::AckCredentialV2;
+
+use crate::protocols::issuance_v2::formats::holder::HolderCredentialIssuanceFormat;
+
+pub struct Complete<T: HolderCredentialIssuanceFormat> {
```
Patrik-Stas, line 7: The complete state contains radically different data than `CredentialReceived`, but in practice they are kinda-sorta the same state, obviously this is just marking that we have sent the final ack. Which makes `CredentialReceived` against kind of ephemeral. Which makes me (again) wonder if the 2 states shall not be merged, as typically you would:
- receive `issue-credential` message (being in `OfferPrepared` stated)
- ephemerally transition to `CredentialReceived` state, as most likely right after, you would send `Ack` and transition to `Complete`

But if the 2 states remain separate, I would expect them to hold the same data.

aries_vcx/src/protocols/issuance_v2/holder/states/complete.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub(crate) ack: Option<AckCredentialV2>,
```
Patrik-Stas, line 8: This should not be here. The ack is generated and sent, it's not useful to track how exactly the `Ack` message we have sent looked like.

aries_vcx/src/protocols/issuance_v2/formats/holder/hyperledger_indy.rs
```diff
@@ -0,0 +1,240 @@
...
+    async fn create_request_attachment_content_independent_of_offer(
+        _: &Self::CreateRequestInput,
+    ) -> VcxResult<(Vec<u8>, Self::CreatedRequestMetadata)> {
+        Err(AriesVcxError::from_msg(
+            AriesVcxErrorKind::ActionNotSupported,
+            "Anoncreds cannot create request payload independent of an offer",
+        ))
+    }
+
+    async fn process_and_store_credential(
```
Patrik-Stas, line 202: Assuming we would keep track of the `credential-request` we have sent, we could also validate that the the received credential is matching credential request we have sent. 

Afaik AFJ is doing this; Not a hard requirement, but perhaps nice2have feature down the line.

aries_vcx/src/protocols/issuance_v2/formats/holder/hyperledger_indy.rs
```diff
@@ -0,0 +1,240 @@
...
+        let rev_reg_id = _parse_rev_reg_id_from_credential(&credential_payload)?;
+        let rev_reg_def_json = if let Some(rev_reg_id) = rev_reg_id {
+            let json = ledger.get_rev_reg_def_json(&rev_reg_id).await?;
+            Some(json)
+        } else {
+            None
+        };
+
+        let cred_id = anoncreds
+            .prover_store_credential(
```
Patrik-Stas, line 227: Following comment is out of scope of this PR, non-actionable. The fact that we store credential here seems quite wrong.

1. Anoncreds API should not store anything, If anything, I would expect a `wallet` to be here.
2. Following up previous point, in fact, there should be no persistence here. This function should rather return an artifact, eg. some sort of credential data which shall be stored by consumer.

Our today's paradigm is that the protocol also end up representing "the artifact" is wrong and lead to application code, where credentials are represented by state machines in `Complete` states. 
The same problem we have with Connection protocol, (but afaik not as much with DidExchange), where the finalized state machine end up representing a didcomm channel (DID + Keys)

The problem with this coupling of 
- "output artifact" (be it "DID + Keys" or <credential in some format>) 
- with "state machine"

Will cause headaches whenever there is new version of protocol. In application code, there should be separate collection of artifacts (such as a didcomm channel, a credential of some type) and seaprate collection of state machines - really just helping out with tracking the ongoing conversations. The state machines should not become interface for working with underlying artifacts.

One more illustration of this problem is revocation. In V1 protocol, issuer state machine in states where a credential came to an existence had api `revoke_credential()`. Convenient but not part of aries protocol, just mean of controlling artifact created within an aries protocol.

Patrik-Stas (changes requested):
Left review,  but nevertheless this is awesome job. 👍 

Also note my review is partial, I don't feel like I yet went through it thoroughly enough. Will continue tomorrow.

bobozaur (changes requested):
I'm leaving this as a general comment as it impacts a big part of the code.

The formats thing feels weird to me for a number of reasons. I've went through RFCs and specifications these past couple of days and thought about this quite a bit thinking of maybe proposing some alternative implementation, but I think that might actually bias things now so instead I'd like to propose an open discussion about the architecture of this.

My two cents are:
- Probably the most important thing is that some of these formats are not even mature/used anywhere yet (like the identity foundation one). I don't even know why they included them.
- In the same note, I tend to believe more and more that these formats are tightly connected to the way a credential gets processed. For the `hlindy` formats, it's through `credx`. For the `anoncreds` one, it would be through `anoncreds-rs`. For the `JSON-LD` stuff, it seems W3C talks about some controllers and stuff, which poses a different processing mechanism. The `dif` one has no such thing yet (lol).
- There are multiple formats that a credential can be provided in, which is why these attachments are represented as arrays.

So what this boils down to in my mind is that a particular agent will support one or more formats based on what *implementations* of credential processing it supports. While there might be some wiggle room for a compatibility layer between credential formats, especially between `hlindy` and `anoncreds`, I don't think we should rely on that (as it seems like they do not either).

So, in that vein, an agent that only works with `credx` will then look for the `hlindy` credential format and process that in a well defined data structure from `credx`. One that implements `anoncreds-rs` will do the same, but for the `anoncreds` format. One that supports both could technically choose the first one that matches a supported format *or* have a preferred format that it searches for first.

I believe the current implementation couples these formats with the protocol way more than it should. The formats are strictly used for the attachments, which get ultimately processed through a `Wallet` and `Anoncreds` implementation. The protocol specification does not directly care about the format.

I think that all the formats should be used for are taking an `Attachment`, converting it to the data structure used by the `Anoncreds` implementation, use a `Wallet` maybe to do some stuff with the cred, and move on. Or, conversely, take some expected data structure and ultimately provide an `Attachment`, which is what the protocol and the messages work with.

Right now the format traits are encapsulating the creation, processing, storing of credential (and they would probably contain the retrieval too when the `PresentProofV2` protocol is introduced). That's too much, if you ask me. Instead, I think the format should strictly work with the `Attachment` type (either producing or processing one).

In fact, given that the `Anoncreds` implementation actually "dictates" the format you're looking for, the whole formatting trait might not even be needed. Anyway, I thought about some specifics in terms of implementation too but I'm wondering what everyone is thinking in terms of this and we can discuss approaches afterwards.

@Patrik-Stas @mirgee @gmulhearn ?

gmulhearn-anonyome:
@bobozaur  thanks for spending time reviewing and thinking, always appreciated.

---

> - In the same note, I tend to believe more and more that these formats are tightly connected to the way a credential gets processed. For the `hlindy` formats, it's through `credx`. For the `anoncreds` one, it would be through `anoncreds-rs`. For the `JSON-LD` stuff, it seems W3C talks about some controllers and stuff, which poses a different processing mechanism. The `dif` one has no such thing yet (lol).

I guess that is true for `hlindy` & `anoncreds` being tethered to `indy-sdk`/`credx` & `anoncreds-rs`. But also not 100% true, i believe `credx` consumers would also be able to handle `anoncreds` format if they wanted; they just might need to handle the data with a bit more care (e.g. `master_secret` -> `link_secret`)..  e.g. internally i use anoncreds-rs as a Holder in my stack, but i can handle the `hlindy` format just fine (which is something i want since most big agents haven't moved onto `anoncreds` format yet) (i see you mention this later on, but i think it's worth noting).

But if you're saying that these formats are tied to the library/sdk they are using for processing, i don't think this statement is true for ld-proof VCs. There are lotsss of ld-proof VC libraries out there, recently i was experimenting with a new typescript one called `onyx`, however there are plenty of others. 
JSON-LD has been around for awhile, ld-proof VCs is just an application of that. So there are lots of different ways an agent could implement support for that format.

> So what this boils down to in my mind is that a particular agent will support one or more formats based on what _implementations_ of credential processing it supports. While there might be some wiggle room for a compatibility layer between credential formats, especially between `hlindy` and `anoncreds`, I don't think we should rely on that (as it seems like they do not either).
> So, in that vein, an agent that only works with `credx` will then look for the `hlindy` credential format and process that in a well defined data structure from `credx`. One that implements `anoncreds-rs` will do the same, but for the `anoncreds` format. One that supports both could technically choose the first one that matches a supported format _or_ have a preferred format that it searches for first.

I agree for `hlindy` and `anoncreds`, however i think it's oversimplifying to suggest that the format is somewhat abstract and doesn't really mean anything other than just being an artifact of the underlying implementation you're using (e.g. `indy-sdk`/`credx`, `anoncreds-rs`, or some ld-proof SDK).

There are business reasons why an issuer/holder would choose ld-proofs or anoncreds on a case-by-case basis. There is a bit of divide in decentralized identity between these formats, which comes from the differing pros and cons. W3C VCs are standardized (it's w3c), but arguably aren't fully there with privacy features. Anoncreds are great with privacy, but are not as standardized or recognised as w3c VCs.

So i think the choice often comes down to business reasons. For instance, some companies may be forced via regulation to choose w3c or anoncreds. it's not neccessarily just "use the first format handler that matches". So with that, alot of the time "have a preferred format that it searches for first" might be the way it goes.

> which get ultimately processed through a `Wallet` and `Anoncreds` implementation.

*in the case of anoncreds/hlindy format, yea. Not other formats* (just stating this for this thread record)

> and they would probably contain the retrieval too when the `PresentProofV2` protocol is introduced

yea correct, this is the path i was heading

> In fact, given that the `Anoncreds` implementation actually "dictates" the format you're looking for, the whole formatting trait might not even be needed.

Just want to be careful here again, `Anoncreds` implementations will likely be capable of handling `anoncreds` & `hlindy`; they are almost identical. Just want to make sure we don't lock in the aries_vcx `Anoncreds` trait to "only support support the `hlindy` format"

---
Anyway, i think all of those are just small comments/corrections on your comment, i think overall i agree with your sentiment, but still unsure if their is a reasonable solution.

If i'm understanding correctly, you'd essentially prefer that the attachment processing/creating be done *outside* of the `HolderV2`/`IssuerV2` state machine? In general, i was trying to have it so that the HolderV2 and IssuerV2 flow (when using `hlindy`/`anoncreds` format) feels almost the same as the original `Holder` and `Issuer` flows in aries_vcx. I.e. the existing state machines do all the format processing and creating inside them, abstracted from the user, same with `Prover`/`Verifier`. But you're saying that this might not be the right approach here?

I would like to see what implementation you had in mind, i guess i'm imagining a similar impl, but the format handler/s are separated from the state machine. and the consumer code would bounce between the state machine and their selected format handler. And maybe the state machine would have helpers to advice the consumers on what formats can be used. If this is the direction you are thinking, do you think that may be asking the consumer to do too much? particularly because they will have to be storing/maintaining format data/metadata outside of the state machine (such as the credential request metadata a holder gets after creating the anoncreds request). My current state machine impl (+ the `hlindy` format impls) does all of that for the consumer and holds onto all their protocol data in one place, which i thought was convenient and matched the simplicity of other state machines. Additional, i believe what you're proposing would make the state machines extremely trivial, as they'd pretty much just be responsible for putting attachments into messages, matching thids and marking a state, not that that is a bad thing.

Let's discuss further in today's vcx meeting if you're keen

aries_vcx/src/protocols/issuance_v2/issuer/mod.rs
```diff
@@ -0,0 +1,571 @@
...
+    /// the offer with the holder until an agreement is reached.
+    ///
+    /// An offer is prepared in the format of [IssuerCredentialIssuanceFormat], using the provided
+    /// input data to create it. Additionally, a [CredentialPreviewV2] is attached to give further
+    /// details to the holder about the offer.
+    ///
+    /// In the event of failure, an error is returned which contains the reason for failure
+    /// and the state machine before any transitions. Consumers should decide whether the failure
+    /// is terminal, in which case they should prepare a problem report.
+    pub async fn prepare_offer(
```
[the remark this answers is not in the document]
gmulhearn-anonyome: kk, latest commit converts to this approach, let me know what you think. Would be helpful to me if we could go over the reason _why_ again though..

I originally thought it was bcus we were switching to the mentality of "our state machines should only store the information required to keep transitioning", but that is not necessarily true, because i'm still storing other convenience items in the state machines: e.g. `CredentialReceived` state is also storing `stored_credential_metadata`, which is just a convenience for the state machine consumer so that they can do a `holder.get_stored_credential_metadata()` and see information about the credential they just stored.

If we were to go all the way with the "our state machines should only store the information required to keep transitioning" mentality, then `stored_credential_metadata` would be removed from `CredentialReceived` state. and instead it would be returned from the `HolderV2::receive_credential` API in a tuple like we do with messages.

Same story with the issuer's `CredentialPrepared` state, which stores `credential_metadata`, allowing them to do a `issuer.get_credential_creation_metadata()` and see information such as the cred-rev-id of the created credential. To remove `credential_metadata` from the state, then the `IssuerV2::prepare_credential` API would need to return this metadata AND the credential message:
```
pub async fn prepare_credential(self,...) -> <(IssuerV2<CredPrepared>, IssueCredentialV2Message, T::CreatedCredentialMetadata)>
```

So i guess it'd be good if we made a decision/statement about how far we want to lean into the "our state machines should only store the information required to keep transitioning" mentality

aries_vcx/Cargo.toml
```diff
@@ -62,3 +62,5 @@ android_logger = "0.5"
 wallet_migrator = { path = "../wallet_migrator" }
 async-channel = "1.7.1"
 tokio = { version = "1.20", features = ["rt", "macros", "rt-multi-thread"] }
+mockall = "0.11.4"
+reqwest = "0.11.18" # TODO - DELETE ONLY FOR TEMPORARY TEST!!
```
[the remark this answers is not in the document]
gmulhearn-anonyome: being used in `aries_vcx/tests/test_credential_issuance_v2.rs` for my temporary test against aca-py.

will keep this comment unresolved though, as a reminder to remove the dep

aries_vcx/src/protocols/issuance_v2/formats/holder/hyperledger_indy.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub schema_id: String,
+    pub cred_def_id: String,
+}
+
+pub struct HyperledgerIndyCreateRequestInput<'a, R, A>
+where
+    R: AnoncredsLedgerRead,
+    A: BaseAnonCreds,
+{
+    pub my_pairwise_did: String,
```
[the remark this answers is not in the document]
gmulhearn-anonyome: as discussed privately, we're going with `entropy_did`, since unfortunately lower-level libs (credx) wants this value to be a DID

aries_vcx/src/protocols/issuance_v2/issuer/states/proposal_received.rs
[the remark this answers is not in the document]
gmulhearn-anonyome: hmmm, i'm not sure i agree, or maybe i need to adjust my mental model a bit.

> shall not have states which purely serve for "storing messages".

I think the `ProposalReceived` state is doing a bit more than just storing the message, it's:
* validating the Thread ID before giving `IssuerV2` the chance to reply with an offer
  * this thread ID then needs to be reused when they send/prepare the offer message
* providing a helper method to view the proposal details (`get_proposal_details`). Which is good IMO since getting these proposal details is not trivial (it's buried within the attachments under a specific format)

> If you receive proposal, you look at what's being proposed in the message and apply some custom business logic to evaluate if it is reasonable. If yes, you right away rebuild offer and re-send it. If it's not reasonable, you sent problem report.

I don't think _"right away rebuild offer and re-send"_ is always the case (or maybe i'm misunderstanding). `XYZMessageReceived` states offer the state machine users a chance to stop and think before proceeding. E.g. some "issuer admin web interface" might show an issuer admin a list of all the state machines in the `ProposalReceived` state, and then the admin can visually inspect the proposal (`get_proposal_details`) before choosing to send another offer or to reject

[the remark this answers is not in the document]
gmulhearn-anonyome: I'm not sure how the API would look like for bulletpoint 2 you're describing.

I guess it would be something like 
```
IssuerV2<OfferPrepared>::prepare_offer_in_response_to_proposal(
    self, 
    proposal: ProposalMessage, 
    input_data: &T::CreateOfferInput,
    preview: CredentialPreviewV2,
    replacement_id: Option<String>,
) -> IssuerV2<OfferPrepared>
```

aries_vcx/src/protocols/issuance_v2/issuer/states/complete.rs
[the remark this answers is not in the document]
gmulhearn-anonyome: ah yea agreed on getting the tense consistent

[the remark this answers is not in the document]
gmulhearn-anonyome: as for `xxxx_prepared`, i think i may have been inspired by the `Prover` API. Also from the `guidelines.md`:

> Example: Do not create states such as `PresentationSent` (in context of `presentation-protocol` in role of prover).
Instead, only create state `PresentationPrepared`

Either way, i don't mind, just let me know

aries_vcx/src/protocols/issuance_v2/holder/states/credential_received.rs
gmulhearn-anonyome: kk agreed

aries_vcx/src/protocols/issuance_v2/holder/states/complete.rs
gmulhearn-anonyome: Part of the reason i have them separate, is because i started off implementing issue-credential-v2 as the V2.2 impl (i.e. what is seen on main branch of the RFC: https://github.com/hyperledger/aries-rfcs/blob/main/features/0453-issue-credential-v2/README.md). This is bcus in V2.1+ of the issue-credential-v2 protocol, the `CredentialReceived` state is not necessarily the end, if there are MULTIPLE credential being offered, then the holder can reply to an issue-credential message with ANOTHER request-credential message (i.e. transitioning from `CredentialReceived` -> `RequestPrepared` in our state machine).

So i kept them as seperate to support that transition logic. however since then i reverted to just the plain old 2.0 implementation (no support for multi credential issuance). Now the `Completed` state is essentially just a "Ack optionally prepared" state. I like having them seperated personally.. but we can discuss further if you prefer them merged.

> But if the 2 states remain separate, I would expect them to hold the same data.

i.e. the `Completed` state should have a
```
    stored_credential_metadata: T::StoredCredentialMetadata,
```
field?

aries_vcx/src/protocols/issuance_v2/formats/holder/hyperledger_indy.rs
gmulhearn-anonyome: yea agreed

Patrik-Stas: tracking this in issue https://github.com/hyperledger/aries-vcx/issues/1044

aries_vcx/src/protocols/issuance_v2/formats/holder/hyperledger_indy.rs
Patrik-Stas: - Ultimately this is problem of applications built on top of `aries_vcx`. We should probably have a readme with some list of reccomendations, best practices for building on top of `aries_vcx`. 
- We could also deter implementations relying on protocols in final states by stripping off state machines of some APIs. For example, `Connection` in any state should not provide `send_message` functionality. `IssuerCredential` in final state should not provide `revoke` functionality.

aries_vcx/tests/test_credential_issuance_v2.rs
```diff
@@ -0,0 +1,565 @@
...
+
+        let offer_details =
+            HLIndyHolderFormat::<MockLedger, MockAnoncreds>::extract_offer_details(&offer_msg)
+                .unwrap();
+        assert_eq!(offer_details.schema_id, cred_def.get_schema_id());
+        assert_eq!(offer_details.cred_def_id, cred_def.get_cred_def_id());
+
+        let attachment_format =
+            HLIndyHolderFormat::<MockLedger, MockAnoncreds>::get_request_attachment_format();
+        let (request_attach, request_metadata) =
```
Patrik-Stas, line 156: analogically applicable my previous comment 

> Perhaps HLIndyIssuerFormat::<MockAnoncreds>::get_offer_attachment_format(); could be used internaly by create_offer_attachment_content and return that in the tuple (of 3 elements)?

Patrik-Stas (changes requested):
Leaving what I have so far, but there's probably space for more review. But let's keep the conversation ping-ponging and see where we get

I would also like to propose following to make the reviewing for me and other easier. How about we actually take out all state machines from here, and have that in separate PR. We can focus on the underlying infra here, then retrofit state machines layer. Given my understanding that you don't use state machines, this might be better for you too?

JamesKEbert:
@gmulhearn what should we do with this PR do you think? I suspect that it might have a lot of old code/methodologies and should be written from the ground up, but I haven't looked at the code so that's just a gut thought.

gmulhearn:
@JamesKEbert yeah agreed. I think it's better if it's re-written with an approach that aligns with us, this approach ended up getting a bit convoluted.

PR #988, opened 2023-09-20
Fix lints from aries_vcx_core

bobozaur:
PR that addresses lints from `aries_vcx_core` crate.


aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -508,38 +501,9 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
         };
 
         let revocation_config = match &mut revocation_config_parts {
-            Some((rev_reg_def, rev_reg_def_priv, rev_reg, rev_reg_info, tails_dir)) => {
+            Some((rev_reg_def, rev_reg_def_priv, rev_reg, rev_reg_info)) => {
                 rev_reg_info.curr_id += 1;
 
-                let tails_file_hash = match rev_reg_def {
```
Patrik-Stas, line 514: well, the CI passing, but how come this entire block is removed?

Patrik-Stas: Ah yes, I see the chain of unused variables now. 

This 
```
                       if rev_reg_info.curr_id > rev_reg_def.value.max_cred_num {
                            return Err(AriesVcxCoreError::from_msg(
                                AriesVcxCoreErrorKind::ActionNotSupported,
                                "The revocation registry is full",
                            ));
                        }
```
was still possibly useful ... I suspect if we don't fail here, some similar error will be thrown from somewhere deeper, but it would be good to know for sure what happens if we pass the capacity of the rev registry.

bobozaur: This is an amazing catch. I messed it up big time.

You're right, this is still needed. Added it back. I removed it as I was clearing out the need to use the tails file but this piece of code was doing more than that and it slipped by me.

You're the real MVP, 💯 .

Patrik-Stas (approved):

mirgee (approved):

PR #990, opened 2023-09-21
Issue Credential V2.0 message structures

gmulhearn-anonyome:
NOTE: I've gone with v2.0 (as opposed to v2.1 & v2.2), this is primarily because it seems to be what others (aca-py & AFJ) are targeting and expect. similarly, V2.0 is what AIP2.0 lists as it's requirement: https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0302-aries-interop-profile/README.md#base-requirements

* Rename all existing issue-credential message types to have a V1 suffix
* Put v1 and v2 messages in different structures
* 

Also note, I'm consuming these messages in the Issuer and Holder V2 impl: https://github.com/hyperledger/aries-vcx/pull/987

Also also note, have tested holderv2 against aca-py for issue-credential-v2, which should confirm that the message structures are interoperable with acapy


aries_vcx/src/handlers/util.rs
```diff
@@ -1,7 +1,7 @@
 use messages::{
     msg_fields::protocols::{
         connection::{invitation::Invitation, Connection},
-        cred_issuance::CredentialIssuance,
+        cred_issuance::{v2::CredentialIssuanceV2, CredentialIssuance},
```
bobozaur, line 4: Please gather the V1 stuff in their own directory too. I didn't implicitly do it because we only had one version for each protocol and separating versions was not a problem.

Now, however, it's easier to figure out where to look for a specific version if they are separated in different modules.

messages/src/lib.rs
```diff
@@ -52,6 +55,7 @@ pub enum AriesMessage {
     Connection(Connection),
     Revocation(Revocation),
     CredentialIssuance(CredentialIssuance),
+    CredentialIssuanceV2(CredentialIssuanceV2),
```
bobozaur, line 58: It would be helpful to rename the previous `CredentialIssuance` stuff to V1.

bobozaur (changes requested):
Looks good overall.

The only thing I'd change would be renaming/compartmentalizing stuff related to the V1 protocol version a bit better, since now there's V2 stuff laying around.

bobozaur:
@gmulhearn-anonyome  I'll try to put together some guidelines for the message crate in a README file to maybe clarify how things get pieced together. But if you have any specific questions before then, feel free to ask here.

bobozaur:
@gmulhearn-anonyome  I'll try to put together some guidelines for the message crate in a README file to maybe clarify how things get pieced together. But if you have any specific questions before then, feel free to ask here.

messages/src/lib.rs
bobozaur, line 58: I also realized now that this is not the intended way to deal with this. The `CredentialIssuance` variant should contain the underlying versions in a separate enum.

bobozaur (changes requested):

messages/src/lib.rs
gmulhearn-anonyome: kk i've gone ahead with this change. A note tho; it made the `transit_to_aries_msg` macros not directly work. i.e. 
```
transit_to_aries_msg!(
    OfferCredentialV1Content: OfferCredentialV1Decorators,
    CredentialIssuanceV1
);
```

i had to add the following explicit `From` statements to complete the `From` chain:
```
impl From<CredentialIssuanceV1> for AriesMessage {
    fn from(value: CredentialIssuanceV1) -> Self {
        AriesMessage::CredentialIssuance(CredentialIssuance::V1(value))
    }
}

impl From<CredentialIssuanceV2> for AriesMessage {
    fn from(value: CredentialIssuanceV2) -> Self {
        AriesMessage::CredentialIssuance(CredentialIssuance::V2(value))
    }
}
```

gmulhearn-anonyome:
@bobozaur thanks for the feedback. let me know what you think about this latest revision

messages/src/msg_fields/protocols/cred_issuance/v2/issue_credential.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[builder(default, setter(strip_option))]
+    #[serde(rename = "~timing")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub timing: Option<Timing>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub enum IssueCredentialAttachmentFormatType {
+    #[serde(rename = "aries/ld-proof-vc@v1.0")]
+    AriesLdProofVcDetail1_0,
```
gmulhearn-anonyome, line 45: invalid name, should be `AriesLdProofVc1_0`

messages/src/msg_fields/protocols/cred_issuance/v1/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+        let protocol = Protocol::from(CredentialIssuanceTypeV1_0::parent());
+        let kind = CredentialIssuanceTypeV1_0::from(self);
+        format_args!("{protocol}/{}", kind.as_ref()).serialize(serializer)
+    }
+}
+
+transit_to_aries_msg!(
+    OfferCredentialV1Content: OfferCredentialV1Decorators,
+    CredentialIssuanceV1
+);
```
bobozaur, line 181: This macro has variadic arguments. You can append `CredentialIssuance` at the end to make the conversions chain longer:

```suggestion
transit_to_aries_msg!(
    OfferCredentialV1Content: OfferCredentialV1Decorators,
    CredentialIssuanceV1, CredentialIssuance
);
```

I'll remove the macro at some point when I find the time to work more on the `transitive` crate so it accepts generics. It will become redundant then.

messages/src/msg_fields/protocols/cred_issuance/mod.rs
```diff
@@ -1,218 +1,26 @@
...
-struct CredentialPreviewMsgType;
-
-impl<'a> From<&'a CredentialPreviewMsgType> for CredentialIssuanceTypeV1_0 {
-    fn from(_value: &'a CredentialPreviewMsgType) -> Self {
-        CredentialIssuanceTypeV1_0::CredentialPreview
+impl From<CredentialIssuanceV2> for AriesMessage {
+    fn from(value: CredentialIssuanceV2) -> Self {
+        AriesMessage::CredentialIssuance(CredentialIssuance::V2(value))
     }
 }
```
bobozaur, line 26: If you use the `transit_to_aries_msg` variadic arguments this becomes redundant. You should rarely need to work with intermediate structures like this (perhaps only when matching on an `AriesMessage`).

Other than that, the conversion is really just implemented for concrete messages (`MsgParts<RequestCredentialV2Content, RequestCredentialV2Decorators>`) to `AriesMessage`.

bobozaur (changes requested):
Almost there.

The only nitpick I have is about the conversions that arise from introducing the `CredentialIssuance` wrapper enum (since there are two versions now).

There are all these intermediate enums throughout the crate which are really there simply for composition purposes. You should rarely have to work with the `CredentialIssuanceV1` enum directly. 

The actual messages all have implementations of `From` (through the `transit_to_aries_msg` macro) to convert them to an `AriesMessage` directly specifically to avoid the need to deal with all the nested enums.

messages/src/msg_fields/protocols/cred_issuance/v1/mod.rs
gmulhearn-anonyome: ahhhhh thank you, this makes sense. nice macro!

messages/src/msg_fields/protocols/cred_issuance/mod.rs
gmulhearn-anonyome: fixed 👍

bobozaur (approved):
LGTM. If CI passes then it's all good on my end.

Patrik-Stas (approved):

PR #992, opened 2023-09-22
trait(BaseWallet): change return type of unpack_message,

nain-F49FF806:
to defined struct (was initially opaque vec<u8>)


Patrik-Stas:
Hi @nain-F49FF806 the ` test-integration-aries-vcx-vdrproxy` was failing intermittently was should be fixed now. You just need to fix the formatting which is still off somewhere.

nain-F49FF806:
@Patrik-Stas Fixed!

It was about some additional fixes from nightly version of rustfmt.

aries_vcx/src/common/signing.rs
Patrik-Stas, line 1: Not going to hold up this PR on this, but if you'd like contribute another small PR - we should get rid of function  `unpack_message_to_string` 

If you follow its usage, you will find that it's only being utilized by `wrappers/vcx-napi-rs/src/api/wallet.rs`. Instead of having this special use-case function in aries-vcx, let's just do the struct -> string conversion in the `wrappers/vcx-napi-rs` at the place it's needed

bobozaur (approved):

PR #998, opened 2023-09-26
Workspace reorganization

bobozaur:
This PR aims to address the ever growing number of crates we have in the workspace so that we make it easier for newcomers as well as us to navigate through the directory structure.

This top comment will be updated with the directory structure after it's debated enough and agreed upon. Combining existent crates is represented by `+` while creating a new directory where existent crates will be moved is represented by a nested bullet point list.

Directory structure:
- .github
- docs
- agents
    - rust
    - node
- aries_vcx
- aries_vcx_core
- did_doc
    - did_doc_core
    - did_doc_sov
- did
    - did_key
    - did_parser
    - did_peer
    - did_resolver
        - did_resolver_core
        - did_resolver_sov 
        - did_resolver_web
        - did_resolver_registry
- public_key
- aries_messages
    - messages
    - messages_macros
- tools
    - simple_message_relay
    - wallet_migrator
- legacy
    - did_doc_legacy
    - libvdrtools
    - agency_client
- shared_vcx
- wrappers
    - node
        - libvcx_core
        - node
        - vcx-napi-rs
    - uniffi-aries-vcx

**NOTE:** The `ci` directory gets moved under `.github` in #1001 .

Questions:
- [x] Remove `did_resolver_registry`? It's unused, small and not really something I personally like due to being an unnecessary batteries included abstraction over something consumers should build from scratch.
**ANSWER**: used in #928 .
- [x] Do we still need `agency_client`? Could we remove it or should it just be moved to the `legacy` directory?
**ANSWER**: `agency_client` is still needed so we'll place it in the `legacy` directory.


bobozaur:
> 1. How about we put all crates under root `/crates` directory?

Idk about this. I personally don't like it. It would make sense if there's a lot of other stuff not related to the Rust code, but there isn't that much really.

>     2. > "Do we still need agency_client? "
>        
>        
>        Unfortunately, along with mediated connection, yes. It's needed by NodeJS tests. In `aries-vcx`, mediated connection module however is not used by other `aries-vcx` modules, so it could be easily feature flagged, and opt-into from `libvcx_core` only.

Fair enough.

>     3. > "Remove did_resolver_registry?"
>        
>        
>        Not sure, probably question for @mirgee

To me it just seems like YAGNI.

>     4. I am wondering if `ci` can be squashed behind `.github` - not sure if github allows that

It might, and it's actually a good idea.

>     5. You are missing [diddoc_legacy](https://github.com/hyperledger/aries-vcx/tree/main/diddoc_legacy) in the described repo structure

Yaeh, it was meant to go under the `legacy` dir. Will fix.

>     6. I imagine `messages_macros` wouldn't really be used by other crates other than `messages`. How about merging `messages` + `messages_macros`? Just throwing the idea out there.

No can do. Procedural macros need to reside in their own crate.

>     7. `libvcx_core` is curretly as much/as little legacy as `node` and `vcx-napi-rs` components which are built on it. Perhaps they should reside together. I wouldn't mind to but these nodejs wrapper under legacy directory. While we care about them for now, we don't plan to in a long run, and we wouldn't want to encourage usage of it. Also the current `agents/node/*` would fall into the same category.
>        
>        * then having nodejs wrappers out of our way from what's in your proposal currently named `wrappers`, we could "rebrand" that a bit and rather call that "agents" - it could contain:
>          
>          * uniffi sample application
>          * mediator agent (once's it's chipped into this repo)
>          * aries-agent-rust (although not executable, it's a basis for AATH backchannel agent)

Okay, I'll update the directory structure.

mirgee:
I would also prefer putting all crates into one directory. `did_resolver_registry` is used in #928 .

bobozaur:
> I would also prefer putting all (or most) crates into one directory and agree with keeping `libvcx_core`, `vcx-napi-rs` and the node wrapper together (not so sure about the node-based agent). `did_resolver_registry` is used in #928 .

Alright so since you both want some upper dir I added it. What's up with the `did_resolver_registry` though? I know this might not be the proper place to ask but shouldn't consumers be able to pass whatever resolver they want (implementor of `DidResolvable`)?

The `did_resolver_registry` is just an opinionated implementation of a resolver that delegates the resolution to underlying, method dedicated resolvers.

mirgee:
@bobozaur Why not put it under `crates/did` or just `crates`.

bobozaur:
> @bobozaur Why not put it under `crates/did` or just `crates`.

Okay, fair.

bobozaur:
As discussed in the community call, the `crates` root dir was removed.

Patrik-Stas:
I would like to propose some further changes to the org

- .github
- docs
- aries 
  - shared_vcx
  - legacy
    - did_doc_legacy
    - libvdrtools
    - agency_client
  - agents
    - rust
    - node
  - aries_vcx
  - aries_messages
    - messages
    - messages_macros
  - wrappers
    - node
        - libvcx_core
        - node
        - vcx-napi-rs
    - uniffi-aries-vcx
- anoncreds
   - vcx_wallet
   - vcx_ledger
   - vcx_anoncreds
- did_document
    - did_doc_core
    - did_doc_sov
- did
    - public_key
    - did_parser
    - did_methods
        - did_resolver_core
        - did_key
        - did_peer
        - did_sov 
        - did_web
        - did_registry
- tools
    - simple_message_relay
    - wallet_migrator

- There's a few "final state" assumptions made here, for example I am assuming that we split `aries_vcx_core` to 3 crates `vcx_wallet`, `vcx_ledger`, `vcx_anoncreds`
- I've slightly renamed some of the did methods crates, eg `did_resolver_sov` to just `did_sov`
- Not sure If I haven't misplaced `public_key` or `did_key`

But what this mainly does is clearly separates aries components and other components which are reusable and aries-agnostic.

Patrik-Stas:
Moving this discussion to github issue 
https://github.com/hyperledger/aries-vcx/pull/998

PR #1002, opened 2023-09-27
Use standard UnpackMessageOutput struct until finally needing to transform.

nain-F49FF806:
As [suggested in](https://github.com/hyperledger/aries-vcx/pull/992#discussion_r1337239159) review of #992


Patrik-Stas (approved):
Awesome, thank you!

PR #1006, opened 2023-10-03
Do not delete target wallet, do not fail migration on item-error

Patrik-Stas:
This is set of changes to make the migration.
- More graceful, one migration error should not halt migration of the rest. It's up to migrating user to evaluate whether the migration result is satisfactory.
- Idempotent - if migration fails for example due an IO error midway, it should be possible to finish it on 2nd try.

Changes:
- If migration fails, do not delete the target wallet.
- Disable adding records to vdrtool cache when running migration.
- Do not fail migration if the record has unexpected format. Just skip it and log the record. 
- If migration of the record itself fail, skip it and log.
- If adding item to target wallet fails due duplication error, skip it. For idempotency it would be ideal to overwrite it, but that would need digging deeper. For now, skipping these records still gives us idempotency under assumption item migration process was not changed between 2 migrations attempts.
- If adding item fails for other reason, fail the migration (likely IO error).

### DIFF
--- a/libvcx_core/src/api_vcx/api_global/wallet.rs
+++ b/libvcx_core/src/api_vcx/api_global/wallet.rs
@@ -15,7 +15,7 @@ use aries_vcx::{
             base_wallet::BaseWallet,
             indy::{
                 internal::{close_search_wallet, fetch_next_records_wallet, open_search_wallet},
-                wallet::{close_wallet, create_and_open_wallet, delete_wallet, import},
+                wallet::{close_wallet, create_and_open_wallet, import},
                 IndySdkWallet, IssuerConfig, RestoreWalletConfigs, WalletConfig,
             },
             structs_io::UnpackMessageOutput,
@@ -282,7 +282,12 @@ pub async fn wallet_import(config: &RestoreWalletConfigs) -> LibvcxResult<()> {
 
 pub async fn wallet_migrate(wallet_config: &WalletConfig) -> LibvcxResult<()> {
     let src_wallet_handle = get_main_wallet_handle()?;
+    info!(
+        "Creating or opening target wallet {}",
+        &wallet_config.wallet_name
+    );
     let dest_wallet_handle = create_and_open_wallet(wallet_config).await?;
+    info!("Target wallet is ready.");
 
     let migration_res = wallet_migrator::migrate_wallet(
         src_wallet_handle,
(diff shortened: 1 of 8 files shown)


libvcx_core/src/api_vcx/api_global/wallet.rs
```diff
@@ -291,18 +296,11 @@ pub async fn wallet_migrate(wallet_config: &WalletConfig) -> LibvcxResult<()> {
...
-        delete_wallet(wallet_config).await.ok();
-        Err(LibvcxError::from_msg(
-            LibvcxErrorKind::WalletMigrationFailed,
-            e,
-        ))
-    } else {
-        setup_wallet(dest_wallet_handle)?;
-        close_wallet(src_wallet_handle).await?;
-        Ok(())
-    }
```
bobozaur, line 305: Why not delete the destination wallet on failure?

libvdrtools/indy-wallet/src/wallet.rs
```diff
@@ -188,7 +189,9 @@ impl Wallet {
         );
 
         self.storage.add(&etype, &ename, &evalue, &etags).await?;
-        self.cache.add(type_, &etype, &ename, &evalue, &etags);
+        if cache_record {
+            self.cache.add(type_, &etype, &ename, &evalue, &etags);
+        }
```
bobozaur, line 194: What is this change supposed to do? I see the flag is set to `true` in both instances where this is used.

libvcx_core/src/api_vcx/api_global/wallet.rs
```diff
@@ -282,7 +282,12 @@ pub async fn wallet_import(config: &RestoreWalletConfigs) -> LibvcxResult<()> {
 
 pub async fn wallet_migrate(wallet_config: &WalletConfig) -> LibvcxResult<()> {
     let src_wallet_handle = get_main_wallet_handle()?;
+    info!(
+        "Creating or opening target wallet {}",
+        &wallet_config.wallet_name
+    );
```
bobozaur, line 288: I think it will *always* attempt to create it.

bobozaur (changes requested):
While I understand that this helped in troubleshooting some issues (practically unrelated to the migration code) I think most changes (maybe apart from logging) should be reverted.

The fact that the migrator cleans up after itself on failure might not be a big deal for us, but it probably is for consumers that will use it.

Likewise, I don't think the migration should just continue on record errors. This is important data and I personally believe that **anything** that does not go well should be manually investigated and dealt with and the migration re-attempted afterwards, essentially succeeding only if everything is in order.

libvdrtools/indy-wallet/src/wallet.rs
Patrik-Stas: In migration, the flag is set to `false`.

Patrik-Stas:
> While I understand that this helped in troubleshooting some issues (practically unrelated to the migration code) I think most changes (maybe apart from logging) should be reverted.

> The fact that the migrator cleans up after itself on failure might not be a big deal for us, but it probably is for consumers that will use it.

If migration fails, you might want to inspect the wallet

> Likewise, I don't think the migration should just continue on record errors. This is important data and I personally believe that anything that does not go well should be manually investigated and dealt with and the migration re-attempted afterwards, essentially succeeding only if everything is in order.

If you fail on first error, you don't even get to know all the problem you have, only the first one. The changed approach in this PR provide flexibility to migrators. You can but don't have to be strict about failures, it's your call based on `MigrationResult`. In case of `libvcx`, as the only users, I chose the approach of being relaxed about it and decide on upper level of abstractions.

bobozaur (approved):
I got mixed up in the files being modified and forgot the design of the migration. Given that this affects `libvcx` stuff, I think the changes are then warranted. Thanks for clarifying!

PR #1007, opened 2023-10-03
Use a sequence of bytes (representing AriesMessage) as input for EncryptionEnvelope::create

nain-F49FF806:
This makes it more general, allowing use of the utility methods without worrying about the exact structure of `AriesMessage` passed in. In doing so, allows for also wrapping message types not yet recognized in aries_vcx.

### DIFF
--- a/aries_vcx/src/protocols/connection/mod.rs
+++ b/aries_vcx/src/protocols/connection/mod.rs
@@ -100,7 +100,13 @@ where
         message: &AriesMessage,
     ) -> VcxResult<EncryptionEnvelope> {
         let sender_verkey = &self.pairwise_info().pw_vk;
-        EncryptionEnvelope::create(wallet, message, Some(sender_verkey), self.their_did_doc()).await
+        EncryptionEnvelope::create(
+            wallet,
+            json!(message).to_string().as_bytes(),
+            Some(sender_verkey),
+            self.their_did_doc(),
+        )
+        .await
     }
 
     pub fn remote_did(&self) -> &str {
--- a/aries_vcx/src/utils/encryption_envelope.rs
+++ b/aries_vcx/src/utils/encryption_envelope.rs
@@ -10,7 +10,6 @@ use uuid::Uuid;
 
 use crate::{errors::error::prelude::*, global::settings, utils::constants};
 
-type AriesMessageBytes = [u8];
 #[derive(Debug)]
 pub struct EncryptionEnvelope(pub Vec<u8>);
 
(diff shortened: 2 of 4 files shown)


Patrik-Stas (approved):

aries_vcx/src/utils/encryption_envelope.rs
```diff
@@ -10,13 +10,16 @@ use uuid::Uuid;
 
 use crate::{errors::error::prelude::*, global::settings, utils::constants};
 
+type AriesMessageBytes = [u8];
```
gmulhearn, line 13: unnecessary abstraction IMO. i think as a developer i'd prefer to just see `&[u8]` as the type, rather than `&AriesMessageBytes`

bobozaur: Agreed.

aries_vcx/src/protocols/connection/mod.rs
```diff
@@ -100,7 +100,13 @@ where
         message: &AriesMessage,
     ) -> VcxResult<EncryptionEnvelope> {
         let sender_verkey = &self.pairwise_info().pw_vk;
-        EncryptionEnvelope::create(wallet, message, Some(sender_verkey), self.their_did_doc()).await
+        EncryptionEnvelope::create(
+            wallet,
+            json!(message).to_string().as_bytes(),
```
gmulhearn, line 105: does the `json!` macro have potential for panic here?

according to the docs, it will panic if the `Serialize` impl fails.

i wonder if it'd just be better to use `serde_json::to_vec?` instead of this macro wish might panic.

gmulhearn: same as below

gmulhearn (changes requested):
just left some comments for discussion before merge

aries_vcx/src/protocols/connection/mod.rs
bobozaur: I personally think this can be disregarded, *for now*. This is not some problem introduced by this PR, but was there for a long time. This whole encryption thing needs a rework, IMO.

nain-F49FF806: Since the code was initially passing in `AriesMessage` struct (for which Serialize is implemented locally) I think the possibility of panics in these instances is remote. 

Still, I could add the `serde_json::to_vec?`, if that's better, no problem.

gmulhearn: Oh sorry, I see now that it was already doing this potentially panicky behaviour. No need to fix right now

aries_vcx/src/utils/encryption_envelope.rs
nain-F49FF806: I can change that. I added it initially for explicitly documenting what kind of bytes are expected. (Since we wouldn't want someone to pass in byte slices that are not an AriesMessage). 

But since I have also added documentation comments to the function mentioning the same, I suppose this is redundant.

nain-F49FF806:
@gmulhearn I have made the suggested change. 
And left the json macro bit as is, per discussion above.

gmulhearn (approved):

Patrik-Stas (approved):

bobozaur (approved):

mirgee (approved):

PR #1013, opened 2023-10-06
Purge vdrtools

bobozaur:
Attempts to remove all unnecessary code from `libvdrtools`, keeping only the wallet and the anoncreds data structures definitions (for the `wallet_migrator` crate to use).

Additionally, removed lint exemption and fixed lints in the crate while also removing the `migration` and `vdrtools/vdrtools_anoncreds` feature flags from `aries_vcx` and `aries_vcx_core`.

Lastly, dependencies have been updated and tweaked. Of particular notice are `sqlx` which was bumped to the latest version (`0.7.1`) and `zeroize` which was bumped to the latest version (`1.6.0`).


Patrik-Stas (approved):

Patrik-Stas (changes requested):
Having further look, I think there's more to be cleaned up

- I think we should be able to drop off some cargo dependencies in vdrtools directories
- there seems to be more stuff to delete in `src/domain/anoncreds` 
- I'd suggest to delete `vdrtool2credx` migration
   - this would get our workspace ursa-free, which if I remember right, was major source of pain with askar addtion and sqlx update
- I believe `signature_serializer.rs` can be dropped and also `Hash` in `indy-utils`. Perhaps there's more stuff to delete in `indy-utils`. Openssl dependency can be removed.
- `mac.build.sh` can be deleted

bobozaur:
>     * I think we should be able to drop off some cargo dependencies in vdrtools directories
Aaah right, I forgot about the dependencies! I'll work on that.

>     * I'd suggest to delete `vdrtool2credx` migration
@Patrik-Stas Do you mean to delete the `wallet_migrator` crate?

bobozaur:
@Patrik-Stas I removed some more code and updated/tweaked dependencies.

For the other things:
- I don't think deleting the migration code is something we want. Is it? Down the line I see how we might want to incorporate the data structures from `libvdrtools` in there for posterity, but not really remove it.
- There's definitely code to delete from `src/domain/anoncreds` but it really revolves around these same data structures used by the `wallet_migrator` and I wouldn't really start hunting down unused methods.

Patrik-Stas:
> Do you mean to delete the wallet_migrator crate?
well, not necessarily the entire crate as it could be used for credx2anoncreds-rs or something similar, but given we have released `0.59.1` which provide migration path from vdrtools to credx, and we no longer support vdrtools, I don't see reason why to keep around the `vdrtool2credx` migration in particular.

bobozaur:
> > Do you mean to delete the wallet_migrator crate?
> 
> well, not necessarily the entire crate as it could be used for credx2anoncreds-rs or something similar, but given we have released `0.59.1` which provide migration path from vdrtools to credx, and we no longer support vdrtools, I don't see reason why to keep around the `vdrtool2credx` migration in particular, especially if it brings in all the ursa dependencies, that's quite heavy cost to pay.

As discussed, we'll keep the data types around for a while longer just for the `wallet_migrator` crate. Perhaps we'll be able to move the data structures to `wallet_migrator` afterwards (or at least some high-level of them, without depending on `ursa`) and thus get rid of more things from `libvdrtools`.

aries_vcx_core/Cargo.toml
```diff
@@ -5,36 +5,34 @@ edition = "2021"
...
+vdr_proxy_ledger = ["credx", "dep:indy-vdr-proxy-client"]
 # Feature flag to allow legacy proof verification
 legacy_proof = []
 
 [dependencies]
 agency_client = { path = "../agency_client" }
-indy-vdr = { git = "https://github.com/hyperledger/indy-vdr.git", rev = "879e29e", default-features = false, features = ["log"] }
-indy-credx = { git = "https://github.com/hyperledger/indy-shared-rs", tag = "v1.0.1", optional = true }
+indy-vdr = { git = "https://github.com/hyperledger/indy-vdr.git", rev = "c143268", default-features = false, features = ["log"] }
+indy-credx = { git = "https://github.com/hyperledger/indy-shared-rs", tag = "v1.1.0", optional = true }
```
Patrik-Stas, line 18: @gmulhearn just a heads up

There was bunch of. changes in `indy-shared-rs` and `indy-credx` dependency tree, perhaps this can solve your issues with askar too

indy-shared-rs cargo.toml:
image

bobozaur: Yeah I updated those too because I figured it might help. The update was prevented before by the `zeroize` crate.

Patrik-Stas (dismissed):

mirgee (dismissed):

Patrik-Stas:
@gmulhearn if all looks good, feel free to merge

aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -56,14 +60,144 @@ pub struct RevocationRegistryInfo {
     pub used_ids: HashSet<u32>,
 }
 
+/// Adapter used so that credx does not depend strictly on the vdrtools-wallet
+/// Will get removed when the wallet and anoncreds interfaces are de-coupled.
+#[derive(Debug)]
+struct WalletAdapter(Arc<dyn BaseWallet>);
```
gmulhearn, line 66: im not sure i understand. i thought credx anoncreds would already be decoupled via `Arc<dyn BaseWallet>`

bobozaur: The long-term plan is to remove the interdependencies between components so that, let's say, some `Anoncreds` implementation does not contain a `Wallet` and vice-versa. 

I temporarily added this so that the `BaseWallet` implementation is close to the actual type and can get removed together. The alternative was to keep the `Arc<dyn BaseWallet>` implementation of the `BaseWallet` trait, but that's too generic and easily forgettable.

Patrik-Stas (dismissed):

aries_vcx/src/utils/mod.rs
```diff
@@ -8,7 +8,6 @@ use crate::{
 };
 
 #[macro_use]
-#[cfg(feature = "vdrtools")]
```
gmulhearn-anonyome, line 11: This causes aries_vcx to not compile with `cargo build --no-default-features`:

```
error[E0433]: failed to resolve: could not find `indy` in `wallet`
  --> aries_vcx/src/utils/devsetup.rs:12:13
   |
12 |     wallet::indy::{
   |             ^^^^ could not find `indy` in `wallet`

error[E0432]: unresolved imports `aries_vcx_core::wallet::indy`, `aries_vcx_core::WalletHandle`
  --> aries_vcx/src/utils/devsetup.rs:12:13
   |
12 |     wallet::indy::{
   |             ^^^^ could not find `indy` in `wallet`
...
17 |     WalletHandle,
   |     ^^^^^^^^^^^^ no `WalletHandle` in the root

Some errors have detailed explanations: E0432, E0433.
```

I'm not sure when it was removed, but there used to be a CI step for `cargo check`ing many variants of the aries_vcx. It would pick up on stuff like this; are we able to restore that CI step please?

gmulhearn-anonyome (changes requested):

aries_vcx/src/utils/mod.rs
gmulhearn-anonyome: Also a general question, i wonder if `devsetup` should be hidden behind `#[test]` too? it seems to just be test helpers

aries_vcx_core/Cargo.toml
```diff
@@ -5,36 +5,34 @@ edition = "2021"
...
 [features]
 ########################## DEP FLAGS ################################
-# Feature flag to include the libvdrtools dependency
-vdrtools_anoncreds = ["dep:libvdrtools", "dep:indy-api-types"]
 vdrtools_wallet = ["dep:libvdrtools", "dep:indy-api-types"]
 # Feature flag to include the 'modular library' dependencies (vdrtools alternatives; indy-vdr, indy-credx)
-modular_libs = ["dep:indy-credx"]
-vdr_proxy_ledger = ["modular_libs", "dep:indy-vdr-proxy-client"]
+credx = ["dep:indy-credx"]
+vdr_proxy_ledger = ["credx", "dep:indy-vdr-proxy-client"]
```
gmulhearn-anonyome, line 11: general question that doesn't need to be answered here: does `vdr_proxy_ledger` (i.e. where `VdrProxySubmitter` is the core component) **actually** need the `credx` feature?

You guys know this part of the code best, but it seems like the only reason `vdr_proxy_ledger` needs `credx` is because `vdr_proxy_ledger` is being used to hide the `VdrProxyProfile` implementation, which depends on `IndyCredxAnonCreds` (and `IndySdkWallet` too actually (`vdrtools_wallet`?)). I wonder if logically `VdrProxyProfile` should be hidden by `credx` AND `vdr_proxy_ledger` rather than making `vdr_proxy_ledger` always depend on `credx`

gmulhearn-anonyome: I guess it's a matter of perspective about whether the primary "feature" of `vdr_proxy_ledger` is the `VdrProxySubmitter` AND `VdrProxyProfile`, or if it's just the `VdrProxySubmitter`

did_resolver_sov/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 
 [dependencies]
 did_resolver = { path = "../did_resolver" }
-aries_vcx_core = { path = "../aries_vcx_core", features = ["modular_libs"] }
+aries_vcx_core = { path = "../aries_vcx_core", features = ["credx"] }
```
gmulhearn-anonyome, line 8: nit; resolve if you want: i think this could be `default-features = false` instead of `credx` dependency

gmulhearn-anonyome: nvm i misread this as `aries-vcx`

gmulhearn-anonyome: or just no features, since `aries_vcx_core` has no default features

aries_vcx/src/utils/mod.rs
bobozaur: I replaced the `cargo check` with some `clippy` jobs, but apparently skipped this one. My bad. Will add a job for this. The funny thing is that running `cargo clippy` on the entire workspace does not catch this - you have to run it on `aries_vcx` alone.

Regarding `devsetup`, this does not belong at all in `aries_vcx` as yeah, it's only used in tests. I'm looking into how to accommodate this right now as moving test utilities requires also moving tests and it's a bigger effort.

aries_vcx_core/Cargo.toml
bobozaur: Idk. @mirgee worked on this. Any insights? I just replaced the underlying implementation being used in the `VdrProxyProfile` since we're deprecating `libvdrtools`.

did_resolver_sov/Cargo.toml
bobozaur: On it.

aries_vcx_core/Cargo.toml
gmulhearn-anonyome: ye no worries, just a passing comment

did_resolver_sov/Cargo.toml
bobozaur: Removed features.

aries_vcx/src/utils/mod.rs
bobozaur: Okay so putting `devsetup` behind the `vdrtools_wallet` feature seems to solve issues. I plan on taking this soon and decouple test code/utils from production `aries-vcx` as it's only becoming more and more annoying to deal with. Hopefully CI passes now (also added a job for plain `aries-vcx` with no features)

bobozaur: In all honesty, probably the way to go would be for `aries-vcx` to have no default features (in which case the job added for this wouldn't be required as this would get caught by the workspace-wide clippy CI job). 

But with the way things are currently I am honestly plain afraid of removing the default features as a lot of tests will start going haywire. So this is perhaps something to consider after the separation between test code and prod code.

gmulhearn (approved):
nice, good stuff!

Patrik-Stas (approved):

PR #1014, opened 2023-10-11
Add uniffi holder API

swaptr:
This PR adds support for uniffi holder API.


uniffi_aries_vcx/core/src/handlers/holder.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    pub fn get_msg_credential_request(&self) -> VcxUniFFIResult<String> {
+        let handler = self.handler.lock()?;
+
+        Ok(serde_json::to_string(
+            &handler.clone().get_msg_credential_request()?,
+        )?)
+    }
+
+    pub fn decline_offer(&self) -> VcxUniFFIResult<String> {
```
gmulhearn, line 105: can this take in an optional string that gets passed to the inner decline_offer? i think this string is for `comment`s

uniffi_aries_vcx/core/src/handlers/holder.rs
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    pub fn decline_offer(&self, comment: Option<String>) -> VcxUniFFIResult<String> {
+        let handler = self.handler.lock()?;
+
+        Ok(serde_json::to_string(
+            &handler
+                .clone()
+                .decline_offer(Some(&comment.unwrap_or("".to_owned())))?,
+        )?)
```
gmulhearn-anonyome, line 112: hmmm, whoops, i only just noticed this: i believe `decline_offer` is mutating the handler. so cloning it would result in that mutation not being registered.

also instead of `Some(&comment.unwrap_or("".to_owned())))` perhaps just `comment.as_deref()`

gmulhearn (approved):
nice, lgtm

PR #1017, opened 2023-10-18
Present Proof V2.0 message structures

gmulhearn-anonyome:
Related, almost identical to: https://github.com/hyperledger/aries-vcx/pull/990

TODO


Patrik-Stas (approved):

messages/src/decorators/attachment.rs
```diff
@@ -58,6 +59,22 @@ pub enum AttachmentType {
...
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, TypedBuilder)]
+#[serde(rename_all = "snake_case")]
+pub struct OptionalIdAttachmentFormatSpecifier<F> {
+    #[builder(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub attach_id: Option<String>,
+    pub format: MaybeKnown<F>,
+}
+
```
bobozaur, line 77: I assume these were placed here because they're shared across the V2 protocols, but IMO they don't really belong next to the decorator as they're protocol specific. 

Maybe it's more suitable to create some sort of `common` module in the `protocols` one to share common stuff across protocols?

bobozaur (approved):
LGTM. Feel free to ignore the comment posted if it's not important for you.

messages/src/decorators/attachment.rs
gmulhearn-anonyome: ah yep. agreed, have addressed this in my latest commit. If you're happy with the changes, and the CI is still green, pls merge 👌

bobozaur (approved):

PR #1018, opened 2023-10-18
Add support for pickup protocol messages

nain-F49FF806:
Ref: https://github.com/hyperledger/aries-rfcs/blob/main/features/0685-pickup-v2/README.md


messages/src/msg_fields/protocols/pickup/delivery.rs
```diff
@@ -0,0 +1,31 @@
...
+    #[serde(rename = "@id")]
+    pub id: String,
+    pub data: DeliveryAttachData,
+}
+
+#[serde_as]
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq, TypedBuilder)]
+pub struct DeliveryAttachData {
+    #[serde_as(as = "Base64")]
+    pub base64: Vec<u8>,
```
gmulhearn, line 30: did you consider reusing the `Attachment` type already within the `messages` crate?

gmulhearn:
looking nice. is it in scope to add some tests? typically each message type will have 2 unit tests in the same file: 1 testing serialization/deserialization with minimal content/decorators, and 1 testing with maximum content/decorators. and also some tests for checking that the message type strings are determined correctly. Check the other messages in the `messages` crate for examples

messages/src/msg_fields/protocols/pickup/delivery.rs
nain-F49FF806: I wasn't aware of `Attachment` type. Thanks for bringing it to my notice!

nain-F49FF806: I have now taken a look at the Attachment type. From reading the pickup-v2 RFC, it is unclear to me if delivery messages contain general attachments or only those with base64 formatted data. I have [sought clarification](https://github.com/hyperledger/aries-rfcs/issues/797).

Another difference is the necessity of "@id" field in pickup (for later marking as received), while it is optional in general attachment. 

For now, I'd like to keep it as a separate structure, but I shall try to align it with the already defined type. Later they can be merged if appropriate. 

Does that sound fine?

gmulhearn: Ahh, thanks for looking into that. In that case, I'm happy with it how it is, so long as the other reviewers are fine with it too 👍

messages/src/msg_fields/protocols/pickup/decorators.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub transport: Option<Transport>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq, TypedBuilder)]
+pub struct Transport {
+    pub return_route: ReturnRoute,
+    #[builder(default, setter(strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub return_route_thread: Option<Thread>,
+}
```
bobozaur, line 24: If this is a decorator, it should be placed in the `decorators` module in the crate.

messages/src/msg_fields/protocols/pickup/decorators.rs
```diff
@@ -0,0 +1,59 @@
...
+pub struct PickupDecoratorsCommon {
+    #[builder(default, setter(strip_option))]
+    #[serde(rename = "~thread")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub thread: Option<Thread>,
+    #[builder(default, setter(strip_option))]
+    #[serde(rename = "~transport")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub transport: Option<Transport>,
+}
```
bobozaur, line 16: The general approach is to have a `MyMessageDecorators` type for each message. While it duplicates the data structures sometimes, it's easier to reason with and modify if needed. 

Additionally, the rest of the crate is like that, and I think consistency is also valuable enough to warrant having individual decorators types for each message.

messages/src/msg_types/protocols/pickup.rs
```diff
@@ -0,0 +1,32 @@
...
+#[msg_type(protocol = "messagepickup")]
+pub enum PickupType {
+    V2(PickupTypeV2),
+}
+
+#[derive(Copy, Clone, Debug, From, TryInto, PartialEq, Transitive, MessageType)]
+#[transitive(into(PickupType, Protocol))]
+#[msg_type(major = 2)]
+pub enum PickupTypeV2 {
+    #[msg_type(minor = 0, roles = "")]
```
bobozaur, line 19: I know I was the one to say to set this to an empty string as there's no use in a dynamic discovery of protocols/roles an agent supports in terms of mediation, but adding a comment might be nice and it might clarify why no roles are added.

No strong opinion though, but I'd definitely do that.

bobozaur (changes requested):

messages/src/msg_types/protocols/pickup.rs
nain-F49FF806: Perhaps it'd be suitable to have a mediator role there?

Mediator agents would perhaps advertise support for pickup protocol.

messages/src/msg_fields/protocols/pickup/delivery.rs
nain-F49FF806: Done.

messages/src/msg_types/protocols/pickup.rs
bobozaur: Yeah, fair.

nain-F49FF806: Done

bobozaur (approved):

Patrik-Stas (approved):

PR #1022, opened 2023-10-22
Add backtrace capturing, prototype a test err handling

Patrik-Stas:
- Removed `run(fn)` test case wraps as they turned out to be no longer necessary. Deleted related dead code.
- Add feature `backtrace_errors`; when enabled, `backtrace` dependency is included and error thrown from aries-vcx contain capture backtrace information
- Tweak `Display` impl for `AriesVcxError`
- Add `Debug` impl for `AriesVcxError` - this is useful when err is thrown from a test run by cargo test runner
- Modified all integration tests:
  - They return `-> Result<(), Box<dyn Error>>` instead of `-> ()`.
  - Therefore we can write test code without being coerced into calling `unwrap()`
  - The only good thing about `unwrap()` panics was that that they included stacktrace, which is not default with returning an error - but with additions mentioned above, we are not loosing this capability

If agreed, we can continue adopting:
- same panic-less approach in the rest of the test code
- backtrace capturing for other sub-crates as well 

### After rebase
~~I've temporarily added `run2` but I want to question 2 things:~~
- ~~What's the purpose of macro `run_setup_test(...)` as opposed to just calling `build_setup_profile().await.run(...)`~~
- ~~I believe that whole run thing `pub async fn run<F>(self, f: impl FnOnce(Self) -> F)` is now obsolete, there's no universal setup&cleanup which is valid for every test. In fact, `run` is not doing anything useful atm, other the running the test itself~~


bobozaur:
@Patrik-Stas Please note that as part of #1023 I actually copied the `test_setup()` function from here. I initially wanted to pull some trickery and allow `Arc<dyn Trait>` in tests but I forgot that the generic arguments in the traits methods make them not trait object safe, so we have to stick with generics. Since I'm moving some things around I figured it could be easier if I copied the implementation there.

bobozaur:
>     * What's the purpose of macro `run_setup_test(...)` as opposed to just calling `build_setup_profile().await.run(...)`
None, really. Just convenience.

>     * I believe that whole run thing `pub async fn run<F>(self, f: impl FnOnce(Self) -> F)` is now obsolete, there's no universal setup&cleanup which is valid for every test. In fact, `run` is not doing anything useful atm, other the running the test itself

I assume it could be removed in this case, indeed.

Patrik-Stas:
This should be ready for re-review. 

- Removed the setup `run(fn)` wraps as they turned out unnecessary
- Removed `unwrap()` from all tests in favor of returning error

gmulhearn (approved):
Lgtm

PR #1029, opened 2023-10-26
feat(mediator): support for pickup protocol processing

nain-F49FF806:


nain-F49FF806:
Todo: tests.

nain-F49FF806:
todo: tests

agents/rust/mediator/mediation/src/router.rs
```diff
@@ -30,7 +27,7 @@ pub async fn create_router() -> Router {
             get(json::echo_message_json).post(respond_message_json),
         )
         .route("/forward", post(handle_forward))
-        .route("/pickup", post(handle_pickup))
+        // .route("/pickup", post(handle_pickup))
```
Patrik-Stas, line 30: what's the endpoint for client agent messages (connection, mediation, pickup) ?

agents/rust/mediator/mediation/src/routes/pickup.rs
```diff
@@ -4,66 +4,81 @@ use std::sync::Arc;
...
+pub async fn handle_pickup_authenticated<T: MediatorPersistence>(
     State(storage): State<Arc<T>>,
-    Json(pickup_message): Json<PickupMsgEnum>,
-) -> (StatusCode, Json<PickupMsgEnum>) {
+    Json(pickup_message): Json<Pickup>,
+    auth_pubkey: &str,
+) -> (StatusCode, Json<Pickup>) {
     match &pickup_message {
-        PickupMsgEnum::PickupStatusReqMsg(status_request) => (
+        Pickup::StatusRequest(status_request) => (
```
Patrik-Stas, line 24: awesome!

agents/rust/mediator/mediation/src/routes/pickup.rs
```diff
@@ -4,66 +4,81 @@ use std::sync::Arc;
 
 use axum::{extract::State, http::StatusCode, Json};
 use log::info;
-
-use crate::{
-    didcomm_types::{
-        pickup_delivery_message_structs::*, PickupDeliveryReqMsg, PickupMsgEnum, PickupStatusMsg,
```
Patrik-Stas, line 10: I suppose you could also delete these custom structures such as `PickupDeliveryReqMsg` as part of this PR? Or is it still used somehow?

agents/rust/mediator/mediation/src/routes/pickup.rs
nain-F49FF806: :)

agents/rust/mediator/mediation/src/router.rs
nain-F49FF806: All client agent messages go through the `/didcomm` endpoint, which is part of the `mediator` src. 

Context: This router file is part of `mediation` crate (previously known as `xum-server`) and is basically dead code at this point. This router is unused by the `mediator`. We are only using the handler functions from this crate, some message structures and persistence functions. Not its server.

agents/rust/mediator/mediation/src/routes/pickup.rs
nain-F49FF806: Yes. All these need to be removed. In fact, more, the whole router module (mentioned in another comment) needs to go too. 
I intended to remove it as part of another PR where I will be removing all the cruft from mediation crate.


Currently, I am focusing on only plumbing: making things work by using relevant function from this crate.

agents/rust/mediator/mediation/src/routes/pickup.rs
```diff
@@ -4,66 +4,81 @@ use std::sync::Arc;
...
             StatusCode::OK,
-            handle_pickup_status_req(status_request, storage).await,
+            handle_pickup_status_req(&status_request.content, storage, auth_pubkey).await,
         ),
         // Why is client sending us status? That's server's job.
-        PickupMsgEnum::PickupStatusMsg(_status) => (
+        Pickup::Status(_status) => (
             StatusCode::BAD_REQUEST,
-            handle_pickup_type_not_implemented().await,
+            handle_pickup_default_status(storage, auth_pubkey).await,
```
Patrik-Stas, line 31: With respect to your comment above, what is this handling we call here? We should probably really just return `BAD_REQUEST` http response, but I don't see what any `handle_*` function is actually called upon receiving invalid request.

On slightly other note, from mediator client perspective, even more graceful response than getting `400` HTTP error, would be to actually return `200` with properly encrypted `ProblemReport` reporting that a message they submitted was unexpected.

agents/rust/mediator/mediation/src/routes/pickup.rs
```diff
@@ -4,66 +4,81 @@ use std::sync::Arc;
...
+use crate::storage::MediatorPersistence;
 
-pub async fn handle_pickup<T: MediatorPersistence>(
+pub async fn handle_pickup_authenticated<T: MediatorPersistence>(
     State(storage): State<Arc<T>>,
-    Json(pickup_message): Json<PickupMsgEnum>,
-) -> (StatusCode, Json<PickupMsgEnum>) {
+    Json(pickup_message): Json<Pickup>,
+    auth_pubkey: &str,
+) -> (StatusCode, Json<Pickup>) {
```
Patrik-Stas, line 22: I could imagine that this function could return custom `Result` type (using custom `MediatorError` type).
In that model you would typically return `Ok(response)` or `Error(<some_err_info_struct/enum>)`

agents/rust/mediator/src/aries_agent/mod.rs
```diff
@@ -166,15 +166,16 @@ impl<T: BaseWallet + 'static, P: MediatorPersistence> Agent<T, P> {
     pub async fn auth_and_get_details(
         &self,
         sender_verkey: &Option<VerKey>,
-    ) -> Result<(String, VerKey, AriesDidDoc), String> {
+    ) -> Result<(String, VerKey, VerKey, AriesDidDoc), String> {
```
Patrik-Stas, line 169: These tuple are starting grow over our heads, it's not easy to see what the function returns without studying the actual implementation. Consider creating "response" structs for these internal responses.

agents/rust/mediator/src/didcomm_handlers/mod.rs
```diff
@@ -35,7 +34,7 @@ pub async fn handle_aries<T: BaseWallet + 'static, P: MediatorPersistence>(
     log::info!("processing message {:?}", &didcomm_msg);
     let unpacked = agent.unpack_didcomm(&didcomm_msg).await.unwrap();
     let aries_message: GeneralAriesMessage =
-        serde_json::from_str(&unpacked.message).expect("Decoding unpacked message as AriesMessage");
+        serde_json::from_str(&unpacked.message).map_err(|e| e.to_string())?;
```
Patrik-Stas, line 37: As mentioned elsewhere, consider introducing custom `MediatorError` error type, along with something like `MediatorResult<T>` as alias to `Result<T, MediatorError>`. See how we do the same in many of our other crates.

agents/rust/mediator/src/didcomm_handlers/mod.rs
```diff
@@ -47,13 +46,21 @@ pub async fn handle_aries<T: BaseWallet + 'static, P: MediatorPersistence>(
             handle_routing_forward(agent.clone(), forward).await?;
             return Ok(Json(json!({})));
         } else {
-            let (account_name, our_signing_key, their_diddoc) =
+            // Auth known VerKey then process account related messages
```
Patrik-Stas, line 49: this seems like a "todo", consider prefixing the comment text with `// todo: `, it tends to have different highlighting in IDEs

agents/rust/mediator/src/didcomm_handlers/mod.rs
```diff
@@ -47,13 +46,21 @@ pub async fn handle_aries<T: BaseWallet + 'static, P: MediatorPersistence>(
...
+                GeneralAriesMessage::AriesVCXSupported(AriesMessage::Pickup(pickup_message)) => {
+                    let pickup_response =
+                        handle_pickup_protocol(&agent, pickup_message, &auth_pubkey).await?;
+                    let aries_response = AriesMessage::Pickup(pickup_response);
+                    let aries_response_bytes =
+                        serde_json::to_vec(&aries_response).map_err(string_from_std_error)?;
+                    agent
+                        .pack_didcomm(&aries_response_bytes, &our_signing_key, &their_diddoc)
+                        .await?
+                }
```
Patrik-Stas, line 63: bit of a duplication between the `GeneralAriesMessage::AriesVCXSupported(AriesMessage::Pickup(pickup_message))` and `GeneralAriesMessage::XumCoord(coord_message)` branches, maybe could be refactored?

agents/rust/mediator/tests/mediator-protocol-pickup.rs
```diff
@@ -0,0 +1,263 @@
...
+        agent,
+        agent_aries_transport,
+        agent_verkey,
+        mediator_diddoc,
+    )
+    .await?;
+    Ok((agent_recipient_key, agent_diddoc))
+}
+
+async fn forward_dummy_anoncrypt_message(
```
Patrik-Stas, line 116: dummy?

agents/rust/mediator/tests/mediator-protocol-pickup.rs
Patrik-Stas, line 1: I didn't go my way to compare this test file against `mediator-routing-forward.rs` test, but I would expect the code be largely identical. Correct me if I am missing something, but I imagine this file would do the same os `mediator-routing-forward`, but add the step of picking up a received message. 
However the both files are long. So either I am missing something, or there should be a way to reuse a "workflow" / extract big chung of code shared between the 2 test suites.

Patrik-Stas (changes requested):

agents/rust/mediator/mediation/src/routes/pickup.rs
nain-F49FF806: Currently, I am responding with a simple Pickup Status message on receipt of any non-request type pickup message that the client sends.

Here for example, if client sends mediator a `status` message (they should be sending `status-request`), then we respond with the correct `status` message, assuming that is what the client actually, wants. 

Agreed, problem report feels more elegant. For the time being, I was having the handler functions map from Pickup (receipt) to Pickup (response). To keep everything within the same protocol at protocol handler level. 

Ideally, these protocol handler functions should return a Result<Pickup> so the error case can be converted to Problem Report messages. In combination with your `MediatorError` suggestion, for example. 

Is it okay if I do this as part of another PR, as it'd be a change for all protocols and not just pickup.

agents/rust/mediator/src/didcomm_handlers/mod.rs
nain-F49FF806: This one is just a comment. The `auth_and_get_details` method does the auth. 
Just had the comment to explain the latter flow. 

Thanks for tip about `// todo:`

agents/rust/mediator/tests/mediator-protocol-pickup.rs
nain-F49FF806: This is done in the another PR that depends on this PR. 

Please have a look at the changes to the test code here  #1040

agents/rust/mediator/tests/mediator-protocol-pickup.rs
nain-F49FF806: It's simply a Basic Message with a trivial hello string wrapped in an encrypted envelope, used for the test.
 
Thanks for the catch, I shall rename to `forward_basic_anoncrypt_message` as this is could cause confusion that it isn't a real message (it is!).

agents/rust/mediator/mediation/src/didcomm_types.rs
```diff
@@ -73,15 +73,13 @@ impl PickupStatusMsg {
...
 pub struct PickupStatusReqMsg {
     #[serde(default)]
-    pub auth_pubkey: String,
     pub recipient_key: Option<String>,
 }
 
 #[skip_serializing_none]
 #[derive(Serialize, Deserialize, Debug)]
 pub struct PickupDeliveryReqMsg {
     #[serde(default)]
```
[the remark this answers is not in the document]
nain-F49FF806: Done

agents/rust/mediator/src/didcomm_handlers/mod.rs
nain-F49FF806: I looked into this. Currently, I don't see a way to. 

Since both the structs are of different types from different sources, there's subtle difference in how they need to be serialised, which is mostly what is going on there. 

I could join the flow immediately after serialising, and have common packing, but that would offer very little advantage..

Once MediatorCoord protocol messages are part of `aries-vcx/messages` this should be easier to do.

Patrik-Stas: ah yeah, i see what you mean. Yeah this should be possible to simplify later

agents/rust/mediator/tests/mediator-protocol-pickup.rs
Patrik-Stas: nice :-)

agents/rust/mediator/tests/mediator-protocol-pickup.rs
Patrik-Stas: Ah, I've missed the fact this is part of testing

agents/rust/mediator/src/didcomm_handlers/mod.rs
Patrik-Stas: Adding custom errors is sounds like a separate PR though :-)

agents/rust/mediator/mediation/src/routes/pickup.rs
Patrik-Stas: > Here for example, if client sends mediator a status message (they should be sending status-request), then we respond with the correct status message, assuming that is what the client actually, wants.

You are too kind to these buggy clients 😆 

> do this as part of another PR
Fair enough, let's leave this for later

Patrik-Stas (approved):
Some stuff has been addressed, some improvements we agreed to do in separate PRs, some of my comments you acknowledged, but I am not sure if you want to address them in this PR or separately. We can merge this but please keep track of the things I've noted. You already have stuff planned out so I don't want to hold you back on a PR when your other upcoming work is already built on this one. 

Let me know if we shall merge

PR #1043, opened 2023-10-31
Enhance `aries_vcx` crate readme

Patrik-Stas:
- Rewrote `aries_vcx` crate readme
- Rename `aries-vcx` package to `aries_vcx`, looking at most of are other crates, it seems that underscore is "the way" :-) I like it though

Also want to point out the distinction between `aries-vcx` and `aries_vcx`:
- `aries-vcx` being name of the HL project currently containing variety of useful components, not particularly tied to aries itself - such as crates `did`, `did_doc`, `did_peer` and number of others
- `aries_vcx` is the main, but also just one of many crates in the repo

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -133,7 +133,7 @@ jobs:
           sudo apt-get install -y libsodium-dev libssl-dev libzmq3-dev
       - name: "Verify clippy across the entire workspace with default features"
         run: |
-          cargo clippy -p aries-vcx --features legacy_proof --features ${{ matrix.backend }} --no-default-features
+          cargo clippy -p aries_vcx --features legacy_proof --features ${{ matrix.backend }} --no-default-features
         env:
           RUSTFLAGS: -D warnings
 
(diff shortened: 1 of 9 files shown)


nain-F49FF806:
I think this is a nice change. It gives a more clear and relevant at glance view to someone approaching `aries_vcx` afresh. 
Combined with the readme on the project. 

The change to underscores is welcome consistency. And should offer slightly lower cognitive effort keeping track of delimiters. 


On the other front, I think we could do more to differentiate *aries-vcx* (project) vs `aries_vcx` (crate), 
which may be confusing for someone not yet familiar with the project's codebase. 

Is it feasible and appropriate to rename the project as "Aries VCX" / "AriesVCX" ?
Capitalisation would help better distinguish between the overarching project (Title Case) and individual crates of code (snake_case). 

[ACA-Py] for example has hyphenated repo but title case project name. Similarly with [Firefly]. 
On casual glance, one other hyperledger repo that retains kebab-case seems to be a single crate library.
[anoncreds-rs], 

[ACA-Py]: https://github.com/hyperledger/aries-cloudagent-python#hyperledger-aries-cloud-agent---python--

[Firefly]: https://github.com/hyperledger/firefly#hyperledger-firefly

[anoncreds-rs]: https://github.com/hyperledger/firefly#hyperledger-firefly

Patrik-Stas:
> Is it feasible and appropriate to rename the project as "Aries VCX" / "AriesVCX" ?

A long time ago we came to agreement we will call the repo "aries-vcx", there was a time of many inconsistencies...at the time of that decision, the repo was pretty much ONLY `aries_vcx` the crate. We had a thoughts lingering recently about bigger change in term of repo name, but I think what you propose is reasonable change, making the repo more human readable, less technical - eg "Aries VCX" / "AriesVCX". Just not sure which one of the two

gmulhearn (approved):
Change sounds good to me!

nain-F49FF806 (approved):

Issue #1047, opened 2023-11-06
Improve `did_parser` crate onboarding


Patrik-Stas:
`did_parser` crate is missing some developer onboarding.

Let's:
  - add `readme.md` with some basic instructions and info about the crate
  - add example into `examples/` directory, demonstrating basic usage and function of the crate

xprazak2:
Fixed by #1051, closing.

PR #1051, opened 2023-11-06
fix(did_parser): Add readme and simple example (#1047)

xprazak2:


did_parser/examples/did_parser/main.rs
```diff
@@ -0,0 +1,10 @@
+use did_parser::{Did, DidUrl};
+
+fn main() {
+    // parse a string into DID
+    Did::parse("did:web:w3c-ccg.github.io".into()).expect("should correctly parse did");
```
Patrik-Stas, line 5: Lets not use `expects` under examples - it distorts the attention toward unhappy path which is not expected to happen in examples. Lets assume the implementation is actually correct (this crate is well unit tested)

Secondly I would suggest instead `println` the parsed DIDs - so if I am newcomer and run examples, I can see some interesting output.

did_parser/README.md
Patrik-Stas, line 1: Let's expand this bit more - perhaps draw inspiration from this one 
https://github.com/hyperledger/aries-vcx/tree/main/did_peer

Mainly that we:
- include link to specs
- include installation instructions (I mean, it's dead simple but I love when readmes enables me to be to be braindead)
- instruction to run demo (partly just to draw attention to the fact a example executables exists)

Patrik-Stas (changes requested):

did_parser/examples/demo.rs
Patrik-Stas, line 1: - Let's demonstrate full public interface of `Did` (there's just some 3 getters)
- Same for `DidUrl`

Patrik-Stas (changes requested):

Patrik-Stas (approved):

PR #1052, opened 2023-11-07
Add mediator coordination messages to aries-vcx messages crate

nain-F49FF806:
Ref: https://github.com/hyperledger/aries-rfcs/blob/main/features/0211-route-coordination/README.md

### DIFF
--- a/messages/src/msg_fields/protocols/coordinate_mediation/keylist.rs
+++ b/messages/src/msg_fields/protocols/coordinate_mediation/keylist.rs
@@ -1,0 +1,87 @@
+use serde::{Deserialize, Serialize};
+use typed_builder::TypedBuilder;
+
+use crate::{decorators::thread::Thread, msg_parts::MsgParts};
+
+pub type Keylist = MsgParts<KeylistContent, KeylistDecorators>;
+
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq, TypedBuilder)]
+pub struct KeylistContent {
+    pub keys: Vec<KeylistItem>,
+    #[builder(default, setter(strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pagination: Option<KeylistPagination>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq, TypedBuilder)]
+pub struct KeylistItem {
+    pub recipient_key: String,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq, TypedBuilder)]
+pub struct KeylistPagination {
+    #[builder(default, setter(strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    count: Option<u64>,
+    #[builder(default, setter(strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    offset: Option<u64>,
+    #[builder(default, setter(strip_option))]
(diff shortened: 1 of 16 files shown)


messages/src/msg_fields/protocols/coordinate_mediation/mediate_request.rs
```diff
@@ -0,0 +1,47 @@
+use serde::{Deserialize, Serialize};
+use typed_builder::TypedBuilder;
+
+use crate::{decorators::thread::Thread, msg_parts::MsgParts};
+
+pub type MediateRequest = MsgParts<MediateRequestContent, MediateRequestDecorators>;
```
Patrik-Stas, line 6: Given the nature of how these messages, I can't quite imagine practical scenario where the thread decorator would be used. I wouldn't mind keeping the second generic param at its default `NoDecorators` value. But not strongly opposing the current implementation either.

messages/src/msg_fields/protocols/coordinate_mediation/keylist_query.rs
```diff
@@ -0,0 +1,71 @@
+use serde::{Deserialize, Serialize};
+use typed_builder::TypedBuilder;
+
+use crate::{decorators::thread::Thread, msg_parts::MsgParts};
+
+pub type KeylistQuery = MsgParts<KeylistQueryContent, KeylistQueryDecorators>;
```
Patrik-Stas, line 6: We don't have this across the board, but I just though it would be nice to leave a comment to the message spec (with fragment in the url), for example here

https://github.com/hyperledger/aries-rfcs/blob/main/features/0211-route-coordination/README.md#key-list-query

messages/src/msg_fields/protocols/coordinate_mediation/keylist.rs
```diff
@@ -0,0 +1,91 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq, TypedBuilder)]
+pub struct KeylistItem {
+    pub recipient_key: String,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq, TypedBuilder)]
+pub struct KeylistPagination {
+    #[builder(default, setter(strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    count: Option<u64>,
```
Patrik-Stas, line 25: while `pagination` field itself is (correctly) optional within `KeylistContent`, I don't think these individual fields are meant to be optional. If mediator decides to provide the pagination information, I think it should be complete. Also don't see indication of otherwise in the rfc

Patrik-Stas (changes requested):

messages/src/msg_fields/protocols/coordinate_mediation/mediate_request.rs
nain-F49FF806: I agree. The thread decorator does not seem very useful in request type messages. (Maybe relevant in response messages, quoting the request message `id`)
Using the default `NoDecorator` also makes message building slightly less verbose. 
In case some other Decorator fields seem relevant, we can later add them back.

messages/src/msg_fields/protocols/coordinate_mediation/keylist_query.rs
nain-F49FF806: Done

Patrik-Stas (approved):

PR #1057, opened 2023-11-10
fix(aries_vcx): make tests run in parallel

### LINKED ISSUE #1050: Enable running `aries_vcx` tests in parallel
Currently, when running `aries_vcx` integration tests, it is a MUST to specify `RUST_TEST_THREADS=1`. That's because all tests are writing transactions on the ledger using the same DID. Hence when running in parallel, one test can invalidate assumptions of another test, made about the ledger state.

To solve this, tests which WRITE something on the ledger shall use distinct DID (Decentralized IDentifier). Indy is permissioned public ledger, meaning that:
- anyone can READ
- only selected entities are allowed to WRITE.

When sending a WRITE tx on the ledger, the send must identify themselves by DID and prove ownership of that DID by respective key. How does ledger knows about what DID exists, and what's the public key for that DID? This information had to be previously submitted to the ledger as a NYM type transaction (FYI: Indy ledger support number of different transaction types - see "Domain Ledger" section of https://hyperledger-indy.readthedocs.io/projects/node/en/latest/transactions.html )
Additionally, the ledger performs not only authentication to decide whether the write is allowed (tx coming from known entity), but there's also authorization. Not all DIDs written on Indy ledger are equal. Different DIDs have different ROLEs - most notable `TRUSTEE` and `ENDORSER`. You can find details in Indy node docs, but the important bottom line is:
- `TRUSTEE` - this is higher privilige role, which among other can write on ledger new DIDs (with various roles)
- `ENDORSER` - is lower "ranked" that `TRUSTEE` and can do transactions needed to perform credential issuance, but can not write new DIDs on the ledger

When we run integration tests, the ledger contains pre-written a DID with Trustee role. We use this DID in every single tests - that's where potential conflicts in parallel setup can come from. And why we currently run `aries_vcx` tests with `RUST_TEST_THREADS=1`.
What should be done is:
- at start of each test, use the default `TRUSTEE` ranked DID to create new random DID with `ENDORSER` role
- in given test, only use the new `ENDORSER` DID afterwards.

xprazak2:
fixes #1050

### DIFF
--- a/aries_vcx/tests/test_pool.rs
+++ b/aries_vcx/tests/test_pool.rs
@@ -39,7 +39,7 @@ use test_utils::{
 use crate::utils::{
     create_and_publish_test_rev_reg, create_and_write_test_cred_def, create_and_write_test_schema,
     scenarios::attr_names_address_list,
-    test_agent::{create_test_agent, create_test_agent_trustee},
+    test_agent::{create_test_agent, create_test_agent_endorser, create_test_agent_trustee},
 };
 
 pub mod utils;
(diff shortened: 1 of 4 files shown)


aries_vcx/tests/utils/test_agent.rs
```diff
@@ -81,3 +83,38 @@ pub async fn create_test_agent(
 > {
     create_test_agent_from_seed(&generate_random_seed(), genesis_file_path).await
 }
+
+pub async fn create_test_endorser<LW, W>(
```
Patrik-Stas, line 87: nitpick: let's call this `create_test_agent_endorser` for consistency

Alternatively we could drop `_test_` from all of these methods keeping it just `create_agent_<foobar>`

aries_vcx/tests/test_pool.rs
```diff
@@ -196,15 +209,28 @@ async fn test_pool_write_new_endorser_did() -> Result<(), Box<dyn Error>> {
 #[ignore]
 async fn test_pool_add_get_service_public() -> Result<(), Box<dyn Error>> {
     let setup = build_setup_profile().await;
-    let did = setup.institution_did.clone();
+    let acme = create_test_endorser(
```
Patrik-Stas, line 212: nitpick: let's call this `agent_endorser` or perhaps just `endorser`

Patrik-Stas (changes requested):
Just some naming comments

PR #1058, opened 2023-11-11
Uniffi demo holder

swaptr:


uniffi_aries_vcx/demo/app/src/main/java/org/hyperledger/ariesvcx/AppDemoController.kt
```diff
@@ -110,4 +121,45 @@ class AppDemoController : ViewModel() {
...
+        onOfferReceived = onShowDialog
+    }
+
+    fun processOfferRequest() {
+        holder?.prepareCredentialRequest(profile!!, "4xE68b6S5VRFrKMMG1U95M")
+        val message = holder?.getMsgCredentialRequest()
+        connection?.sendMessage(profile!!, message!!)
+
+        _state.update { it.copy(offerReceived = false) }
+    }
```
swaptr, line 135: @gmulhearn the credential is not accepted here (atm I am sort of forcing the credential to be accepted).

uniffi_aries_vcx/demo/app/src/main/res/raw/transactions
Patrik-Stas, line 1: perhaps you could name this file something else - whatever network this is representing, currently it's not quite clear

But I think it would be also valuable to keep the "localhost" genesis in place as well. For anyone coming and wanting to build and develop locally, having localhost as default would be the minimal friction dev configuration

gmulhearn: agreed on having a name which is specific to the network. 

Altho worth noting that localhost won't work given the most common usecase of this being built onto a separate android device. Maybe unless people are working with an emulator, not sure how networking works on emulators...

uniffi_aries_vcx/demo/app/src/main/java/org/hyperledger/ariesvcx/HolderScreen.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    LaunchedEffect(Unit) {
+        demoController.awaitCredentialPolling()
+    }
+
+    if (demoState.offerReceived) {
+        AlertDialog(
+            onDismissRequest = { },
+            title = { Text("Accept this invitation?") },
```
gmulhearn, line 42: nit: should say "Accept this credential?"

gmulhearn (changes requested):
Left some comments above, most just clean up stuff.

Also a general note: try to ensure that Kotlin calls into rust/uniffi are done within `Dispatchers.IO`. i.e. `..launch(Dispatchers.IO) {.. }`

For instance, `processOfferRequest` is launching with `viewModelScope.launch`, but you'd want to launch with `.launch(Dispatchers.IO)`. If you don't do this, then the Uniffi/rust calls are going to block the main UI thread, which is what causes UI freezes etc

uniffi_aries_vcx/demo/app/src/main/java/org/hyperledger/ariesvcx/HolderScreen.kt
```diff
@@ -0,0 +1,70 @@
...
+    val scope = rememberCoroutineScope()
+    val context = LocalContext.current
+
+    LaunchedEffect(Unit) {
+        demoController.awaitCredentialPolling()
+    }
+
+    LaunchedEffect(Unit) {
+        demoController.awaitCredentialPolling()
+    }
```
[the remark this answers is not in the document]
swaptr: Yes I can confirm that the polling stops (more like pauses) when switching between screens.

gmulhearn (approved):

PR #1059, opened 2023-11-12
Reorganize crates

Patrik-Stas:


gmulhearn-anonyome:
should this readme be moved? https://github.com/hyperledger/aries-vcx/tree/reorg1/agents/rust

gmulhearn (dismissed):

aries/public_key/Cargo.toml
mirgee, line 1: Is this crate really strictly aries-specific?

aries/shared_vcx/Cargo.toml
mirgee, line 1: AFAIU, this crate is intended to be used from anywhere in the repo, and so is not strictly aries-specific either?

mirgee (changes requested):
I second George's point above; moreover, are there some pending changes to the messages crates or can the be grouped too?

aries/shared_vcx/Cargo.toml
Patrik-Stas: in practice, this is only used by vcx related crates and should probably stay that way. It contains number of dependencies you probably don't want to bring in, if you consume small subset of the crate from isolated project.

Outside of Aries scope, `shared_vcx` is  temporarily only used in `did_parser` 

image

Patrik-Stas:
@mirgee didn't move messages as the did-exchange PR is modifying those

aries/shared_vcx/Cargo.toml
mirgee: I know it is as I have intentionally not used `shared_vcx` in the `did_*` crates, but conceptually there is nothing that makes the crate reusable only by "aries" crates (and even if it was, it doesn't necessarily mean it should be located in this folder) and can be potentially reused by any future "non-aries" crate.

mirgee (approved):

Issue #1045, opened 2023-11-02
Reorganize repo directories


Patrik-Stas:
Moving here the discussion from PR https://github.com/hyperledger/aries-vcx/pull/998

Current structure proposal:

- .github
- docs
- aries 
  - shared_vcx
  - legacy
    - did_doc_legacy
    - libvdrtools
    - agency_client
  - agents
    - rust
       - mediator
       - aries-agent-rust
    - node
    - app-kotlin-swift
  - aries_vcx
  - aries_messages
    - messages
    - messages_macros
  - wrappers
    - node
        - libvcx_core
        - node
        - vcx-napi-rs
    - uniffi-aries-vcx
- aries_vcx_core (rename to something else)
   - vcx_wallet
   - vcx_anoncreds 
   - vcx_ledger
- did_document
    - did_doc
    - did_doc_core
    - did_doc_sov
- did
    - public_key
    - did_parser
    - did_methods
        - did_resolver_core
        - did_key
        - did_peer
        - did_sov 
        - did_web
        - did_registry
- tools
    - simple_message_relay
    - wallet_migrator

- The uncertain thing is naming and structure of `aries_vcx_core`, but it will be definitely renamed
- I've slightly renamed some of the did methods crates, eg `did_resolver_sov` to just `did_sov`
- Not sure If I haven't misplaced `public_key` or `did_key`

The main idea is to clearly separate aries components and other reusable components (even outside of ariex context) such as `did_document`, `did`, perhaps whather `aries_vcx_core` will endup being called

PR #1061, opened 2023-11-13
refactor: Move agents to new root (aries) as per #1045

nain-F49FF806:


Patrik-Stas (changes requested):
Thanks Naian, for now please don't move `aries-vcx-agent`... there's did-exchange related PRs doing modifications in that area, so just to make sure rebasing is as simple as possible. We'll move that one afterwards

PR #1042, opened 2023-10-31
draft: refactored primitives interfaces

bobozaur:
PoC redesign of `Wallet`, `Ledger` and `Vc`(`Anoncreds`) traits.

The `Wallet` and `Ledger` traits were split an designed to work with the `WalletRecord` and `LedgerRequest` traits.

The `WalletRecord` and `LedgerRequest` traits are meant to be implemented on data types that would convert to/from an associated type for the `Wallet` or `Ledger` respectively, enabling the interfaces to work with multiple sets of VC data structures, for instance.

The `Anoncreds` trait is split on roles into `VcIssuer`, `VcVerifier` and `VcProver`. Strong typing and associate types have been added for them.

The `Vc` traits are meant to work with the `Wallet` and `WalletRecord` and have some intense trait bounds that essentially allow a couple of things:
- the ability to store a record in the wallet.
- the ability to retrieve a record from the wallet using the designated ID.
- conversion/linkage between ID data types for the `Wallet` and `WalletRecord` that enables the things above.

There are some compile time tests for the `VcIssuer` implementation for `indy` to ensure that the traits and their extreme trait bounds are actually usable from consumer code. (`VcVerifier` is very simple).

**NOTE**: The trait bounds were a lot of work and involved a lot of trial and error. It's possible some could be removed without affecting the code functionality. Feel free to experiment, as the compiler will surely yell at you if something needed is removed 😅 .

**IMPORTANT**: While this is sort of replicating the `credx` implementation for verifiable credentials, it must be reviewed and refactored to use more static typing here and there, as well as ensuring it does the proper thing as some stuff was removed as part of adding strong types for arguments.

PR #1063, opened 2023-11-16
feat: support aries-askar

xprazak2:


aries_vcx_core/src/wallet/askar/askar_wallet.rs
```diff
@@ -0,0 +1,289 @@
...
+    ) -> Result<Self, AriesVcxCoreError> {
+        let backend = Store::open(db_url, key_method, pass_key, profile.clone()).await?;
+
+        Ok(Self { backend, profile })
+    }
+
+    pub async fn create_key(
+        &self,
+        name: &str,
+        alg: KeyAlg,
```
xprazak2, line 53: We will need our own type for `alg` should this method become a part of public interface.

aries_vcx_core/src/wallet/askar/askar_wallet.rs
```diff
@@ -0,0 +1,289 @@
...
+        key_name: &str,
+        profile: Option<String>,
+    ) -> Result<(), AriesVcxCoreError> {
+        let mut session = self.backend.session(profile).await?;
+
+        let key_entry = self.fetch_key_entry(&mut session, key_name).await?;
+
+        let local_key = key_entry.load_local_key()?;
+
+        let did_bytes = &local_key.to_public_bytes()?[0..16];
```
xprazak2, line 76: `create_and_store_my_did` from `BaseWallet` always derives did from verkey, but it might be nice to allow did to be passed in here as an argument.

aries_vcx_core/src/wallet/askar/askar_wallet.rs
```diff
@@ -0,0 +1,289 @@
...
+        todo!("Not yet implemented");
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[tokio::test]
+    #[cfg_attr(not(feature = "askar_tests"), ignore)]
```
xprazak2, line 232: Do not run tests yet since they will require additional setup.

aries_vcx_core/src/wallet/askar/askar_wallet.rs
```diff
@@ -0,0 +1,289 @@
...
+        let active_profile = wallet.backend.get_active_profile();
+        assert_eq!(profile, active_profile);
+    }
+
+    #[tokio::test]
+    #[cfg_attr(not(feature = "askar_tests"), ignore)]
+    async fn test_should_open_askar_wallet() {
+        let uri = "postgres://postgres:postgres@localhost/askar";
+
+        AskarWallet::open(uri, Some(StoreKeyMethod::Unprotected), None.into(), None)
```
xprazak2, line 257: We need to make sure the wallet is created first for this test.

aries_vcx_core/src/errors/error.rs
Patrik-Stas, line 1: Ultimately the fact that a particular wallet implementation is being used should not be visible to the rest of the codebase. 
We should rather reuse existing wallet errors - those might not be covering all of our needs, so we could tweak them, rename them, extend them as need - but the general approach should be to for example use `WalletCreate` error type regardless underlying wallet implementation.

aries_vcx_core/Cargo.toml
```diff
@@ -11,9 +11,12 @@ credx = ["dep:indy-credx"]
 vdr_proxy_ledger = ["credx", "dep:indy-vdr-proxy-client"]
 # Feature flag to allow legacy proof verification
 legacy_proof = []
+askar_tests = []
```
Patrik-Stas, line 14: Not only tests, but anything Askar related in the codebase should be behind a feature flag. Given the implementation must stay hidden behind the BaseWallet trait, the rest of the codebase should not know of Askar's existence.
You might be aware of this, but keeping that constraint will "reviewlessly" assure these desired properties.

aries_vcx_core/src/wallet/askar/askar_wallet.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    // ---- records
+
+    async fn add_wallet_record(
+        &self,
+        xtype: &str,
+        id: &str,
+        value: &str,
+        tags: Option<HashMap<String, String>>,
```
Patrik-Stas, line 132: For completeness sake just remind of what we mentioned on the community call earlier.

 The BaseWallet method signatures were catered for the particular vdrtools based wallet implementation. We didn't try to over-generalize for the future at the time we designed `BaseWallet`.
So probably you'll find it desirable to redesign `BaseWallet` trait - you can draw inspiration from this POC https://github.com/hyperledger/aries-vcx/pull/1042  which didn't meant to add support for Askar in particular, but I guess even more ambitiously tried to enable arbitrary wallet implementation in the future. That might or might not been too long shot.

aries_vcx_core/Cargo.toml
Patrik-Stas: For example having it feature flagged from get go would probably naturally lead you to not create `Askar*` specific error kinds. It might be somewhat restrictive while POCing, but ultimately will guide you toward the optimal design/integration with the rest of the codebase

xprazak2:
I added an implementation of `Wallet` trait from #1071. It is very rough but it illustrates the trait usage.

xprazak2:
Closing in favor of https://github.com/hyperledger/aries-vcx/pull/1085

PR #1069, opened 2023-11-25
feature: Derive Display for all aries messages

Patrik-Stas:
- Derives `Display` trait for all aries messages 
- Removed default Decorators type parameter `D = NoDecorators` - turned out somewhat problematic in conjunction with the custom `Display` proc macro. Not a big price to pay though, only slightly inconveniences developer working in `messages` crate, but not `messages` consumers.

### DIFF
--- a/aries/messages/src/msg_fields/protocols/did_exchange/request.rs
+++ b/aries/messages/src/msg_fields/protocols/did_exchange/request.rs
@@ -37,7 +37,7 @@ pub struct RequestDecorators {
 #[cfg(test)]
 #[allow(clippy::unwrap_used)]
 #[allow(clippy::field_reassign_with_default)]
-mod tests {
+pub mod tests {
     use diddoc_legacy::aries::diddoc::AriesDidDoc;
     use serde_json::json;
 
@@ -52,7 +52,7 @@ mod tests {
         msg_types::protocols::did_exchange::DidExchangeTypeV1_0,
     };
 
-    fn request_content() -> RequestContent {
+    pub fn request_content() -> RequestContent {
         let did_doc = AriesDidDoc::default();
         RequestContent {
             label: "test_request_label".to_owned(),
--- a/aries/messages/src/msg_fields/protocols/did_exchange/request.rs
+++ b/aries/messages/src/msg_fields/protocols/did_exchange/request.rs
@@ -37,7 +37,7 @@ pub struct RequestDecorators {
 #[cfg(test)]
 #[allow(clippy::unwrap_used)]
 #[allow(clippy::field_reassign_with_default)]
-pub mod tests {
+mod tests {
     use diddoc_legacy::aries::diddoc::AriesDidDoc;
     use serde_json::json;
 
@@ -49,6 +49,7 @@ pub mod tests {
             timing::tests::make_extended_timing,
         },
         misc::test_utils,
+        msg_fields::protocols::did_exchange::request::{Request, RequestDecorators},
         msg_types::protocols::did_exchange::DidExchangeTypeV1_0,
     };
 
@@ -71,6 +72,18 @@ pub mod tests {
                     .build(),
             ),
         }
+    }
+
+    #[test]
+    fn test_print_message() {
+        let msg: Request = Request::builder()
+            .id("test_id".to_owned())
+            .content(request_content())
+            .decorators(RequestDecorators::default())
+            .build();
+        let printed_json = format!("{}", msg);
+        let parsed_request: Request = serde_json::from_str(&printed_json).unwrap();
+        assert_eq!(msg, parsed_request);
     }
 
     #[test]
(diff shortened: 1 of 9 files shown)


nain-F49FF806 (dismissed):
I can't comment on `display_as_json` code, as that's beyond my understanding, but other than that, LGTM!

aries/messages/src/msg_parts.rs
```diff
@@ -14,9 +14,9 @@ use typed_builder::TypedBuilder;
 /// `~attach` used in some messages that are in fact part of the protocol itself and are
 /// instrumental to the message processing, not an appendix to the message (such as `~thread` or
 /// `~timing`).
-#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, TypedBuilder)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, TypedBuilder, Display)]
 #[builder(build_method(vis = "", name = __build))]
-pub struct MsgParts<C, D = NoDecorators> {
+pub struct MsgParts<C, D> {
```
mirgee, line 19: Can you please specify the problem that the default type was causing?

aries/messages/src/msg_fields/protocols/did_exchange/request.rs
```diff
@@ -52,7 +52,7 @@ mod tests {
         msg_types::protocols::did_exchange::DidExchangeTypeV1_0,
     };
 
-    fn request_content() -> RequestContent {
+    pub fn request_content() -> RequestContent {
```
mirgee, line 55: I would much rather move the test into this module instead of exposing random test utilities from wherever. This practice leads to nothing but headache during test refactoring.

aries/messages/src/msg_parts.rs
Patrik-Stas: Prompted me to dig into this more - the issue was the proc macro was too simplistic, it was generating stuff like 
```
impl<C, D = NoDecorators> std::fmt::Display for MsgParts<C, D>
    where
        C: serde::Serialize,
        D: serde::Serialize,
```
while is should had been 
```
impl<C, D> std::fmt::Display for MsgParts<C, D>
    where
        C: serde::Serialize,
        D: serde::Serialize,
```
This is addressed now, syn crate has useful function `split_for_impl` available for `Generics` type, which allows to work with the generics related statements in more granular fashion.

So I am re-introducing the default NoDecorators type, reverting some of the changes done here

aries/messages/src/msg_fields/protocols/did_exchange/request.rs
Patrik-Stas: The reason why I placed it here is because I am not something did-exchange specific, but rather macro generated on `MsgParts`, and just happened to choose did-exchange as one of the specific MsgParts concrete type for it. 
I can move it if you insist, but I'd argue this will bury the test a level lower where it actually belongs.

aries/messages/src/msg_parts.rs
```diff
@@ -65,3 +65,24 @@ impl<C, D> MsgPartsBuilder<C, D, ((String,), (C,), (D,))> {
...
+
+    #[test]
+    fn test_print_message() {
+        let msg: Request = Request::builder()
+            .id("test_id".to_owned())
+            .content(request_content())
+            .decorators(RequestDecorators::default())
+            .build();
+        let printed = format!("{}", msg);
+        let expected = r#"{"@id":"test_id","label":"test_request_label","goal_code":"aries.rel.build","goal":"test_goal","did":"","did_doc~attach":{"data":{"json":{"@context":"https://w3id.org/did/v1","authentication":[],"id":"","publicKey":[],"service":[{"id":"did:example:123456789abcdefghi;indy","priority":0,"recipientKeys":[],"routingKeys":[],"serviceEndpoint":"https://dummy.dummy/dummy","type":"IndyAgent"}]}}}}"#;
```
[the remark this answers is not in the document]
Patrik-Stas: updated

aries/messages/src/msg_fields/protocols/did_exchange/request.rs
mirgee: I understand the reason for placing it where it is but I would still move it. It can be seen as testing a property or behavior of the DID exchange `Request` message.

aries/messages/src/msg_parts.rs
mirgee: I hope that the fact that the decision to rather modify the sites of usage instead of fixing the macro seemed as a better approach is more about due to skill issue rather than the fact that procedural macros are on harder to maintain :) Probably both, though...

Patrik-Stas: @mirgee I think you misunderstood me, I fixed the macro. No change to callsites anymore

aries/messages/src/msg_fields/protocols/did_exchange/request.rs
Patrik-Stas: moved

aries/messages/src/msg_parts.rs
mirgee: @Patrik-Stas No, I am talking about the initial decision.

Patrik-Stas: Ah yeah. In fact I misread. Indeed exactly as you said, it's definitely objectively better to make the macro more robust.

mirgee (approved):

PR #1072, opened 2023-11-28
feat(mediator): Better errors for MediatorPersistence

nain-F49FF806:
This is the first part of effort to introduce better structured errors to the mediator. 
We are using this_error library for the same. And designing the error hierarchy/mapping from bottom up.

### DIFF
--- a/aries/agents/rust/mediator/src/aries_agent/client/transports.rs
+++ b/aries/agents/rust/mediator/src/aries_agent/client/transports.rs
@@ -27,12 +27,8 @@ pub trait AriesTransport {
     ) -> Result<Value, AriesTransportError>;
 }
 
-pub struct AriesReqwest {
-    pub client: reqwest::Client,
-}
-
 #[async_trait]
-impl AriesTransport for AriesReqwest {
+impl AriesTransport for reqwest::Client {
     async fn send_aries_envelope(
         &mut self,
         envelope_json: Value,
@@ -46,7 +42,6 @@ impl AriesTransport for AriesReqwest {
             serde_json::to_string(&envelope_json).unwrap()
         );
         let res = self
-            .client
             .post(oob_invited_endpoint)
             .json(&envelope_json)
             .send()
--- a/aries/agents/rust/mediator/src/aries_agent/client/transports.rs
+++ b/aries/agents/rust/mediator/src/aries_agent/client/transports.rs
@@ -1,5 +1,3 @@
-use std::collections::VecDeque;
-
 use async_trait::async_trait;
 use diddoc_legacy::aries::diddoc::AriesDidDoc;
 use log::debug;
@@ -30,7 +28,6 @@ pub trait AriesTransport {
 }
 
 pub struct AriesReqwest {
-    pub response_queue: VecDeque<Value>,
     pub client: reqwest::Client,
 }
 
--- a/aries/agents/rust/mediator/src/aries_agent/mod.rs
+++ b/aries/agents/rust/mediator/src/aries_agent/mod.rs
@@ -228,7 +228,8 @@ impl<T: BaseWallet + 'static, P: MediatorPersistence> Agent<T, P> {
     ) -> Result<(), String> {
         self.persistence
             .create_account(their_vk, our_vk, &json!(did_doc).to_string())
-            .await?;
+            .await
+            .map_err(string_from_std_error)?;
         Ok(())
     }
 }
--- a/aries/agents/rust/mediator/src/aries_agent/mod.rs
+++ b/aries/agents/rust/mediator/src/aries_agent/mod.rs
@@ -156,13 +156,11 @@ impl<T: BaseWallet + 'static, P: MediatorPersistence> Agent<T, P> {
             .get_account_details(&auth_pubkey)
             .await
             .map_err(string_from_std_error)?;
-        let diddoc = serde_json::from_value::<AriesDidDoc>(account_details.did_doc_json)
-            .map_err(string_from_std_error)?;
         Ok((
             account_details.account_name,
             account_details.auth_pubkey,
             account_details.our_signing_key,
-            diddoc,
+            account_details.their_did_doc,
         ))
     }
     pub async fn handle_connection_req(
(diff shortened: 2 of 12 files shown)


Patrik-Stas:
Good start, I am wondering about the "LowerLevelError" name, but I am don't have better suggestion either, and given the `transparent` annotation, it probably doesn't even matter much.

Just 1 ask: if this is WIP, can you switch it to `draft` state (option under the reviewers panel)? We didn't do this very strictly before, but seems like a good practice. Then having PRs in non-draft state should prompt maintainers to review them.

nain-F49FF806:
Agreed on renaming the `LowerLevelError` to something else.  I was thinking of renaming it to `StorageLayerError`, but any other suggestions are welcome.  

I think of `LowerLevelError`/`LowerLayerError` as a generic error variant name that can be used across the error hierarchy (as opposed to saying just "Other").

aries/agents/rust/mediator/src/aries_agent/client/transports.rs
```diff
@@ -30,7 +28,6 @@ pub trait AriesTransport {
 }
 
 pub struct AriesReqwest {
-    pub response_queue: VecDeque<Value>,
     pub client: reqwest::Client,
 }
```
Patrik-Stas, line 32: Seems like the Reqwest wrapper doesn't need to exist anymore. But that seems out of scope, feel free to address it separately.

aries/agents/rust/mediator/src/aries_agent/mod.rs
```diff
@@ -151,11 +151,17 @@ impl<T: BaseWallet + 'static, P: MediatorPersistence> Agent<T, P> {
...
+            .persistence
+            .get_account_details(&auth_pubkey)
+            .await
+            .map_err(string_from_std_error)?;
+        Ok((
+            account_details.account_name,
+            account_details.auth_pubkey,
+            account_details.our_signing_key,
+            account_details.their_did_doc,
+        ))
```
Patrik-Stas, line 164: out of scope comment: should return the account details type rather than tuple

aries/agents/rust/mediator/src/persistence/errors.rs
```diff
@@ -0,0 +1,62 @@
...
+
+errorset!(GetAccountIdError[StorageBackendError, AccountNotFound]);
+errorset!(ListAccountsError[StorageBackendError, DecodeError]);
+
+errorset!(GetAccountDetailsError[StorageBackendError, AccountNotFound, DecodeError]);
+errorset!(RetrievePendingMessageCountError[StorageBackendError, AccountNotFound]);
+errorset!(RetrievePendingMessagesError[StorageBackendError, AccountNotFound]);
+errorset!(AddRecipientError[StorageBackendError, AccountNotFound]);
+// Same error modes as AddRecipientError
+pub type RemoveRecipientError = AddRecipientError;
```
Patrik-Stas, line 51: hmmm, perhaps indication of what can be problematic with this error management pattern

The thing, is, `RetrievePendingMessagesError, RetrievePendingMessageCountError, AddRecipientError, RemoveRecipientError` are all the same set of errors, but it's difficult to give the set proper name. The same set of errors might, as we observe, much possibly be returned by various unrelated functions.
I wanted to suggest - why don't we name `[StorageBackendError, AccountNotFound]` set as `ModificationError`, but it would be confusing to return such named error from `retrieve_pending_message_count`. 
On the other hand, if we define `ModificationError` as `[StorageBackendError, AccountNotFound]`, it's possible some other modifying function would actually want to return different set of errors.
So it seems that finding right name for "sets" is just not sensible goal. 

I am commenting on this because of the asymmetry - given that `RetrievePendingMessageCountError, RetrievePendingMessagesError, AddRecipientError` are unique error types, despite representing same set of circumstances underneath, we should probably keep the approach also with  `RemoveRecipientError` and defined is a new type, rather than alias.
Unless there is argument why to give a different treatment to `RemoveRecipientError` and say `RetrievePendingMessagesError`

Patrik-Stas: Also comes to my mind that with this approach of - error type per operation - perhaps it would make sense to defined this errors nearby the function they are being thrown from.

Patrik-Stas (changes requested):
left review, some out of comments, and one longer note about the approach

aries/agents/rust/mediator/src/persistence/errors.rs
nain-F49FF806: > I am commenting on this because of the asymmetry - given that `RetrievePendingMessageCountError, RetrievePendingMessagesError, AddRecipientError` are unique error types, despite representing same set of circumstances underneath, we should probably keep the approach also with `RemoveRecipientError` and defined is a new type, rather than alias. Unless there is argument why to give a different treatment to `RemoveRecipientError` and say `RetrievePendingMessagesError`

One way to think about this is that `RetrievePendingMessagesError`, `RetrievePendingMessageCountError`, `AddRecipientError` are all actually not the same set. They are different sets corresponding to different operations that happen to have the same contents. Each may be individually extended or shrunk as the code/implementation evolves. Hence, they need to be different types.

For example, in future, we could add an error case to `AddRecipientError` ~  `WritesAcessDenied`. Which would not be relevant to the others. 

`RemoveRecipientError`, `AddRecipientError` are slightly different because their functions are performing pretty much the same operation (modification). And are expected to fail similarly.

nain-F49FF806: > Also comes to my mind that with this approach of - error type per operation - perhaps it would make sense to defined this errors nearby the function they are being thrown from.

Yes. I was thinking the same. Here, since these composed errors are used in a trait definition (with no default implementation) I choose to keep them all together along with the base errors. To keep things tidy.

aries/agents/rust/mediator/src/aries_agent/mod.rs
nain-F49FF806: Done

aries/agents/rust/mediator/src/aries_agent/client/transports.rs
nain-F49FF806: Done

Patrik-Stas (approved):

aries/agents/rust/mediator/src/persistence/errors.rs
Patrik-Stas: Fair enough. Let's see how it develops as time goes :-)

Patrik-Stas:
Great improvements, well done 👍 💯

PR #1075, opened 2023-11-29
change: simplify didcore components, did-exchange protocol

Patrik-Stas:
The changes in this PR are primarily stemming by removing generics on `DidDocument` and `Service` data structures. This change trickles-up significantly all the way to did-exchange implementation. This enable many simplifications and removal of lots of code. Some other improvements has been done on the way as well. 

I recommend start looking at the PR from top down to understand public API effect of the changes.

# Demo
- `aries/aries_vcx/tests/test_did_exchange.rs` is the top-most level of changes
### Simplified `DidExchangeRequester`
- Now has single method `construct_request` which simple accepts a `Did` as input - can be **any resolvable DID**.  This means it's also possible to establish didcomm any 2 entities which have DID - be it `did:peer`, `did:sov` or any other DID.
- Trimmed down support for invitations - we do not support OOB invitations with service objects "embedded" into the invitation. This Aries design was objected in https://github.com/hyperledger/aries-rfcs/issues/802. We can implemented later if desired by any vcx users, but solution is ugly - it would be needed to convert Service object to Did Document which doesn't make sense.

These changes also lead to internals simplification where bunch of transformation code could been removed 

### Simplified `DidExchangeResponder`

- Just like requester (invitee), `DidExchangeRequester`(inviter) needs to provide their pairwise `PeerDid` for the counterparty upon constructing `Response` message.  This means that currently the responder always rotates their DID, regardless of what kind of DID they used in invitation. Making rotation optional will be addressed in the future.
- Removes required argument `invitation_id` from `receive_request` method signature. Also, don't keep this information in SM state - the responder might receive request which is not constructed on basis of **any invitation** - but rather simply on basis of knowing responder's public DID. Invitation might not exist in a first place, so it doesn't make sense to store its ID.

# DidCore

### Removed generics
The main goal of the PR was simplification of DidCore components - mainly removing generic type parameters from `DidDocument<E>` and `Service<E>`. This lead to many changes, but in scope of DidCore components:
- Instance of `DidDocument` can represent any valid Did Document with any extra fields. Instance of `Service` can represent any Service object, with any extra fields.
- In practice, in didcomm we usually use certain extra fields on Service object - such as `routingKeys`, `recipientKeys`, `accept` etc. Given the "extra" typing we dropped, the current typing design has following considerations:
  1. When resolving did document, we don't know what extra field we might find. We don't know what resolver will return for a particular DID. We don't know what fields the applications build on top of `did_doc` crate wants to support, work with. Hence it's up to applications to decide either work with `DidDocument`/`Service` as their are, or convert these to custom types in runtime, if typed environment for extra fields is needed.
  2. We also provide fully typed version of common didcomm Services - such as `struct ServiceDidCommV1`, `struct ServiceDidCommV2`. These tend to be useful when the user wants to construct particular service type themselves. Equally, as mentioned in the previous point `i., user can also choose to convert `Service` to more explicit types such as `ServiceDidCommV1` upon resolution. However, in practical terms so far in `aries_vcx`, this was not really needed.

### New features
- Implement new util methods on DidDocument such as `get_service_of_type, get_key_agreement_of_type` to support data lookups within did documents.

### Refactoring 
- Remove custom url wrapper for `url::Url` crate in `did_doc` crate
- Refactoring around did:peer:2 de/abbreviation

### Error handling
- Refactoring of errors to reduce number of variants, for example `DidDocumentBuilderError` now has only single `KeyDecodingError` variant, which itself hold further more details about particular decoding issue
- Introduce new error types in `did_doc` crate, such as `JsonWebKeyError`, `MultibaseWrapperError`, `KeyDecodingError` and others to avoid misusing `DidDocumentBuilderError` in contexts where we are not in fact using did document builder.

# Other changes
### Testing
- introduce `pretty_assertions` https://github.com/rust-pretty-assertions/rust-pretty-assertions#usage as dev dependency to some of the crates and tests 

### Logging
- Derive `Display` for more structs

### Refactoring
- `aries-vcx-agent` did-exchange service is generating invitation with `did:peer` as value of `services`, rather than inlining service object.
- ~~Make `DidPeer<Numalgo2> --- DDO` user friendlier. Instead of calling "util" function `resolve_numalgo2(did PublicKeyEncoding::Base58)` this is now done `did.to_did_doc(PublicKeyEncoding::Base58)`~~
- Didcomm v1 refactoring (`EncryptionEnvelope`)
   - added new constructor which takes in 2 `DidDocument` structures. Noteworthy is that `key_agreement` is used for encryption rather than `recipientKeys` "extra field". The thing is, `peer:did:2` doesn't even define how `recipientKeys` should be abbreviated, I believe it assumes it wouldn't be used anymore.

### Fix
- `did:peer:3` validation regex did not allow DIDs which contains Service only. But all fields in Did Document are optional - except id, did:peer:3.S.* is valid DID too.

# Note
- This PR contains intermediary refactoring work done in https://github.com/hyperledger/aries-vcx/pull/1097


Patrik-Stas:
`aries-vcx -- aries-vcx` AATH did-exchange tests passing
image
image

gmulhearn (approved):
Skimmed thru aries changes. nice stuff

PR #1078, opened 2023-11-30
feat: implement RecordWallet for askar 

xprazak2:
extracted from https://github.com/hyperledger/aries-vcx/pull/1063 as it was growing too large.

I can't really set https://github.com/hyperledger/aries-vcx/pull/1071 as a base for merge since I opened it from my fork, but the commits here are separated, so it should still be possible to view only the implementation part with a relative ease.


xprazak2:
Closing in favour of https://github.com/hyperledger/aries-vcx/pull/1085

PR #1083, opened 2023-12-06
fix: #1053 Base64URLSafe padded and unpadded decoding

### LINKED ISSUE #1053: Tolerance to Base64URLSafe encoding padding (or lack thereof) in Connection Protocol Signatures
During the Connection Response of Connection Protocol, a signature is attached within a `connection~sig` decorator/field. https://github.com/hyperledger/aries-rfcs/tree/main/features/0160-connection-protocol#2-connection-response . 

```
  "connection~sig": {
    "@type": "https://didcomm.org/signature/1.0/ed25519Sha512_single",
    "signature": "<digital signature function output>",
    "sig_data": "<base64URL(64bit_integer_from_unix_epoch||connection_attribute)>",
    "signer": "<signing_verkey>"
  }
```
The spec describes it as so, but does not specify whether `base64URL` is with or without padding. 

There is a note elsewhere on the page which says:
> NOTE: In receiving the invitation, the base64url decode implementation used MUST correctly decode padded and unpadded base64URL encoded data.

This makes me believe that this sentiment would apply for `sig_data` too. As such, we may want to make our code handle both. Currently `common/signing.rs` is hardcoded to use `URL_SAFE` (expects padding). It has been found that the animo.demo.id DEMO (which presumably uses AFJ) can produce `sig_data` which does not have padding. This causes common/signing.rs to throw an InvalidPadding error.

payload from animo:
```
{"@type":"did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/response","@id":"5a557835-13cd-4cbf-b165-96242f138936","connection~sig":{"@type":"https://didcomm.org/signature/1.0/ed25519Sha512_single","sig_data":"AAABi6h0IMJ7IkRJRCI6IjJWMktCQUFGRlpiM2FlelVMTFRKWEUiLCJESUREb2MiOnsiQGNvbnRleHQiOiJodHRwczovL3czaWQub3JnL2RpZC92MSIsInB1YmxpY0tleSI6W3siaWQiOiIyVjJLQkFBRkZaYjNhZXpVTExUSlhFIzEiLCJjb250cm9sbGVyIjoiMlYyS0JBQUZGWmIzYWV6VUxMVEpYRSIsInR5cGUiOiJFZDI1NTE5VmVyaWZpY2F0aW9uS2V5MjAxOCIsInB1YmxpY0tleUJhc2U1OCI6Im90TTJwTHVVM0NTQkdvZ2JHNXAyRUtLZ0ZobU01ZWdQMkxLcU1RdnZEdlMifV0sInNlcnZpY2UiOlt7ImlkIjoiMlYyS0JBQUZGWmIzYWV6VUxMVEpYRSNJbmR5QWdlbnRTZXJ2aWNlIiwic2VydmljZUVuZHBvaW50IjoiaHR0cHM6Ly9kaWRjb21tLmRlbW8uYW5pbW8uaWQiLCJ0eXBlIjoiSW5keUFnZW50IiwicHJpb3JpdHkiOjAsInJlY2lwaWVudEtleXMiOlsib3RNMnBMdVUzQ1NCR29nYkc1cDJFS0tnRmhtTTVlZ1AyTEtxTVF2dkR2UyJdLCJyb3V0aW5nS2V5cyI6W119XSwiYXV0aGVudGljYXRpb24iOlt7InB1YmxpY0tleSI6IjJWMktCQUFGRlpiM2FlelVMTFRKWEUjMSIsInR5cGUiOiJFZDI1NTE5U2lnbmF0dXJlQXV0aGVudGljYXRpb24yMDE4In1dLCJpZCI6IjJWMktCQUFGRlpiM2FlelVMTFRKWEUifX0","signer":"otM2pLuU3CSBGogbG5p2EKKgFhmM5egP2LKqMQvvDvS","signature":"pctzlHij_A2a9RLBqMDQS7iz8J0fsIESSZWV9JW5tOuZcS6VcqN2visj2FjL6Lw7QelVNjY0jbALbbckh0y4Bw"},"~thread":{"thid":"151b33e9-0259-4d01-8b79-341b2a08b36a"}}
```

reproducable error:
```rust
general_purpose::URL_SAFE.decode("pctzlHij_A2a9RLBqMDQS7iz8J0fsIESSZWV9JW5tOuZcS6VcqN2visj2FjL6Lw7QelVNjY0jbALbbckh0y4Bw").unwrap();
```

lukewli-anonyome:
This PR introduces a fix for the Base64URL encoding/decoding outlined in issue #1053.

Closes #1053

### DIFF
--- a/aries/aries_vcx/src/common/signing.rs
+++ b/aries/aries_vcx/src/common/signing.rs
@@ -10,14 +10,15 @@ use crate::errors::error::prelude::*;
 
 // Utility function to handle both padded and unpadded Base64URL data
 fn base64url_decode(encoded: &str) -> VcxResult<Vec<u8>> {
-    general_purpose::URL_SAFE_NO_PAD.decode(encoded).or_else(|_|
-        general_purpose::URL_SAFE.decode(encoded)
-    ).map_err(|err| {
-        AriesVcxError::from_msg(
-            AriesVcxErrorKind::InvalidJson,
-            format!("Cannot decode Base64URL data: {:?}", err),
-        )
-    })
+    general_purpose::URL_SAFE_NO_PAD
+        .decode(encoded)
+        .or_else(|_| general_purpose::URL_SAFE.decode(encoded))
+        .map_err(|err| {
+            AriesVcxError::from_msg(
+                AriesVcxErrorKind::InvalidJson,
+                format!("Cannot decode Base64URL data: {:?}", err),
+            )
+        })
 }
 
 async fn get_signature_data(
--- a/aries/aries_vcx/src/common/signing.rs
+++ b/aries/aries_vcx/src/common/signing.rs
@@ -1,5 +1,9 @@
 use aries_vcx_core::wallet::base_wallet::BaseWallet;
-use base64::{self, engine::general_purpose, Engine};
+use base64::{
+    self, alphabet,
+    engine::{general_purpose, DecodePaddingMode, GeneralPurpose, GeneralPurposeConfig},
+    Engine,
+};
 use messages::msg_fields::protocols::connection::{
     response::{ConnectionSignature, ResponseContent},
     ConnectionData,
@@ -8,17 +12,23 @@ use time;
 
 use crate::errors::error::prelude::*;
 
+/// A default [GeneralPurposeConfig] configuration with a [decode_padding_mode] of
+/// [DecodePaddingMode::Indifferent]
+pub const LENIENT_PAD: GeneralPurposeConfig =
+    GeneralPurposeConfig::new().with_decode_padding_mode(DecodePaddingMode::Indifferent);
+
+/// A [GeneralPurpose] engine using the [alphabet::URL_SAFE] base64 alphabet and
+/// [DecodePaddingMode::Indifferent] config to decode both padded and unpadded.
+pub const URL_SAFE_LENIENT: GeneralPurpose = GeneralPurpose::new(&alphabet::URL_SAFE, LENIENT_PAD);
+
 // Utility function to handle both padded and unpadded Base64URL data
 fn base64url_decode(encoded: &str) -> VcxResult<Vec<u8>> {
-    general_purpose::URL_SAFE_NO_PAD
-        .decode(encoded)
-        .or_else(|_| general_purpose::URL_SAFE.decode(encoded))
-        .map_err(|err| {
-            AriesVcxError::from_msg(
-                AriesVcxErrorKind::InvalidJson,
-                format!("Cannot decode Base64URL data: {:?}", err),
-            )
-        })
+    URL_SAFE_LENIENT.decode(encoded).map_err(|err| {
+        AriesVcxError::from_msg(
+            AriesVcxErrorKind::InvalidJson,
+            format!("Cannot decode Base64URL data: {:?}", err),
+        )
+    })
 }
 
 async fn get_signature_data(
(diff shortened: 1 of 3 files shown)


aries/aries_vcx/src/common/signing.rs
```diff
@@ -8,6 +8,19 @@ use time;
...
+fn base64url_decode(encoded: &str) -> VcxResult<Vec<u8>> {
+    general_purpose::URL_SAFE_NO_PAD
+        .decode(encoded)
+        .or_else(|_| general_purpose::URL_SAFE.decode(encoded))
+        .map_err(|err| {
+            AriesVcxError::from_msg(
+                AriesVcxErrorKind::InvalidJson,
+                format!("Cannot decode Base64URL data: {:?}", err),
+            )
+        })
```
gmulhearn-anonyome, line 21: nice, this should definitely work. However as a nitpick, it would be preferred to intentionally trigger the error path in order to try `general_purpose::URL_SAFE.decode`.

In other languages, this would sort of be like wrapping it with a `try catch` block, and trying again in the `catch` block. this is usually an anti-pattern because throwing exceptions has cost. In Rust there are no exceptions, just Errors, and Errors are less costly... however IMO it would still be nicer to avoid this pattern.

Additionally, I'm only making this comment because I've had experience with this situation when using the `base64` crate before and found that there is a better way. `URL_SAFE_NO_PAD` and `URL_SAFE` are pre-made instances of `GeneralPurpose` base64 engines. I'd have a look into these implementations and see if you can construct your own `GeneralPurpose` engines which use the URL_SAFE alphabet, but knows how to handle with or without padding (i.e. it is lenient/indifferent to padding when decoding). 

Once you construct one of those, i'd make it into a `const URL_SAFE_<some appropriate name>: GeneralPurpose` like how they do with `URL_SAFE_NO_PAD` and `URL_SAFE`.

gmulhearn-anonyome: oh also, if you could, maybe add a comment above the const explaining what it is, and maybe explain that in the future this const engine could be used globally in the crate:
```
NOTE: we may wish to move and use this engine as a global util in the aries-vcx project
```

Also note, there is an example of all of this within the aries-vcx repo... i forgot where..

aries/aries_vcx/src/common/signing.rs
```diff
@@ -146,3 +152,22 @@ pub async fn decode_signed_connection_response(
...
+        let encoded_padded = general_purpose::URL_SAFE.encode(data);
+        let encoded_unpadded = general_purpose::URL_SAFE_NO_PAD.encode(data);
+
+        let decoded_padded = base64url_decode(&encoded_padded).unwrap();
+        let decoded_unpadded = base64url_decode(&encoded_unpadded).unwrap();
+
+        assert_eq!(decoded_padded, data.to_vec());
+        assert_eq!(decoded_unpadded, data.to_vec());
+    }
+}
```
gmulhearn-anonyome, line 173: nit; for completeness... could you pls test the error case too? (checking that `InvalidJson` is thrown)

Patrik-Stas:
Hi @lukewli-anonyome! Welcome and thank you for your contribution! George will handle the review here, apart from warm welcome just one small tip: the CI is failing on formatting, you can always easily verify formatting running `cargo +nightly-2023-05-08 fmt --check` and fix the formatting as follows. `cargo +nightly-2023-05-08 fmt --check`.
We have currently hardcoded this particular nightly version of rust for formatting verification in CI, hence the strange looking command ;-)

gmulhearn-anonyome (approved):
nice, thanks for the effort!

gmulhearn (approved):

PR #1084, opened 2023-12-06
feat: add new Wallet trait, implement for indy wallet

xprazak2:
Obsoletes https://github.com/hyperledger/aries-vcx/pull/1071 and takes an approach outlined in https://github.com/hyperledger/aries-vcx/pull/1071#pullrequestreview-1759757590, meaning:

* no associated types
* new trait does not expose any implementation-specific types
* initial implementation for vdrtools wallet

### DIFF
--- a/aries/aries_vcx/src/errors/error.rs
+++ b/aries/aries_vcx/src/errors/error.rs
@@ -150,6 +150,9 @@ pub enum AriesVcxErrorKind {
     #[error("Could not parse a value")]
     ParsingError,
 
+    #[error("Unexpected wallet error")]
+    WalletError,
+
     // A2A
     #[error("Invalid HTTP response.")]
     InvalidHttpResponse,
--- a/aries/aries_vcx_core/src/wallet2/entry_tag.rs
+++ b/aries/aries_vcx_core/src/wallet2/entry_tag.rs
@@ -1,32 +1,9 @@
-use std::collections::HashMap;
-
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq, PartialOrd, Ord)]
 pub enum EntryTag {
     Encrypted(String, String),
     Plaintext(String, String),
-}
-
-#[cfg(feature = "vdrtools_wallet")]
-impl From<EntryTag> for (String, String) {
-    fn from(value: EntryTag) -> Self {
-        match value {
-            EntryTag::Encrypted(key, val) => (key, val),
-            EntryTag::Plaintext(key, val) => (format!("~{}", key), val),
-        }
-    }
-}
-
-#[cfg(feature = "vdrtools_wallet")]
-impl From<(String, String)> for EntryTag {
-    fn from(value: (String, String)) -> Self {
-        if value.0.starts_with('~') {
-            EntryTag::Plaintext(value.0.trim_start_matches('~').into(), value.1)
-        } else {
-            EntryTag::Encrypted(value.0, value.1)
-        }
-    }
 }
 
 #[derive(Debug, Default, Clone, PartialEq)]
@@ -35,6 +12,10 @@ pub struct EntryTags {
 }
 
 impl EntryTags {
+    pub fn new(inner: Vec<EntryTag>) -> Self {
+        Self { inner }
+    }
+
     pub fn add(&mut self, tag: EntryTag) {
         self.inner.push(tag)
     }
@@ -77,31 +58,5 @@ impl From<Vec<EntryTag>> for EntryTags {
 impl From<EntryTags> for Vec<EntryTag> {
     fn from(value: EntryTags) -> Self {
         value.inner
-    }
-}
-
-#[cfg(feature = "vdrtools_wallet")]
-impl From<EntryTags> for HashMap<String, String> {
-    fn from(value: EntryTags) -> Self {
-        let tags: Vec<EntryTag> = value.into();
-        tags.into_iter().fold(Self::new(), |mut memo, item| {
-            let (key, value) = item.into();
-
-            memo.insert(key, value);
-            memo
-        })
-    }
-}
-
-#[cfg(feature = "vdrtools_wallet")]
-impl From<HashMap<String, String>> for EntryTags {
-    fn from(value: HashMap<String, String>) -> Self {
-        Self {
-            inner: value
-                .into_iter()
-                .map(|(key, value)| (key, value))
-                .map(From::from)
-                .collect(),
-        }
     }
 }
--- a/aries/aries_vcx_core/src/wallet2/entry_tag.rs
+++ b/aries/aries_vcx_core/src/wallet2/entry_tag.rs
@@ -1,0 +1,107 @@
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum EntryTag {
+    Encrypted(String, String),
+    Plaintext(String, String),
+}
+
+#[cfg(feature = "vdrtools_wallet")]
+impl From<EntryTag> for (String, String) {
+    fn from(value: EntryTag) -> Self {
+        match value {
+            EntryTag::Encrypted(key, val) => (key, val),
+            EntryTag::Plaintext(key, val) => (format!("~{}", key), val),
+        }
+    }
+}
+
+#[cfg(feature = "vdrtools_wallet")]
+impl From<(String, String)> for EntryTag {
+    fn from(value: (String, String)) -> Self {
+        if value.0.starts_with('~') {
+            EntryTag::Plaintext(value.0.trim_start_matches('~').into(), value.1)
+        } else {
+            EntryTag::Encrypted(value.0, value.1)
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct EntryTags {
+    inner: Vec<EntryTag>,
+}
+
+impl EntryTags {
+    pub fn add(&mut self, tag: EntryTag) {
+        self.inner.push(tag)
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+}
+
+impl IntoIterator for EntryTags {
+    type Item = EntryTag;
+
+    type IntoIter = std::vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.inner.into_iter()
+    }
+}
+
+impl FromIterator<EntryTag> for EntryTags {
+    fn from_iter<T: IntoIterator<Item = EntryTag>>(iter: T) -> Self {
+        let mut tags = Self::default();
+
+        for item in iter {
+            tags.add(item);
+        }
+        tags
+    }
+}
+
+impl From<Vec<EntryTag>> for EntryTags {
+    fn from(value: Vec<EntryTag>) -> Self {
+        value.into_iter().fold(Self::default(), |mut memo, item| {
+            memo.add(item);
+            memo
+        })
+    }
+}
+
+impl From<EntryTags> for Vec<EntryTag> {
+    fn from(value: EntryTags) -> Self {
+        value.inner
+    }
+}
+
+#[cfg(feature = "vdrtools_wallet")]
+impl From<EntryTags> for HashMap<String, String> {
+    fn from(value: EntryTags) -> Self {
+        let tags: Vec<EntryTag> = value.into();
+        tags.into_iter().fold(Self::new(), |mut memo, item| {
+            let (key, value) = item.into();
+
+            memo.insert(key, value);
+            memo
+        })
+    }
+}
+
+#[cfg(feature = "vdrtools_wallet")]
+impl From<HashMap<String, String>> for EntryTags {
+    fn from(value: HashMap<String, String>) -> Self {
+        Self {
+            inner: value
+                .into_iter()
+                .map(|(key, value)| (key, value))
+                .map(From::from)
+                .collect(),
+        }
+    }
+}
--- a/aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
+++ b/aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
@@ -1,0 +1,120 @@
+use async_trait::async_trait;
+use public_key::KeyType;
+use vdrtools::{DidMethod, DidValue, KeyInfo, Locator, MyDidInfo};
+
+use crate::{
+    errors::error::{AriesVcxCoreError, AriesVcxCoreErrorKind, VcxCoreResult},
+    wallet::{indy::IndySdkWallet, structs_io::UnpackMessageOutput},
+    wallet2::{DidData, DidWallet, Key},
+};
+
+#[async_trait]
+impl DidWallet for IndySdkWallet {
+    async fn create_and_store_my_did(
+        &self,
+        seed: Option<&str>,
+        method_name: Option<&str>,
+    ) -> VcxCoreResult<DidData> {
+        let res = Locator::instance()
+            .did_controller
+            .create_and_store_my_did(
+                self.wallet_handle,
+                MyDidInfo {
+                    method_name: method_name.map(|m| DidMethod(m.into())),
+                    seed: seed.map(Into::into),
+                    ..MyDidInfo::default()
+                },
+            )
+            .await?;
+
+        Ok(DidData {
+            did: res.0,
+            verkey: Key::from_base58(&res.1, KeyType::Ed25519).map_err(|err| {
+                AriesVcxCoreError::from_msg(AriesVcxCoreErrorKind::WalletError, err)
+            })?,
+        })
+    }
+
+    async fn did_key(&self, did: &str) -> VcxCoreResult<Key> {
+        let res = Locator::instance()
+            .did_controller
+            .key_for_local_did(self.wallet_handle, DidValue(did.into()))
+            .await?;
+
+        Key::from_base58(&res, KeyType::Ed25519)
+            .map_err(|err| AriesVcxCoreError::from_msg(AriesVcxCoreErrorKind::WalletError, err))
+    }
+
+    async fn replace_did_key_start(&self, did: &str, seed: Option<&str>) -> VcxCoreResult<Key> {
+        let key_info = KeyInfo {
+            seed: seed.map(Into::into),
+            ..Default::default()
+        };
+
+        let key_string = Locator::instance()
+            .did_controller
+            .replace_keys_start(self.wallet_handle, key_info, DidValue(did.into()))
+            .await?;
+
+        let key = Key::from_base58(&key_string, KeyType::Ed25519)
+            .map_err(|err| AriesVcxCoreError::from_msg(AriesVcxCoreErrorKind::WalletError, err))?;
+
+        Ok(key)
+    }
+
+    async fn replace_did_key_apply(&self, did: &str) -> VcxCoreResult<()> {
+        Ok(Locator::instance()
+            .did_controller
+            .replace_keys_apply(self.wallet_handle, DidValue(did.into()))
+            .await?)
+    }
+
+    async fn sign(&self, key: &Key, msg: &[u8]) -> VcxCoreResult<Vec<u8>> {
+        Locator::instance()
+            .crypto_controller
+            .crypto_sign(self.wallet_handle, &key.base58(), msg)
+            .await
+            .map_err(From::from)
+    }
+
+    async fn verify(&self, key: &Key, msg: &[u8], signature: &[u8]) -> VcxCoreResult<bool> {
+        Locator::instance()
+            .crypto_controller
+            .crypto_verify(&key.base58(), msg, signature)
+            .await
+            .map_err(From::from)
+    }
+
+    async fn pack_message(
+        &self,
+        sender_vk: Option<Key>,
+        receiver_keys: Vec<Key>,
+        msg: &[u8],
+    ) -> VcxCoreResult<Vec<u8>> {
+        let receiver_keys_str = receiver_keys.into_iter().map(|key| key.base58()).collect();
+
+        Ok(Locator::instance()
+            .crypto_controller
+            .pack_msg(
+                msg.into(),
+                receiver_keys_str,
+                sender_vk.map(|key| key.base58()),
+                self.wallet_handle,
+            )
+            .await?)
+    }
+
+    async fn unpack_message(&self, msg: &[u8]) -> VcxCoreResult<UnpackMessageOutput> {
+        let unpacked_bytes = Locator::instance()
+            .crypto_controller
+            .unpack_msg(serde_json::from_slice(msg)?, self.wallet_handle)
+            .await?;
+
+        let res: UnpackMessageOutput =
+            serde_json::from_slice(&unpacked_bytes[..]).map_err(|err| {
+                AriesVcxCoreError::from_msg(AriesVcxCoreErrorKind::ParsingError, err.to_string())
+            })?;
+
+        Ok(res)
+    }
+}
--- a/aries/aries_vcx_core/src/wallet2/indy_wallet/indy_record_wallet.rs
+++ b/aries/aries_vcx_core/src/wallet2/indy_wallet/indy_record_wallet.rs
@@ -1,0 +1,138 @@
+use async_trait::async_trait;
+use indy_api_types::domain::wallet::Record as IndyRecord;
+use serde::Deserialize;
+use serde_json::Value;
+use vdrtools::Locator;
+
+use super::{SEARCH_OPTIONS, WALLET_OPTIONS};
+use crate::{
+    errors::error::{AriesVcxCoreError, VcxCoreResult},
+    wallet::indy::IndySdkWallet,
+    wallet2::{entry_tag::EntryTags, Record, RecordWallet, SearchFilter},
+};
+
+#[async_trait]
+impl RecordWallet for IndySdkWallet {
+    async fn add_record(&self, record: Record) -> VcxCoreResult<()> {
+        let tags_map = if record.tags.is_empty() {
+            None
+        } else {
+            Some(record.tags.into())
+        };
+
+        Ok(Locator::instance()
+            .non_secret_controller
+            .add_record(
+                self.wallet_handle,
+                record.category,
+                record.name,
+                record.value,
+                tags_map,
+            )
+            .await?)
+    }
+
+    async fn get_record(&self, name: &str, category: &str) -> VcxCoreResult<Record> {
+        let res = Locator::instance()
+            .non_secret_controller
+            .get_record(
+                self.wallet_handle,
+                category.into(),
+                name.into(),
+                WALLET_OPTIONS.into(),
+            )
+            .await?;
+
+        let indy_record: IndyRecord = serde_json::from_str(&res)?;
+
+        Ok(indy_record.into())
+    }
+
+    async fn update_record_tags(
+        &self,
+        name: &str,
+        category: &str,
+        new_tags: EntryTags,
+    ) -> VcxCoreResult<()> {
+        Ok(Locator::instance()
+            .non_secret_controller
+            .update_record_tags(
+                self.wallet_handle,
+                category.into(),
+                name.into(),
+                new_tags.into(),
+            )
+            .await?)
+    }
+
+    async fn update_record_value(
+        &self,
+        name: &str,
+        category: &str,
+        new_value: &str,
+    ) -> VcxCoreResult<()> {
+        Ok(Locator::instance()
+            .non_secret_controller
+            .update_record_value(
+                self.wallet_handle,
+                category.into(),
+                name.into(),
+                new_value.into(),
+            )
+            .await?)
+    }
+
+    async fn delete_record(&self, name: &str, category: &str) -> VcxCoreResult<()> {
+        Ok(Locator::instance()
+            .non_secret_controller
+            .delete_record(self.wallet_handle, category.into(), name.into())
+            .await?)
+    }
+
+    async fn search_record(
+        &self,
+        category: &str,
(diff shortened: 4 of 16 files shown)


aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+        &self,
+        seed: &str,
+        method_name: Option<&str>,
+    ) -> VcxCoreResult<DidData>;
+
+    async fn did_key(&self, name: &str) -> VcxCoreResult<String>;
+
+    async fn replace_did_key(&self, did: &str, seed: &str) -> VcxCoreResult<String>;
+
+    async fn sign(&self, key: &str, msg: &[u8], sig_type: SigType) -> VcxCoreResult<Vec<u8>>;
```
Patrik-Stas, line 111: Wouldn't key itself imply the signature type? I cannot just choose my existing key of certain type, and then choose to select arbitrary `SigType`, can i?

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    async fn verify(
+        &self,
+        key: &str,
+        msg: &[u8],
+        signature: &[u8],
+        sig_type: SigType,
```
Patrik-Stas, line 118: same as comment on `sign`

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum EntryTag {
+    Encrypted(String, String),
+    Plaintext(String, String),
+}
+
+#[derive(Debug, Default, Clone, Builder)]
+pub struct Record {
```
Patrik-Stas, line 39: Is this structure inspired based on what askar's storage model looks like?

aries/aries_vcx_core/src/wallet2/indy_wallet/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+
+const WALLET_OPTIONS: &str =
+    r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true}"#;
+
+const SEARCH_OPTIONS: &str = r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true, "retrieveRecords": true}"#;
+
+impl BaseWallet2 for IndySdkWallet {}
+
+#[allow(dead_code)]
+pub(crate) mod test_helper {
```
Patrik-Stas, line 15: You might want to place this into `test_utils` crate

Patrik-Stas: Though I am noticing this is not used in this PR - is this actually a dead code, or used by somewhere in the subsequent PRs?

It seems like this is some util code specific for the one and only mysql test suite `test_mysql_wallet`

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+
+#[async_trait]
+pub trait BaseWallet2: RecordWallet + DidWallet {}
+
+#[async_trait]
+pub trait DidWallet {
+    async fn create_and_store_my_did(
+        &self,
+        seed: &str,
+        method_name: Option<&str>,
```
Patrik-Stas, line 104: We probably ultimately don't want the method name  (as we don't want to create these wallet "dids" iin first place), but we'll get that sorted out later

aries/aries_vcx_core/src/wallet2/mod.rs
xprazak2: Removed, it is indeed not needed.

aries/aries_vcx_core/src/wallet2/mod.rs
xprazak2: Yes.
https://github.com/hyperledger/aries-askar/blob/main/askar-storage/src/entry.rs#L25-L42

aries/aries_vcx_core/src/wallet2/indy_wallet/mod.rs
xprazak2: It is used in all tests to set up wallet, though compiler sees it as unused for some reason.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
```diff
@@ -0,0 +1,145 @@
...
+use crate::{
+    errors::error::{AriesVcxCoreError, VcxCoreResult},
+    wallet2::{DidData, DidWallet},
+};
+
+#[async_trait]
+impl DidWallet for crate::wallet::indy::IndySdkWallet {
+    async fn create_and_store_my_did(
+        &self,
+        seed: &str,
```
Patrik-Stas, line 13: The existing `BaseWallet` interface has this as `seed: Option<&str>`. When `None` is specified, a random seed is generated. In fact, most of the code where `create_and_store_my_did` is being called passed `None`. 
Is there a reason why, seemingly, you've decided to strip this feature?

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
```diff
@@ -0,0 +1,145 @@
+use async_trait::async_trait;
+use vdrtools::{DidMethod, DidValue, KeyInfo, Locator, MyDidInfo};
+
+use crate::{
+    errors::error::{AriesVcxCoreError, VcxCoreResult},
+    wallet2::{DidData, DidWallet},
+};
+
+#[async_trait]
+impl DidWallet for crate::wallet::indy::IndySdkWallet {
```
Patrik-Stas, line 10: To stay consistent with most of the codebase, import `IndySdkWallet` rather than fully qualify directly in the code

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
```diff
@@ -0,0 +1,145 @@
...
+            .create_and_store_my_did(
+                self.wallet_handle,
+                MyDidInfo {
+                    method_name: method_name.map(|m| DidMethod(m.into())),
+                    seed: Some(seed.into()),
+                    ..MyDidInfo::default()
+                },
+            )
+            .await
+            .map_err::<AriesVcxCoreError, _>(From::from)?;
```
Patrik-Stas, line 27: can be just 
```
.await?;
```

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
```diff
@@ -0,0 +1,145 @@
...
+            .replace_keys_start(self.wallet_handle, key_info, DidValue(did.into()))
+            .await?;
+
+        Locator::instance()
+            .did_controller
+            .replace_keys_apply(self.wallet_handle, DidValue(did.into()))
+            .await?;
+
+        Ok(key)
+    }
```
Patrik-Stas, line 60: The higher levels of the stack require these separate. For example `rotate_verkey` in `aries/aries_vcx/src/common/keys.rs` first stars the rotation, then does some other work (tries to publish new key on ledger), and only after that applies the rotation.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_record_wallet.rs
Patrik-Stas, line 1: 👍  +1 for testing

Patrik-Stas (changes requested):
Left some comments. 

However, either way; before we merge this, I would like to see this https://github.com/hyperledger/aries-vcx/pull/1093 passing, proving that vdrtools wallet implementation of the new traits in fact satisfy the contract the rest of the codebase expects.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
xprazak2: I reverted it back, it is `Option<&str>` now.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
xprazak2: Changed.

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+                memo
+            });
+        Self {
+            id: record.name,
+            type_: record.category,
+            value: record.value,
+            tags,
+        }
+    }
+}
```
Patrik-Stas, line 77: It seems that this conversion is never used, I tried to delete it, both this and #1093 still work without it

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+#[cfg(feature = "vdrtools_wallet")]
+impl From<UnpackMessageOutput> for UnpackedMessage {
+    fn from(value: UnpackMessageOutput) -> Self {
+        Self {
+            message: value.message,
+            recipient_verkey: value.recipient_verkey,
+            sender_verkey: value.sender_verkey,
+        }
+    }
+}
```
Patrik-Stas, line 95: why not to reuse `UnpackMessageOutput`?

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
Patrik-Stas, line 1: Probably the most important comment within this review:

You placed the tests into indy implementations directly, but the tests themselves are not necessarily indy specific - they rely on the `DidWallet`, `RecordWallet` traits in general, regardless of the underlying implementations. 
The current approach would indicate that given you have test `test_indy_should_sign_and_verify`, you will likewise have `test_askar_should_sign_and_verify` where the test case would only differ by its setup.
Also, you are using `create_indy_test_wallet_handle` for all of the tests you have added, which sets up mysql wallet - that's somewhat unexpected, most of the test for sake of simplicity are using the default sqlite version of vdrtools wallet. There's only 1 particular test which specifically aims to verify that mysql works (`test_mysql_wallet.rs`).

So in my view, solving that boils down to:
- Make tests agnostic of underlying implementations - test against interfaces, not specific implementation
  - For that you can use `dev_build_featured_wallet(seed).await;` (all existing tests, except for the mysql one, are using this)
- Because `dev_build_featured_wallet` is choosing wallet implementation based on feature flags, it will be easy to use the same test cases for testing askar implementation too
- Consequently you shouldn't need `create_indy_test_wallet_handle`, which sets up mysql wallet, anymore. If it for some reason should stay, it would be nice to deduplicate it with `test_mysql_init_issuer_with_mysql_wallet`

Patrik-Stas (changes requested):

aries/aries_vcx_core/src/wallet2/mod.rs
xprazak2: Correct, not needed.

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+    async fn unpack_message(&self, msg: &[u8]) -> VcxCoreResult<UnpackedMessage>;
+}
+
+#[async_trait]
+pub trait RecordWallet {
+    async fn add_record(&self, record: Record) -> VcxCoreResult<()>;
+
+    async fn get_record(&self, name: &str, category: &str) -> VcxCoreResult<Record>;
+
+    async fn update_record(&self, record: RecordUpdate) -> VcxCoreResult<()>;
```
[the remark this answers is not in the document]
xprazak2: Changed. Askar also has tags on records. Both approaches are possible, I just generally tend to lean towards having a single method for updating attributes.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
xprazak2: Sounds good, there is one small complication though. `dev_build_featured_wallet` cannot really be used in `aries_vcx_core`, because that means adding `test_utils` crate as a dependency for `aries_vcs_core` thus creating a dependency cycle. While cargo initially does not object, having cycles in deps is generally not a good idea and problems start to surface quite rapidly with failing type checks.

I created a custom setup for now, any thoughts on this are appreciated.

Patrik-Stas (dismissed):

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+pub trait RecordWallet {
+    async fn add_record(&self, record: Record) -> VcxCoreResult<()>;
+
+    async fn get_record(&self, name: &str, category: &str) -> VcxCoreResult<Record>;
+
+    async fn update_record_tags(
+        &self,
+        name: &str,
+        category: &str,
+        new_tags: EntryTags,
```
mirgee, line 92: Can somebody please explain the concepts of name, category and tags, please? Are we using them anywhere? I assume they are specific for indy, libvdrtools and askar wallets (not for every conceivable wallet impl)?

aries/aries_vcx_core/src/wallet2/entry_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+            let (key, value) = item.into();
+
+            memo.insert(key, value);
+            memo
+        })
+    }
+}
+
+#[cfg(feature = "vdrtools_wallet")]
+impl From<HashMap<String, String>> for EntryTags {
```
mirgee, line 119: Is this conversion (i.e. the order of items in the resulting vec) deterministic?

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
mirgee: As agreed to, we will look into removing `aries_vcx_core` as `test_utils` dependency and use the setup code in `aries_vcx_core` tests directly.

aries/aries_vcx_core/src/wallet2/mod.rs
Patrik-Stas: Afaik `category + name` is askar vocabulary which maps to vdrtools concept of "type + id". Semantically the same thing, pair together representing a key in the wallet. In vdrtools terms there's example in `credx_anoncreds.rs` 
```
        wallet
            .add_wallet_record(CATEGORY_CRED_DEF, &cred_def_id.0, &str_cred_def, None)
            .await?;
```
- `category`/ `type` value is `VCX_CRED_DEF`
- `name` / `id` is some random Id
- `tags` seems to be key-value pair in `vdrtools`, never used it myself. @xprazak2 mentioned that concept of tags exists in askar too, I presume without much changes.  Given it's a key-value (and the value can be optionally encrypted, in case of sensitive data) I think one could for example store `date: 05.01.2024` key-value pair. Unless I am wrong, "tags" might be a bit of a misnomer from old times, and perhaps "attributes" would better reflect the nature of the data.

aries/aries_vcx_core/src/wallet2/utils.rs
```diff
@@ -0,0 +1,9 @@
+use rand::{distributions::Alphanumeric, Rng};
+
+pub fn random_seed() -> String {
```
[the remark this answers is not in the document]
xprazak2: Perhaps, but using it means adding `test_utils` as a dependency for `aries_vcx_core` and creating dependency cycle aries_vcx_core -> test_utils -> aries_vcx_core.

aries/aries_vcx/src/errors/error.rs
```diff
@@ -150,6 +150,9 @@ pub enum AriesVcxErrorKind {
     #[error("Could not parse a value")]
     ParsingError,
 
+    #[error("Unexpected wallet error")]
+    WalletUnexpected,
```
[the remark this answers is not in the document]
xprazak2: Changed to `WalletError` but adding a string would make the type not `Copy`. I think the whole enum would deserve a refactor as there are many variants specific to vdrtools wallet, but it would be better to do it separately.

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+pub mod indy_wallet;
+
+pub mod entry_tag;
+pub mod utils;
+
+pub struct Key {
+    pub pubkey_bs58: String,
+}
+
+#[derive(Debug, Default, Clone, Builder)]
```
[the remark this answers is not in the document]
xprazak2: Good catch.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
```diff
@@ -0,0 +1,129 @@
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum EntryTag {
```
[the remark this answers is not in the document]
xprazak2: All good questions, I am afraid I do not have answers to all of them. 
For some reason, the tags are key-value pairs in both vdrtools wallet and askar. Determining whether the tag value is encrypted is a responsibility of the wallet implementation.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn from(value: EntryTag) -> Self {
+        match value {
+            EntryTag::Encrypted(key, val) => (key, val),
+            EntryTag::Plaintext(key, val) => (format!("~{}", key), val),
+        }
+    }
+}
+
+#[cfg(feature = "vdrtools_wallet")]
+impl From<(String, String)> for EntryTag {
```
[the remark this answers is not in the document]
xprazak2: As you say, it is vdrtools specific so it should be implemented only for that particular wallet implementation, hence the feature flag.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+        if value.0.starts_with('~') {
+            EntryTag::Plaintext(value.0.trim_start_matches('~').into(), value.1)
+        } else {
+            EntryTag::Encrypted(value.0, value.1)
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct EntryTags {
```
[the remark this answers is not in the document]
xprazak2: The underlying implementation is irrelevant. I needed a type that would represent a collection of tags which would be independent on any implementation as vdrtools wallet uses `HashMap<String, String>` and askar `Vec<AskarEntryTag>`. I started with `Vec<EntryTag>`, but I can't implement `From` for that.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn add(&mut self, tag: EntryTag) {
+        self.inner.push(tag)
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+}
+
+pub struct EntryTagsIntoIterator {
```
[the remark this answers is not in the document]
xprazak2: Changed.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl From<EntryTags> for Vec<EntryTag> {
+    fn from(value: EntryTags) -> Self {
+        value.inner
+    }
+}
+
+#[cfg(feature = "vdrtools_wallet")]
+impl From<EntryTags> for HashMap<String, String> {
```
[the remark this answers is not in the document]
xprazak2: Because that is the type vdrtools uses for tags. Other implementations might use different type.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct EntryTags {
+    inner: Vec<EntryTag>,
+}
+
+impl EntryTags {
+    pub fn from_vec(inner: Vec<EntryTag>) -> Self {
```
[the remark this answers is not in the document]
xprazak2: Removed.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_record_wallet.rs
```diff
@@ -0,0 +1,138 @@
...
+
+use super::{SEARCH_OPTIONS, WALLET_OPTIONS};
+use crate::{
+    errors::error::{AriesVcxCoreError, VcxCoreResult},
+    wallet::indy::IndySdkWallet,
+    wallet2::{entry_tag::EntryTags, Record, RecordWallet, SearchFilter},
+};
+
+#[async_trait]
+impl RecordWallet for IndySdkWallet {
```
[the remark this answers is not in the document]
xprazak2: The intended use case is to manipulate wallet records. We could have all the wallet methods in a single trait if we prefer, this seems a bit more structured.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
```diff
@@ -0,0 +1,110 @@
...
+            )
+            .await?;
+
+        Ok(DidData {
+            did: res.0,
+            verkey: res.1,
+        })
+    }
+
+    async fn did_key(&self, did: &str) -> VcxCoreResult<String> {
```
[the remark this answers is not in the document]
xprazak2: Even though there might be multiple keys associated with a DID (on a data model level), `BaseWallet` currently does not expose any methods to supports it and `key_for_local_did` always returns a single item so this was modelled after it for backward compatibility.

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    #[builder(default = "EntryTags::default()")]
+    pub tags: EntryTags,
+}
+
+#[cfg(feature = "vdrtools_wallet")]
+impl From<IndyRecord> for Record {
+    fn from(ir: IndyRecord) -> Self {
+        Self {
+            name: ir.id,
+            category: ir.type_,
```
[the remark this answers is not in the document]
xprazak2: This is [an existing struct](https://github.com/hyperledger/aries-vcx/blob/main/aries/misc/legacy/libvdrtools/indy-api-types/src/domain/wallet/mod.rs#L80).

aries/aries_vcx_core/src/wallet2/indy_wallet/mod.rs
```diff
@@ -0,0 +1,12 @@
+use super::BaseWallet2;
+use crate::wallet::indy::IndySdkWallet;
+
+pub mod indy_did_wallet;
+pub mod indy_record_wallet;
+
+const WALLET_OPTIONS: &str =
+    r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true}"#;
+
+const SEARCH_OPTIONS: &str = r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true, "retrieveRecords": true}"#;
```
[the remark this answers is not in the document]
xprazak2: I see any such optimisations as out of scope for this. These are also pretty implementation-specific options.

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct DidData {
+    did: String,
+    verkey: String,
+}
+
+pub enum SearchFilter {
+    JsonFilter(String),
```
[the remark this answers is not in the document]
xprazak2: I agree that using json is unfortunate, but these changes are about introducing a new backward-compatible trait that would allow us a transition to different wallet implementation(s). These changes are not about improving the existing wallet implementation. 

Even though I could come up with a new type to use here, it would still need to be turned into json at some point because that is what vdrtools [wallet requires](https://github.com/hyperledger/aries-vcx/blob/main/aries/misc/legacy/libvdrtools/src/controllers/non_secrets.rs#L357) and that would just result in more overhead.

aries/aries_vcx_core/src/wallet2/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+
+#[async_trait]
+pub trait BaseWallet2: RecordWallet + DidWallet {}
+
+#[async_trait]
+pub trait DidWallet {
+    async fn create_and_store_my_did(
+        &self,
+        seed: Option<&str>,
+        method_name: Option<&str>,
```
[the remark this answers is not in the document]
xprazak2: See [previous comments](https://github.com/hyperledger/aries-vcx/pull/1084#discussion_r1434187492)

aries/aries_vcx_core/src/wallet2/entry_tag.rs
xprazak2: Apparently, iterator over hash map visits elements in arbitrary order and the iteration is stable if the map is not changed. Are there any requirements on tag ordering?

xprazak2:
Also, regarding multiple keys for a single DID: Looking at the `BaseWallet` methods, I do not think the trait supports that behaviour. Is that something that `BaseWallet2` should address?

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
mirgee: Tracking issue #1101 .

aries/misc/test_utils/src/devsetup.rs
```diff
@@ -49,6 +54,7 @@ use chrono::{DateTime, Duration, Utc};
 use lazy_static::lazy_static;
 use libvcx_logger::init_test_logging;
 use log::{debug, info, warn};
+use serde_json::json;
```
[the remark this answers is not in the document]
mirgee: Tracking issue #1102 .

aries/aries_vcx_core/src/wallet2/entry_tag.rs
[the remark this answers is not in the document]
mirgee: Is it possible to make it clear from the location of entry tag that it is used only by vdrtools implementation by grouping them together in a module, and then hide the entire vdrtools implementation behind a feature flag?

aries/aries_vcx_core/src/wallet2/entry_tag.rs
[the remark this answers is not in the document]
mirgee: [This](https://github.com/hyperledger/aries-vcx/pull/1084#discussion_r1445870711) comment applies here as well.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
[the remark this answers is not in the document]
mirgee: I think that those implementation-specific impls should be grouped together with the impl in order to minimize the number of places a feature flag is invoked.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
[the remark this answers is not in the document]
mirgee: [This comment](https://github.com/hyperledger/aries-vcx/pull/1084#discussion_r1445881126) applies here as well.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_record_wallet.rs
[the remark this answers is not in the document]
mirgee: I see, but the question was about what is the purpose of explicitly manipulating wallet records?

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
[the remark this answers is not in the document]
mirgee: OK, let's address this when the need arises then.

aries/aries_vcx_core/src/wallet2/mod.rs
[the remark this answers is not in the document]
mirgee: OK, it's just a nitpick anyway :)

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_did_wallet.rs
mirgee: I see it as a simpler and safer API design force the client to explicitly choose the RNG and randomness source instead of allowing multiple code paths by hiding an implementation which may unknowing change beneath their feet.

aries/aries_vcx_core/src/wallet2/entry_tag.rs
mirgee: I am not sure, but superficially seems potentially problematic.

aries/aries_vcx_core/src/wallet2/indy_wallet/indy_record_wallet.rs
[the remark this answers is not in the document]
xprazak2: Some current uses of accessing wallet records that I found:

[get_wallet_record](https://github.com/hyperledger/aries-vcx/blob/main/aries/aries_vcx_core/src/wallet/indy/wallet_non_secrets.rs#L30)
[add_wallet_record](https://github.com/hyperledger/aries-vcx/blob/main/aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs#L433)
[delete_wallet_record](https://github.com/hyperledger/aries-vcx/blob/main/aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs#L1208)
[update_wallet_record](https://github.com/hyperledger/aries-vcx/blob/main/aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs#L1317)

mirgee (approved):

Patrik-Stas (approved):

PR #1085, opened 2023-12-06
feat: implement new wallet trait for askar

xprazak2:
These changes include implementation of `BaseWallet` for askar.

These changes intentionally do not include any data migrations between wallets using `BaseWallet` or compatibility tests between different wallet implementations. Even though both of them are important, they are not needed for a new wallet implementation to work on its own, therefore they can (and will) be added separately as this is already large enough.


.github/workflows/main.yml
Patrik-Stas, line 1: we are missing running integration tests with `askar`

aries/aries_vcx_core/src/wallet/askar/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl BaseWallet for AskarWallet {}
+
+impl AskarWallet {
+    pub async fn create(
+        db_url: &str,
+        key_method: StoreKeyMethod,
+        pass_key: PassKey<'_>,
+        recreate: bool,
+        profile: Option<String>,
```
Patrik-Stas, line 40: i was looking at askar code trying to figure out what `profile` is ... what it is? Seems like some wallet-scoped multi-tenancy identifier?

xprazak2: Yeah, you can essentially have multiple "accounts" in a single wallet:

https://github.com/hyperledger/aries-askar/blob/main/docs/storage.md#profiles

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
```diff
@@ -0,0 +1,188 @@
...
+                .await?;
+
+            let verkey = local_key_to_bs58_public_key(&local_key)?;
+
+            self.insert_did(&mut tx, did, TMP_DID_CATEGORY, &verkey, None)
+                .await?;
+
+            tx.commit().await?;
+
+            return key_from_base58(&new_key_name);
```
mirgee, line 86: IMO, early exits (and thus, by extension, usage of the `return` keyword) as used here unnecessarily decrease readability (having one return statement is just simpler). Applies *many* places in this PR.

aries/aries_vcx_core/src/wallet/askar/askar_record_wallet.rs
```diff
@@ -0,0 +1,129 @@
...
+            )),
+        }
+    }
+
+    async fn delete_record(&self, category: &str, name: &str) -> VcxCoreResult<()> {
+        let mut session = self.backend.session(self.profile.clone()).await?;
+        Ok(session.remove(category, name).await?)
+    }
+
+    #[allow(unreachable_patterns)]
```
mirgee, line 100: Why is this necessary?

aries/aries_vcx_core/Cargo.toml
```diff
@@ -13,8 +13,12 @@ vdr_proxy_ledger = ["credx", "dep:indy-vdr-proxy-client"]
 # Feature flag to allow legacy proof verification
 legacy_proof = []
 
+askar_wallet = ["dep:aries-askar"]
+
 [dependencies]
 agency_client = { path = "../misc/legacy/agency_client" }
+aries-askar = { git = "https://github.com/hyperledger/aries-askar.git", tag = "v0.3.0", optional = true }
+bs58 = { version = "0.5" }
```
mirgee, line 21: Since these deps (`bs58`, `sodiumoxide`) were added as part of askar impl, they should be also optional.

aries/aries_vcx_core/src/wallet/utils.rs
```diff
@@ -0,0 +1,52 @@
...
+pub fn bytes_to_bs58(bytes: &[u8]) -> String {
+    bs58::encode(bytes).into_string()
+}
+
+pub fn from_json_str<T: for<'a> Deserialize<'a>>(json: &str) -> VcxCoreResult<T> {
+    serde_json::from_str::<T>(json)
+        .map_err(|err| AriesVcxCoreError::from_msg(AriesVcxCoreErrorKind::InvalidJson, err))
+}
+
+pub fn random_seed() -> String {
```
mirgee, line 27: Does not need to be public.

aries/aries_vcx_core/src/wallet/utils.rs
```diff
@@ -0,0 +1,52 @@
...
+        Some(val) => val.into(),
+        None => random_seed(),
+    }
+}
+
+pub fn key_from_base58(value: &str) -> VcxCoreResult<Key> {
+    Ok(Key::from_base58(value, public_key::KeyType::Ed25519)?)
+}
+
+pub fn key_from_bytes(value: Vec<u8>) -> VcxCoreResult<Key> {
```
mirgee, line 46: Where is this used?

aries/aries_vcx_core/src/wallet/utils.rs
mirgee, line 1: Most of these utils are used by the askar implementation itself, so why are they exposed unnecessarily (e.g. forcing inclusion of otherwise unnecessary dependencies)?

aries/aries_vcx_core/src/wallet/mod.rs
```diff
@@ -1,7 +1,11 @@
...
+#[cfg(feature = "askar_wallet")]
+pub mod askar;
 pub mod base_wallet;
+pub mod constants;
 pub mod entry_tags;
 #[cfg(feature = "vdrtools_wallet")]
 pub mod indy;
 pub mod mock_wallet;
 pub mod structs_io;
+pub mod utils;
```
mirgee, line 11: Does not need to be public or exist (here) at all.

aries/aries_vcx_core/src/wallet/base_wallet/search_filter.rs
```diff
@@ -1,3 +1,9 @@
+#[cfg(feature = "askar_wallet")]
+use aries_askar::entry::TagFilter;
```
mirgee, line 2: Fully qualifying the type would avoid this unnecessary feature flag.

aries/aries_vcx_core/src/wallet/indy/indy_record_wallet.rs
```diff
@@ -92,6 +92,7 @@ impl RecordWallet for IndySdkWallet {
             .await?)
     }
 
+    #[allow(unreachable_patterns)]
```
mirgee, line 95: dtto

aries/aries_vcx_core/src/wallet/askar/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+mod askar_did_wallet;
+mod askar_record_wallet;
+mod askar_tags;
+mod askar_utils;
+mod crypto_box;
+mod entry;
+mod packing;
+mod packing_types;
+mod rng_method;
+mod sig_type;
```
mirgee, line 24: 👍

mirgee (changes requested):
Just noticed a couple of things on quick inspection.

aries/aries_vcx_core/src/wallet/base_wallet/search_filter.rs
xprazak2: 👍

aries/aries_vcx_core/src/wallet/mod.rs
xprazak2: This is no longer public.

aries/aries_vcx_core/src/wallet/utils.rs
xprazak2: Removed, good catch.

aries/aries_vcx_core/Cargo.toml
xprazak2: Updated.

aries/aries_vcx_core/src/wallet/askar/askar_record_wallet.rs
xprazak2: * `-F askar_wallet,vdrtools_wallet` requires all match patterns to be covered
* `-F askar_wallet` complains about unreachable patterns since there is only one

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
```diff
@@ -0,0 +1,188 @@
...
+        Err(AriesVcxCoreError::from_msg(
+            AriesVcxCoreErrorKind::WalletRecordNotFound,
+            format!("did not found in wallet: {}", did),
+        ))
+    }
+
+    async fn replace_did_key_start(&self, did: &str, seed: Option<&str>) -> VcxCoreResult<Key> {
+        let mut tx = self.backend.transaction(self.profile.clone()).await?;
+
+        let data = self.find_current_did(&mut tx, did).await?;
```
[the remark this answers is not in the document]
xprazak2: Not using variables only once does not necessarily lead to less lines of code. Consider the following 2 variants of the same method:

```rs
async fn delete_record(&self, category: &str, name: &str) -> VcxCoreResult<()> {
        let mut session = self.backend.session(self.profile.clone()).await?;
        Ok(session.remove(category, name).await?)
}

async fn delete_record(&self, category: &str, name: &str) -> VcxCoreResult<()> {
        Ok(self
            .backend
            .session(self.profile.clone())
            .await?
            .remove(category, name)
            .await?)
}
```

This also leads to passing longer sequences of characters as arguments to method calls, so it requires arguments to be visually parsed first in order to understand what is actually passed into the method, which leads to reading code "from inside out" rather than top to bottom. If that is something you are used to, then yes, it will improve readability.

xprazak2:
I stripped the empty newlines and removed the local variables where possible if they were used only once.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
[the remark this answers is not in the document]
mirgee: Yeah, I fully admit this is a matter of personal preference.

.github/workflows/main.yml
```diff
@@ -273,11 +273,32 @@ jobs:
           default: true
           skip-docker-setup: true
       - name: "Run workspace unit tests"
-        run: RUST_TEST_THREADS=1 cargo test --workspace --lib --exclude aries-vcx-agent --exclude libvdrtools --exclude wallet_migrator --exclude mediator
+        run: RUST_TEST_THREADS=1 cargo test --workspace --lib --exclude aries-vcx-agent --exclude libvdrtools --exclude wallet_migrator --exclude mediator --exclude aries_vcx_core
+
+  test-aries_vcx_core:
```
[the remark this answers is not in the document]
xprazak2: Now using only `-`.

aries/aries_vcx_core/src/wallet/utils.rs
mirgee: You will not get warnings because a function is not used in the file it is defined in (which doesn't make sense). You get warnings for some feature flag combinations when not compiling tests because for those features this function is used only in test code, but it is not hidden under `#[cfg(test)]`. So allowing `dead_code` is not a solution because this warning is warranted.

xprazak2: You are right, this is not caused because the function is not called in the same file.

The function is called in code for askar and in tests, so for `cargo clippy -p aries_vcx --features legacy_proof --features credx,vdrtools_wallet` it is never used.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
```diff
@@ -0,0 +1,178 @@
...
+                    KeyAlg::Ed25519,
+                    seed_from_opt(seed).as_bytes(),
+                    RngMethod::RandomDet,
+                )
+                .await?;
+            let verkey = local_key_to_bs58_public_key(&local_key)?;
+            self.insert_did(&mut tx, did, TMP_DID_CATEGORY, &verkey, None)
+                .await?;
+            tx.commit().await?;
+            key_from_base58(&new_key_name)
```
Patrik-Stas, line 78: `new_key_name` is first 16 bytes of the public key, so this seems wrong - the function seem to expect key, but gets "key name" instead
Before fixing this, it would be worthy having a look why no test catched this.

aries/aries_vcx_core/src/wallet/askar/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+    ) -> VcxCoreResult<Option<DidData>> {
+        self.find_did(session, did, DID_CATEGORY).await
+    }
+
+    async fn insert_did(
+        &self,
+        session: &mut Session,
+        did: &str,
+        category: &str,
+        verkey: &str,
```
Patrik-Stas, line 129: The callsites (albeit these are just internal) are required to know, that verkey is base58 encoded. Perhaps we could name this more explicitly `create_did_from_local_key` and accept `local_key: LocalKey`  instead of `verkey: &str`.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
```diff
@@ -0,0 +1,178 @@
...
+                AriesVcxCoreErrorKind::WalletRecordNotFound,
+                format!("did not found in wallet: {}", did),
+            ))
+        }
+    }
+
+    async fn replace_did_key_apply(&self, did: &str) -> VcxCoreResult<()> {
+        let mut tx = self.backend.transaction(self.profile.clone()).await?;
+        if let Some(did_data) = self.find_did(&mut tx, did, TMP_DID_CATEGORY).await? {
+            let verkey_did = did_data.did_from_verkey();
```
Patrik-Stas, line 90: At this point, `did` must equal to the value of the variable `verkey_did`, do you agree? If it's not, something must be had been done prior to this. This is contract expected to hold - it points me more fundamentally to the fact that there seem to be redundancy in how the data is being stored - it seem that it is pair of :
```
did  ----> DidData { did, verkey }
```
which is what brings us on this line into "dilemma", what's the source of truth of what `did` we work with at this point, and we have 3 options in fact:
- arg `did`
- `did_data.did`
-  derive the did from `did_data.verkey` (which is what is being done now)

When we pass in our temporary did as argument to this function, we only need it to resolve the full value of verkey. That should be the only purpose of the the `find_did` lookup here.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
```diff
@@ -0,0 +1,178 @@
...
+            tx.remove(TMP_DID_CATEGORY, did).await?;
+            tx.remove_key(&verkey_did).await?;
```
Patrik-Stas, line 92: I don't know so honest question: Is the removal of these records + subsequent update on line `93` fully transactional?

If this sequence of updates is atomic, that's ideal. if not, it would be slightly less worse to run removal after successfull `update_did` on line 83. That way, if update fails, it's still possible to rerun `replace_did_key_apply`.

Patrik-Stas (changes requested):
I have not went through everything yet, but leaving feedback in the meantime.

aries/aries_vcx_core/Cargo.toml
```diff
@@ -13,8 +13,12 @@ vdr_proxy_ledger = ["credx", "dep:indy-vdr-proxy-client"]
 # Feature flag to allow legacy proof verification
 legacy_proof = []
 
+askar_wallet = ["dep:aries-askar", "dep:bs58", "dep:sodiumoxide"]
+
 [dependencies]
 agency_client = { path = "../misc/legacy/agency_client" }
+aries-askar = { git = "https://github.com/hyperledger/aries-askar.git", tag = "v0.3.0", optional = true }
```
gmulhearn-anonyome, line 20: is there a reason for using git instead of the crates io version? also note crates.io is up to 0.3.1

https://crates.io/crates/aries-askar

gmulhearn-anonyome (commented):
had a skim thru. nice stuff!

Just 1 general query, did you consider not using `sodiumoxide` for the sealedbox crypto? AFAIK Askar should have all of the sealedbox functionality we need. e.g.; https://github.com/hyperledger/aries-askar/blob/main/src/kms/envelope.rs

a little hesistant on `sodiumoxide` just because it doesn't seem super maintained, and requires the libsodium binary be packed/linked in. I believe askar's implementation is using a pure rust impl of crypto box (`crypto_box_rs`/`crypto_box`) https://crates.io/crates/crypto_box

anyway, something to consider. it could reduce complexity

gmulhearn-anonyome:
had a skim thru. nice stuff!

Just 1 general query, did you consider not using sodiumoxide for the sealedbox crypto? AFAIK Askar should have all of the sealedbox functionality we need. e.g.; https://github.com/hyperledger/aries-askar/blob/main/src/kms/envelope.rs

a little hesistant on sodiumoxide just because it doesn't seem super maintained, and requires the libsodium binary be packed/linked in. I believe askar's implementation is using a pure rust impl of crypto box (crypto_box_rs/crypto_box) https://crates.io/crates/crypto_box

anyway, something to consider. it could reduce complexity

aries/aries_vcx_core/src/wallet/askar/crypto_box.rs
```diff
@@ -0,0 +1,173 @@
+use sodiumoxide::crypto::{
```
gmulhearn-anonyome, line 1: Just 1 general query, did you consider not using `sodiumoxide` for the sealedbox crypto? AFAIK Askar should have all of the sealedbox functionality we need. e.g.; https://github.com/hyperledger/aries-askar/blob/main/src/kms/envelope.rs

a little hesistant on `sodiumoxide` just because it doesn't seem super maintained, and requires the libsodium binary be packed/linked in. I believe askar's implementation is using a pure rust impl of crypto box (`crypto_box_rs`/`crypto_box`) https://crates.io/crates/crypto_box

anyway, something to consider. it could reduce complexity

aries/aries_vcx_core/src/wallet/askar/askar_utils.rs
```diff
@@ -0,0 +1,72 @@
...
+    Ok(bs58::encode(local_key.to_public_bytes()?).into_string())
+}
+
+pub fn local_key_to_private_key_bytes(local_key: &LocalKey) -> VcxCoreResult<Vec<u8>> {
+    Ok(local_key.to_secret_bytes()?.to_vec())
+}
+
+pub fn local_key_to_public_key_bytes(local_key: &LocalKey) -> VcxCoreResult<Vec<u8>> {
+    Ok(local_key.to_public_bytes()?.to_vec())
+}
```
Patrik-Stas, line 25: These functions don't really need to exist - both of these are converting `SecretBytes` to `Vec<u8>` but then in practice, Vec<u8> is used as `&[u8]` slices.
And askar's `SecretBytes` is basically kind of smart pointer, it implements `Deref` for `[u8]` target. So for example in `pack_anoncrypt_recipients`  instad of :
```
let enc_key_secret = local_key_to_private_key_bytes(enc_key)?;
self.crypto_box.sealedbox_encrypt(&recipient_public_bytes, &enc_key_secret)?;
```

you can just directly get slice of secret bytes from `enc_key`
```
self.crypto_box.sealedbox_encrypt(&recipient_public_bytes, &enc_key.to_secret_bytes()?)?;
```

aries/aries_vcx_core/src/wallet/askar/packing.rs
```diff
@@ -0,0 +1,303 @@
...
+            &recipient.header.sender.decode()?,
+        )?;
+        Ok((
+            LocalKey::from_secret_bytes(
+                KeyAlg::Chacha20(Chacha20Types::C20P),
+                &self.crypto_box.box_decrypt(
+                    &private_bytes,
+                    &ed25519_to_x25519_public(&LocalKey::from_public_bytes(
+                        Ed25519,
+                        &bs58_to_bytes(&bytes_to_string(sender_vk_vec.clone())?)?,
```
Patrik-Stas, line 106: seem like unnecessary conversion doing `bytes_to_string`. The `bs58_to_bytes` can be changed to accept argument `key: &[u8]` (which is what it needs internally), rather than its current `key: &str`.

aries/aries_vcx_core/src/wallet/askar/askar_utils.rs
```diff
@@ -0,0 +1,72 @@
...
+pub fn from_json_str<T: for<'a> Deserialize<'a>>(json: &str) -> VcxCoreResult<T> {
+    serde_json::from_str::<T>(json)
+        .map_err(|err| AriesVcxCoreError::from_msg(AriesVcxCoreErrorKind::InvalidJson, err))
+}
+
+pub fn bytes_to_bs58(bytes: &[u8]) -> String {
+    bs58::encode(bytes).into_string()
+}
+
+pub fn bs58_to_bytes(key: &str) -> VcxCoreResult<Vec<u8>> {
```
Patrik-Stas, line 63: as mentioned elsewhere, would be better to use `key: &[u8]`

aries/aries_vcx_core/src/wallet/askar/packing.rs
```diff
@@ -0,0 +1,303 @@
...
+        Ok(UnpackMessageOutput {
+            message: self.unpack_msg(&jwe, enc_key)?,
+            recipient_verkey: recipient.unwrap_kid().to_owned(),
+            sender_verkey: sender_verkey.map(|key| key.base58()),
+        })
+    }
+
+    pub fn unpack_protected_data(&self, jwe: &Jwe) -> VcxCoreResult<ProtectedData> {
+        from_json_str(&jwe.protected.decode_to_string()?)
+    }
```
Patrik-Stas, line 56: this is used on single place `unpack_protected_data(..)`, how about inlining it?

aries/aries_vcx_core/src/wallet/askar/packing.rs
```diff
@@ -0,0 +1,303 @@
...
+        let local_key = key_entry.load_local_key()?;
+        let (enc_key, sender_verkey) = self.unpack_recipient(recipient, &local_key)?;
+        Ok(UnpackMessageOutput {
+            message: self.unpack_msg(&jwe, enc_key)?,
+            recipient_verkey: recipient.unwrap_kid().to_owned(),
+            sender_verkey: sender_verkey.map(|key| key.base58()),
+        })
+    }
+
+    pub fn unpack_protected_data(&self, jwe: &Jwe) -> VcxCoreResult<ProtectedData> {
```
Patrik-Stas, line 54: Like other functions in `Packing`, this is also taking, but not actually using `&self`

aries/aries_vcx_core/src/wallet/askar/packing.rs
Patrik-Stas, line 1: I have number of comments on the `Packing` struct, holistically indicating for me that `Packing` struct should not really exist. Functions are taking `&self` but not using it, the struct has state which is not meant to be reused across function calls.

Patrik-Stas (changes requested):
added more comments

.github/workflows/main.yml
```diff
@@ -273,11 +273,32 @@ jobs:
           default: true
           skip-docker-setup: true
       - name: "Run workspace unit tests"
-        run: RUST_TEST_THREADS=1 cargo test --workspace --lib --exclude aries-vcx-agent --exclude libvdrtools --exclude wallet_migrator --exclude mediator
+        run: RUST_TEST_THREADS=1 cargo test --workspace --lib --exclude aries-vcx-agent --exclude libvdrtools --exclude wallet_migrator --exclude mediator --exclude aries_vcx_core
+
+  test-aries-vcx-core:
```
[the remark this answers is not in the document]
mirgee: It's debatable whether these are unit tests or integration tests, though :)

aries/aries_vcx_core/src/wallet/askar/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+mod entry;
+mod packing;
+mod packing_types;
+mod rng_method;
+mod sig_type;
+
+#[derive(Debug)]
+pub struct AskarWallet {
+    backend: Store,
+    profile: Option<String>,
```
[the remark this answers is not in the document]
mirgee: FYI we are using the same approach of hardcoding an identifier for the link secret as well.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
```diff
@@ -0,0 +1,178 @@
...
+        if let Some(key) = self
+            .backend
+            .session(self.profile.clone())
+            .await?
+            .fetch_key(&did_from_key(key.to_owned()), false)
+            .await?
+        {
+            let local_key = key.load_local_key()?;
+            let key_alg = SigType::try_from_key_alg(local_key.algorithm())?;
+            Ok(local_key.sign_message(msg, Some(key_alg.into()))?)
```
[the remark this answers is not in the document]
xprazak2: From what I know, `KeyEntry` is an item fetched from database while `LocalKey` gives you access to the key content - certs data and crypto operations.

aries/aries_vcx_core/src/wallet/askar/mod.rs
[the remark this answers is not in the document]
xprazak2: I changed profile to `String`, which now forces the code consumer to always specify a profile.

From what I understood, `None` represents a default profile and it does not prevent us from reusing a wallet that was previously closed. If and how the wallet multitenancy is used is up to the consumer (aries_vcx in our case), so if we decide to use a single hardcoded profile identifier, we definitely can.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
```diff
@@ -0,0 +1,178 @@
...
+                &mut tx,
+                KeyAlg::Ed25519,
+                seed_from_opt(seed).as_bytes(),
+                RngMethod::RandomDet,
+            )
+            .await?;
+        let verkey = local_key_to_bs58_public_key(&local_key)?;
+        self.insert_did(&mut tx, &did, DID_CATEGORY, &verkey, None)
+            .await?;
+        tx.commit().await?;
```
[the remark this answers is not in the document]
xprazak2: There is no guidance that I found, however when using `session`, the changes are persisted without calling `commit` while `transaction` requires it to be called for any changes to apply.

I added a test to make creating did and key works atomically.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
```diff
@@ -0,0 +1,178 @@
...
+        msg: &[u8],
+    ) -> VcxCoreResult<Vec<u8>> {
+        if recipient_keys.is_empty() {
+            Err(AriesVcxCoreError::from_msg(
+                AriesVcxCoreErrorKind::InvalidInput,
+                "recipient keys should not be empty",
+            ))
+        } else {
+            let enc_key = LocalKey::generate(KeyAlg::Chacha20(Chacha20Types::C20P), true)?;
+            let packing = Packing::new();
```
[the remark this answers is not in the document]
xprazak2: Changed, why do you think reusing a `Packing` struct might result in an undefined behaviour? Do you have a reproducer?

The only state that `Packing` has is a crypto box, which is itself stateless. If there is a problem with reusing the `SodiumCryptoBox` then it means the struct is probably not stateless (assuming the underlying crypto functions work correctly), which does not seem to be the case.

I am not sure if adding our specific functionality to a foreign type (`LocalKey`) in this case is in any measurable way better that having a dedicated struct we own, but that is just a matter of preference I guess.

Main reason for having the `Packing` struct was to have a single source of truth for the implementation of `crypto_box`, so swapping the implementation in the future would require only a single line change. It would also make it impossible to accidentally pack with one implementation and unpack with another. Now you need to hunt down all the occurrences of `SodiumCryptoBox::new` by search and replace in 2 files.

aries/aries_vcx_core/src/wallet/askar/packing.rs
```diff
@@ -0,0 +1,303 @@
...
+                    &private_bytes,
+                    &public_bytes,
+                    &recipient.encrypted_key.decode()?,
+                )?,
+            )?,
+            None,
+        ))
+    }
+
+    async fn find_recipient_key<'a>(
```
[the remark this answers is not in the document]
xprazak2: Solved by moving packing to a trait.

aries/aries_vcx_core/src/wallet/askar/packing.rs
```diff
@@ -0,0 +1,303 @@
...
+            };
+        }
+
+        Err(AriesVcxCoreError::from_msg(
+            AriesVcxCoreErrorKind::WalletRecordNotFound,
+            "recipient key not found in wallet",
+        ))
+    }
+
+    pub fn pack_all(
```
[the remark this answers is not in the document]
xprazak2: Solved by moving packing to a trait.

aries/aries_vcx_core/src/wallet/askar/mod.rs
xprazak2: Changed to `public_key::Key` as a part of fix for key name above.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
xprazak2: > At this point, did must equal to the value of the variable verkey_did, do you agree?
Yes.

I removed `did_data.verkey` as it is not needed. It seems like the did itself does not need to be stored as a part of the record value, it is just something the existing wallet does - removed as well.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
xprazak2: Yes, this is supposed to be transactional. It might be useful to add `delete_did` into `BaseWallet` so that we could test...

aries/aries_vcx_core/src/wallet/askar/askar_utils.rs
xprazak2: Changed.

aries/aries_vcx_core/src/wallet/askar/askar_did_wallet.rs
xprazak2: This was rolled back. `did_data.verkey` for the did/key names to correctly connect when rotating key more than once.

Patrik-Stas (approved):
Looks good to me, well done! ✅  Let's get this baby merged. Just need to sort out that `napi` image

mirgee (approved):

Patrik-Stas (approved):

Issue #1080, opened 2023-12-03
Mediated connection - empty string for DIDDoc's publickey controller in the Invitee Request message


mkempa:
Hi,
I've found out that the mediated connection request has an empty string as a controller of the DIDDoc's public key.

The similar bug has been reported and already resolved - #877 

The bug appears to originate in the `protocols/mediated_connection/invitee/state_machine.rs`, function `build_connection_request_msg` on lines
```
let mut did_doc = AriesDidDoc::default();
did_doc.set_service_endpoint(service_endpoint);
did_doc.set_routing_keys(routing_keys);
did_doc.set_recipient_keys(recipient_keys);
did_doc.set_id(self.pairwise_info.pw_did.clone());
```
which look very much like the ones in the aforementioned bug report.

I would very much appreciate if you had a look into it. I would open a PR myself but I have troubles building the lib for the android so I haven't tested the possible fix.
Thanks in advance.

Patrik-Stas:
Hi @mkempa apologies for late feedback. Indeed, it looks like it should be straightforward, but addressing this is currently beyond our bandwidth capacity. 
The best course of actions I advise for you is try to reproduce the issue locally on Rust level, rather than dealing with it through extra android layer.

Also, please beware that mediated connection along with [vcxagency-node](https://github.com/AbsaOSS/vcxagencynode) are deprecated. To give you some additional context where we are, where we heading:
- The mediated connection - `MediatedConnection` - abstraction has been deprecated early 2023
- Instead we created non-mediated version of connection - `Connection<I, S>`. Using this, there can be still mediation, but it must be dealt with externally by the `aries_vcx` consumer.
- One of the reason why `MediatedConnection` is deprecated is because it relies on non-standard protocol to talk to the mediator itself. 
- In 2024, we have created new mediator project (as part of mentorship project, by our mentee) https://github.com/hyperledger/aries-vcx/tree/main/aries/agents/rust/mediator - this follow standard mediation protocols
   - There's no "client" component to it yet, which is missing piece. Technically there's kind of client which is used tests, but practically there's outstanding work to make that a user friendly black box component.

Patrik-Stas:
The mediated connection has been decomissioned and is no longer available.

PR #1087, opened 2023-12-11
fix: Mediated Connection - Decrypt invitation request at Invitee's as No-Auth

mkempa:
This PR fixes an issue when Invitee is unable to decrypt connection request downloaded from mediator. The request was being attempted to decrypt as Auth when in fact it should still be No-Auth.


Patrik-Stas:
Hi @mkempa much appreciated first contribution! 

A small technical comment - Perhaps you've run into some edge case, things worked in past without this change, but I think this does no harm. - CI is passing and the Mediated Connection is generally [not used](https://github.com/hyperledger/aries-vcx/issues/1080#issuecomment-1852040569) anymore (though Mediated Connection is still getting tested indirectly in our NodeJS wrapper tests).

Patrik-Stas (approved):

PR #1092, opened 2023-12-18
feat: derive x25519 keys from Ed25519

xprazak2:


xprazak2:
Included in #1085, closing.

PR #1093, opened 2023-12-20
feat: replace BaseWallet with BaseWallet2

xprazak2:
* replaces `BaseWallet` with `BaseWallet2` in all crates.
* adds implementation of `BaseWallet2` for `AgencyClientWallet` and `MockWallet`


Patrik-Stas (approved):

aries/agents/rust/aries-vcx-agent/src/helper.rs
```diff
@@ -3,7 +3,7 @@ use aries_vcx::{
     messages::AriesMessage,
     utils::{encryption_envelope::EncryptionEnvelope, from_did_doc_sov_to_legacy},
 };
-use aries_vcx_core::wallet::base_wallet::BaseWallet;
+use aries_vcx_core::wallet2::BaseWallet2;
```
mirgee, line 6: Can the original `BaseWallet` trait be removed? If so, why not remove it and rename `BaseWallet2` to `BaseWallet`, and if not, why not?

xprazak2: Renamed.

aries/aries_vcx_core/src/wallet/base_wallet.rs
```diff
@@ -1,124 +1,440 @@
...
+pub trait BaseWallet: RecordWallet + DidWallet + Send + Sync + std::fmt::Debug {}
+
+#[async_trait]
+pub trait DidWallet {
     async fn create_and_store_my_did(
         &self,
         seed: Option<&str>,
-        kdf_method_name: Option<&str>,
-    ) -> VcxCoreResult<(String, String)>;
+        method_name: Option<&str>,
```
Patrik-Stas, line 96: If the type doesn't reflect what the thing is, at least name should. I am in favor of the original name here

aries/aries_vcx_core/src/wallet/base_wallet.rs
```diff
@@ -1,124 +1,440 @@
...
+    ) -> VcxCoreResult<DidData>;
 
-    async fn key_for_local_did(&self, did: &str) -> VcxCoreResult<String>;
+    async fn did_key(&self, name: &str) -> VcxCoreResult<Key>;
```
Patrik-Stas, line 99: Yeah indeed ultimately these "local wallet dids" are just somewhat of an aliases for keys. The terminology is mixed bag and in future want to basically remove this "did alias" feature, but for now still it is what it is. I'd keep the original method name, or perhaps a shorter variation `key_for_did`. Given the function name, `name` argument name would also be better off named `did` to avoid discrepancy in vocabulary.

aries/aries_vcx/tests/utils/test_agent.rs
```diff
@@ -103,15 +103,14 @@ where
     W: BaseWallet,
 {
     let acme = create_test_agent(genesis_file_path.to_string()).await;
-    let acme_vk =
-        get_verkey_from_wallet(acme.wallet.get_wallet_handle(), &acme.institution_did).await?;
+    let acme_vk = acme.wallet.did_key(&acme.institution_did).await?;
```
Patrik-Stas, line 106: The original `get_verkey_from_wallet` seem to be dead code now, can be deleted. 

Along with that I've also noticed `libindy_replace_keys_start`, `libindy_replace_keys_apply` to be dead code now. 
We should review why clippy is not catching this, but let's look into that separately elsewhere. There might be possibly more dead code across the codebase we are unaware of.

aries/aries_vcx/tests/utils/test_agent.rs
```diff
@@ -103,15 +103,14 @@ where
...
 
     write_endorser_did(
         &trustee_wallet,
         &ledger_write,
         trustee_did,
         &acme.institution_did,
-        &acme_vk,
+        &acme_vk.base58(),
```
Patrik-Stas, line 113: Improvement but generally not a huge fan of the current state either. This comment is not actionable, just noting for awareness or to spark any discussion. But this shouldn't block this PR.

- Arguably `write_endorser_did` would also take a typed key and there wouldn't be to run the encoding at this spot, just passing around raw data at higher levels of code (like here)
- Arguably we would have types for specific keys, eg. currently there's silent contract that the `target_vk: &str,` argument of `write_endorser_did` in expected ed25519 public key. The `Key` is very general type, as it can represent any key. So ideally, there would be type such as `PublicKeyEd25519` which could be used here.
- All that being said, I would be reluctant to do relevant changes across the codebase just yet - I suppose `askar` might be coming with its own key types and for me it's yet open question whether we'd use those, or abstract them out with custom type.

aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -60,131 +59,47 @@ pub struct RevocationRegistryInfo {
     pub used_ids: HashSet<u32>,
 }
 
-/// Adapter used so that credx does not depend strictly on the vdrtools-wallet
-/// Will get removed when the wallet and anoncreds interfaces are de-coupled.
```
Patrik-Stas, line 64: I suppose this might be still relevant comment? The "adapter" thingie seems like an overhead.

aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -60,131 +59,47 @@ pub struct RevocationRegistryInfo {
...
     }
 
-    async fn update_wallet_record_tags(
+    async fn update_record_tags(
         &self,
-        xtype: &str,
-        id: &str,
-        tags: HashMap<String, String>,
+        name: &str,
+        category: &str,
```
Patrik-Stas, line 78: reviewer's note: the order has been swapped
- `name` is equivalent to former `id`
- `category` is equivalent to former `xtype`

Is there rationale for the changed order? I have actually slight preference for the original order (and would reduce PR delta) but it's not a biggie. Personal preference, feel free to mark this resolved. (or if you would be changing this, note that InteliJ has great refactoring feature for changing function arguments order)

aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -200,23 +115,19 @@ impl IndyCredxAnonCreds {
     where
         T: DeserializeOwned,
     {
-        let str_record = wallet.get_wallet_record_value(category, id).await?;
-        serde_json::from_str(&str_record).map_err(From::from)
+        let str_record = wallet.get_record(id, category).await?;
+        serde_json::from_str(str_record.get_value()).map_err(From::from)
```
Patrik-Stas, line 119: Should been noted in previous PR, but I just noticed it now. 

Looking at `did_doc` and `public_key` crates as reference, we tend to not use `get_` prefix on getter functions. Eg. example from `public_key`
```
pub fn key_type(&self) -> &KeyType {
        &self.key_type
    }

    pub fn key(&self) -> &[u8] {
        self.key.as_ref()
    }

```
example from `did_doc`
```
pub fn controller(&self) -> Option<&OneOrList<Did>> {
        self.controller.as_ref()
    }

    pub fn verification_method(&self) -> &[VerificationMethod] {
        self.verification_method.as_ref()
    }
```

I am actually fine keeping get on the BaseWallet method, because those are rather IO operations than just plain in memory "getters". But I would suggest `value` rather than `get_value` for the `Record` type (and analogically for its other getters)

aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -251,24 +158,20 @@ impl IndyCredxAnonCreds {
...
-        let mut record_iterator = wallet
-            .iterate_wallet_records(CATEGORY_CREDENTIAL, wql, "{}")
+        let records = wallet
+            .search_record(
+                CATEGORY_CREDENTIAL,
+                Some(SearchFilter::JsonFilter(wql.into())),
+            )
             .await?;
-        let records = record_iterator.collect().await?;
+        // let records = record_iterator.collect().await?;
```
Patrik-Stas, line 167: forgotten comment

aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -235,14 +146,10 @@ impl IndyCredxAnonCreds {
...
-            .get_wallet_record(CATEGORY_CREDENTIAL, credential_id, "{}")
+            .get_record(credential_id, CATEGORY_CREDENTIAL)
             .await?;
-        let cred_record: Value = serde_json::from_str(&cred_record)?;
-        let cred_record_value = (&cred_record).try_get("value")?;
 
-        let cred_json = cred_record_value.try_as_str()?;
-
-        let credential: CredxCredential = serde_json::from_str(cred_json)?;
+        let credential: CredxCredential = serde_json::from_str(cred_record.get_value())?;
```
Patrik-Stas, line 152: This reads so much nicer 👍

aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -428,38 +331,36 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
...
-
-        wallet
-            .add_wallet_record(
-                CATEGORY_REV_REG_INFO,
-                &rev_reg_id.0,
-                &str_rev_reg_info,
-                None,
-            )
-            .await?;
+        let record = RecordBuilder::default()
```
Patrik-Stas, line 334: `Record` is using `derive_builder` - we would far more benefit here using `typed-builder` (used in the crate `messages`) as it gives you compile time guarantees you are using it right and it doesn't every return `Err` result. Obivously you don't get runtime flexibility, but I don't that's a property of the the `RecordBuilder` being leveraged

Switching from `derive_builder` to `typed-builder` should be possible with minimum changes, as they produce similar builder APIs.

aries/aries_vcx_core/src/anoncreds/credx_anoncreds.rs
```diff
@@ -510,50 +411,47 @@ impl BaseAnonCreds for IndyCredxAnonCreds {
             )?;
 
         let str_cred_def = serde_json::to_string(&cred_def)?;
-
+        let record = RecordBuilder::default()
+            .name(cred_def_id.0.clone())
+            .category(CATEGORY_CRED_DEF.to_string())
+            .value(str_cred_def.clone())
+            .build()?;
         // Store stuff in wallet
```
Patrik-Stas, line 419: while we are at it, let's delete this comment ^_^

aries/aries_vcx_core/src/wallet/agency_client_wallet.rs
```diff
@@ -1,166 +1,132 @@
...
-        Err(unimplemented_agency_client_wallet_method(
-            "get_verkey_from_wallet",
-        ))
-    }
-
-    async fn replace_did_keys_start(&self, target_did: &str) -> VcxCoreResult<String> {
-        Err(unimplemented_agency_client_wallet_method(
-            "replace_did_keys_start",
-        ))
+impl RecordWallet for AgencyClientWallet {
```
Patrik-Stas, line 28: Non actionable here

Just wondering how difficult it would be to remove this. I have feeling that pretty easy now, that `AgencyClientWallet` was originally conceived to avoid cyclical dependency between `agency_client` and `aries_vcx`, but that would not longer be the case given the wallet stuff was take out to `aries_vcx_core`.

aries/aries_vcx_core/src/wallet/base_wallet.rs
```diff
@@ -1,124 +1,440 @@
-use std::collections::HashMap;
-
 use async_trait::async_trait;
+use derive_builder::Builder;
```
Patrik-Stas, line 2: Ah, I see that `derive_builder` was in fact added in this PR. I've left a comment about it somewhere above.

aries/misc/legacy/libvcx_core/Cargo.toml
```diff
@@ -26,6 +26,8 @@ futures = { version = "0.3", default-features = false }
 aries_vcx = { path = "../../../aries_vcx", features = ["credx", "vdrtools_wallet"] }
 aries_vcx_core = { path = "../../../aries_vcx_core" }
 diddoc_legacy = { path = "../diddoc_legacy" }
+public_key = { path = "../../../../did_core/public_key" }
+indy-api-types = { path = "../libvdrtools/indy-api-types" }
```
Patrik-Stas, line 30: Is this really being used? Can't see the usage

Patrik-Stas (changes requested):
Left some comment to yet address, nevertheless good improvement!

mirgee (changes requested):

xprazak2:
main changes:
  * using TypedBuilder
  * remove `get_` from attribute getters
  * introduce wrappers for types returned by Indy wallet
  * implement conversions between types without `From` and `TryFrom`
  * swap args in wallet methods so that `category` is first

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/wallet.rs
```diff
@@ -208,21 +248,45 @@ pub async fn wallet_delete_wallet_record_tags(
...
+    _options: &str,
 ) -> LibvcxResult<String> {
     let wallet = get_main_wallet()?;
-    map_ariesvcx_core_result(wallet.get_wallet_record(xtype, id, options).await)
+
+    map_ariesvcx_result(
+        wallet
+            .get_record(id, xtype)
+            .map(|item| {
+                let wallet_record: WalletRecord = item?.try_into()?;
```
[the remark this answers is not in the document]
xprazak2: Nope, that does not seem to work as the error types do not align.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/wallet.rs
```diff
@@ -208,21 +259,44 @@ pub async fn wallet_delete_wallet_record_tags(
...
+    map_ariesvcx_result(
+        wallet
+            .get_record(xtype, id)
+            .map(|res| {
+                let wallet_record = WalletRecord::from_record(res?)?;
+
+                Ok(serde_json::to_string(&wallet_record)?)
+            })
+            .await,
+    )
```
mirgee, line 294: Why doesn't `WalletRecord` implement `Serialize`? This entire function can be a single line:
```rust
serde_json::to_string(&get_main_wallet()?.get_record(id, xtype).await?).map_err(Into::into)
```

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -133,6 +81,23 @@ pub struct WalletRecord {
     tags: Option<String>,
 }
 
+impl WalletRecord {
```
mirgee, line 84: In the codebase, there is

* `Record` struct
* `WalletRecord` trait
* `RecordWallet` trait
* `IndyRecord` struct which is a name for `RecordStruct` in `indy_api_types`

now with all kinds of conversions between them. This is quite confusing if you ask me as a reader. If there is no way to eradicate some of those, we should at least find better names.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/wallet.rs
xprazak2: Because`Record` would need to implement `Serialize` exactly the same way as `WalletRecord`.

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: I could remove `RecordWallet` by squashing it together with `DidWallet` into `BaseWallet` if you prefer. 

`Record` was recently introduced to `aries_vcx_core` as an effort to improve the types, so I think this one will stay. 

As for the others, I do not see them going away as long as we keep the code that is using them. `WalletRecord` is returned as a json string from methods in `api_global`, so that is effectively a part of public API exposed by that crate. `Record` from `indy_api_types` is used in `libvidrtools`, I imagine this one will go away as we continue to refactor.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/wallet.rs
mirgee: Sorry, I missed the point of the function. It (and others in this module) can be simplified regardless, although no point, this is code to be removed as per location. `Record` could definitely implement `Serialize`, but also no need.

aries/aries_vcx_core/src/wallet/indy/mod.rs
mirgee: Nah, I would not squash them. Considering alternative names for `RecordWallet`, e.g. `CRUDWallet`, `DataWallet` comes to mind, but none seems preferable.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/wallet.rs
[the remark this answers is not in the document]
mirgee: Yeah, but error types can be mapped. Nevertheless, resolving, we don't want to invest excessively into the legacy components

aries/aries_vcx_core/src/wallet/indy/mod.rs
Patrik-Stas: > Record struct
> WalletRecord trait
> RecordWallet trait
> IndyRecord struct which is a name for RecordStruct in indy_api_types

`IndyRecord` is not actually type/struct defined, but only used as alias
```
use indy_api_types::domain::wallet::Record as IndyRecord;
```

Perhaps we wouldn't need to use alias if we have better names for the remaining 3:


> Record struct 
> WalletRecord trait
> RecordWallet trait

- `WalletRecord` is not trait but struct, but  there's actually 2 of them:
  - one defined in `wallet/indy/mod.rs`
  - another one defined in `aries/misc/legacy/libvdrtools/indy-wallet/src/lib.rs`
I would keep the one inside `libvdrtools` as it is, it's local to libvdrtools. But we could rename `wallet/indy/mod.rs`, as that's seems to be something we defined, something that we made indy specific. It's usage seems to be narrowly limited in `libvcx` specifically for method `wallet_get_wallet_record`. If we don't want to remove it altogether in this PR to avoid making unnecessary changes, we could rename this to `IndyWalletRecord`. Ultimately remove it altogether later. 

- `Record` 
there's also 2 of this, one our ours, and one in `libvdrtools` - which is why we then consequently do
```
use indy_api_types::domain::wallet::Record as IndyRecord;
```
We could rename the `Record` inside `libvdrtools` to:
1. avoid this aliasing
2. avoid conflicts and being explicit about the fact it's coupled with indy/vdrtools.

- `RecordWallet` trait can stay IMO as is, it fits nice with the other trait name -`DidWallet`

mirgee (changes requested):

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: Rename sounds ok to me, I think some of these structs will be dropped in time as we continue to refactor.

aries/aries_vcx_core/src/wallet/mock_wallet.rs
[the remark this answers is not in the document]
xprazak2: It will stay for the time being. This is an existing file, any modifications here are only because of changes to `BaseWallet`.

I agree that we should remove testing-specific code from release builds, but that is out of scope for this PR.
Opened https://github.com/hyperledger/aries-vcx/issues/1111 to track it.

aries/aries_vcx/src/common/signing.rs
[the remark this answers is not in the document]
xprazak2: I am not sure.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -142,3 +107,72 @@ pub struct RestoreWalletConfigs {
...
+const WALLET_OPTIONS: &str =
+    r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true}"#;
+
+const SEARCH_OPTIONS: &str = r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true, "retrieveRecords": true}"#;
+
+impl BaseWallet for IndySdkWallet {}
+
+pub struct IndyTag((String, String));
+
+impl IndyTag {
```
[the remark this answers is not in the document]
xprazak2: Moved.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -27,59 +25,9 @@ impl IndySdkWallet {
...
-    fn drop(&mut self) {
-        let search_handle = self.search_handle;
 
-        thread::spawn(move || {
-            block_on(async {
-                internal::close_search_wallet(search_handle).await.ok();
-            });
-        });
+    pub fn get_wallet_handle(&self) -> WalletHandle {
+        self.wallet_handle
```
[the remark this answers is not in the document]
xprazak2: Changed, though the field being public was not something introduced in these changes.

aries/aries_vcx_core/src/wallet/indy/indy_record_wallet.rs
```diff
@@ -4,35 +4,38 @@ use serde::Deserialize;
...
+                record.category().into(),
+                record.name().into(),
+                record.value().into(),
                 tags_map,
             )
             .await?)
     }
 
-    async fn get_record(&self, name: &str, category: &str) -> VcxCoreResult<Record> {
+    async fn get_record(&self, category: &str, name: &str) -> VcxCoreResult<Record> {
```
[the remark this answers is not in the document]
xprazak2: Perhaps, but that can be done in a separate PR.

aries/aries_vcx_core/src/wallet/base_wallet.rs
```diff
@@ -1,124 +1,436 @@
...
-        msg: &[u8],
-    ) -> VcxCoreResult<Vec<u8>>;
+    async fn build_test_wallet() -> impl BaseWallet {
+        #[cfg(feature = "vdrtools_wallet")]
+        return dev_setup_indy_wallet().await;
+    }
 
-    async fn unpack_message(&self, msg: &[u8]) -> VcxCoreResult<UnpackMessageOutput>;
-}
+    #[cfg(feature = "vdrtools_wallet")]
```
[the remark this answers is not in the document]
xprazak2: This one should stay if we want to reuse the same code for running tests with different wallet implementations.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -1,18 +1,16 @@
...
-    SearchHandle, WalletHandle,
+use super::{
+    base_wallet::{BaseWallet, Record},
+    entry_tag::{EntryTag, EntryTags},
 };
+use crate::{errors::error::VcxCoreResult, WalletHandle};
 
-pub mod indy_wallet;
+pub mod indy_did_wallet;
+pub mod indy_record_wallet;
```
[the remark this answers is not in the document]
xprazak2: Agreed, these are no longer public.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -142,3 +107,72 @@ pub struct RestoreWalletConfigs {
...
+
+    pub fn from_entry_tag(tag: EntryTag) -> Self {
+        match tag {
+            EntryTag::Encrypted(key, val) => Self((key, val)),
+            EntryTag::Plaintext(key, val) => Self((format!("~{}", key), val)),
+        }
+    }
+}
+
+pub struct IndyTags(HashMap<String, String>);
```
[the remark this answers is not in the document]
xprazak2: They need to be public because the conversions they provide are used by other crates.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/profile.rs
```diff
@@ -20,14 +22,14 @@ use crate::{
     errors::error::{LibvcxError, LibvcxErrorKind, LibvcxResult},
 };
 
-pub fn try_get_main_wallet() -> LibvcxResult<Option<Arc<impl BaseWallet>>> {
+pub fn try_get_main_wallet() -> LibvcxResult<Option<Arc<IndySdkWallet>>> {
```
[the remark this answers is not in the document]
xprazak2: Because there is a code that expects `try_get_main_wallet` to return a wallet that has a wallet handle.

aries/aries_vcx_core/src/wallet/indy/indy_record_wallet.rs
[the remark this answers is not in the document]
mirgee: Yeah, definitely no need to do it in this one. Just an idea.

aries/aries_vcx_core/src/wallet/base_wallet.rs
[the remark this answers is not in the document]
mirgee: `dev_setup_indy_wallet` can be hidden in a `vdrtools_wallet`-specific module. Only one conditional compilation attribute per feature is necessary in test setup.

aries/aries_vcx_core/src/wallet/indy/mod.rs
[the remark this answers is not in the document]
mirgee: Besides `libvcx_core`, which is not too relevant, I see `IndyTags` are used in `credx_anoncreds`, which is clearly suspicious; anoncreds impl should not rely on a particular wallet implementation. Moreover, it is used only to construct `EntryTags` from a hashmap, while performing the `Encrypted` / `Plaintext` differentiation, which is (seemingly?) unnecessary in this particular case. Is it a problem of the current `credx_anoncreds` impl that it relies on a particular wallet impl, or is the problem unnecessary `IndyTags` abstraction?

aries/aries_vcx/src/common/signing.rs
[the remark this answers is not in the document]
mirgee: Why keep them if they are unused?

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/profile.rs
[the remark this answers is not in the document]
mirgee: The purpose of the profiles is to enable switching trait implementation. Constraining the output to a particular implementation defeats this purpose. In this context, there should not be code that relies on a wallet handle (indy wallet specific).

aries/aries_vcx/src/common/signing.rs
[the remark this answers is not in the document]
xprazak2: Removed.

aries/aries_vcx_core/src/wallet/indy/mod.rs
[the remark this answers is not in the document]
xprazak2: Well, `HashMap<String, String>` was the interface type for tags and we wrapped it in a custom type, so now the custom type is used instead.

I added deserialization for `EntryTags` so `IndyTags` and `HashMap` are no longer used in `credx_anoncreds`.

aries/aries_vcx_core/src/wallet/base_wallet.rs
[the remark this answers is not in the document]
xprazak2: Changed.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/profile.rs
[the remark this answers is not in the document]
xprazak2: `api_global` exposes API that mostly just wraps methods from `BaseWallet`. Any methods it adds, such as [close_main_wallet](https://github.com/hyperledger/aries-vcx/blob/main/aries/misc/legacy/libvcx_core/src/api_vcx/api_global/wallet.rs#L84), rely on having wallet handle as a global constant. My understanding was that we do not wish to support this crate in a long-term, so I see little  value in refactoring a code that we aim to drop.

If we intent to keep this, why do we need a wrapper crate?

Patrik-Stas (approved):

aries/aries_vcx_core/src/wallet/entry_tag.rs
```diff
@@ -77,3 +110,42 @@ impl From<EntryTags> for Vec<EntryTag> {
...
+        ]);
+
+        let res = serde_json::to_string(&tags).unwrap();
+
+        assert_eq!("{\"~a\":\"b\",\"c\":\"d\"}", res);
+    }
+
+    #[test]
+    fn shoud_deserialize_entry_tags() {
+        let json = "{\"a\":\"b\",\"~c\":\"d\"}";
```
mirgee, line 140: Using [this](https://docs.rs/serde_json/latest/serde_json/macro.json.html) makes the json more easily editable.

aries/aries_vcx_core/src/wallet/entry_tag.rs
```diff
@@ -19,15 +21,46 @@ impl Serialize for EntryTags {
         let mut map = serializer.serialize_map(Some(self.inner.len()))?;
         for tag in self.inner.iter() {
             match tag {
-                EntryTag::Encrypted(key, val) | EntryTag::Plaintext(key, val) => {
-                    map.serialize_entry(&key, &val)?
-                }
+                EntryTag::Encrypted(key, val) => map.serialize_entry(&key, &val)?,
+                EntryTag::Plaintext(key, val) => map.serialize_entry(&format!("~{}", key), &val)?,
```
mirgee, line 25: [As you said](https://github.com/hyperledger/aries-vcx/pull/1084#discussion_r1444318095) (and I agree), determining whether the tag is encrypted or not should be responsibility of the wallet implementation.

aries/aries_vcx_core/src/wallet/entry_tag.rs
```diff
@@ -77,3 +110,42 @@ impl From<EntryTags> for Vec<EntryTag> {
...
+    #[test]
+    fn should_serialize_entry_tags() {
+        let tags = EntryTags::new(vec![
+            EntryTag::Plaintext("a".into(), "b".into()),
+            EntryTag::Encrypted("c".into(), "d".into()),
+        ]);
+
+        let res = serde_json::to_string(&tags).unwrap();
+
+        assert_eq!("{\"~a\":\"b\",\"c\":\"d\"}", res);
```
mirgee, line 135: dtto

aries/aries_vcx_core/src/wallet/entry_tag.rs
```diff
@@ -77,3 +110,42 @@ impl From<EntryTags> for Vec<EntryTag> {
...
+pub fn pair_into_entry_tag(pair: (String, String)) -> EntryTag {
+    if pair.0.starts_with('~') {
+        EntryTag::Plaintext(pair.0.trim_start_matches('~').into(), pair.1)
+    } else {
+        EntryTag::Encrypted(pair.0, pair.1)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
mirgee, line 123: The (de)serialization of `EntryTag` enum is incongruent with (de)serialization of `EntryTags` because of the default derived impls for `EntryTag` and custom `EntryTag` impls, meaning e.g. that plaintext `EntryTag` (de)serialization doesn't add (strip) the `~`, but if the `EntryTag` is a member of `EntryTags`, it does. Can you please fix that and add a test?

aries/aries_vcx_core/src/wallet/entry_tag.rs
```diff
@@ -77,3 +110,42 @@ impl From<EntryTags> for Vec<EntryTag> {
         value.inner
     }
 }
+
+pub fn pair_into_entry_tag(pair: (String, String)) -> EntryTag {
```
mirgee, line 114: In these cases, how do you decide whether this should be standalone function, a static method or a `From` impl? I am not sure I see the pattern.

aries/aries_vcx_core/src/wallet/entry_tag.rs
```diff
@@ -77,3 +110,42 @@ impl From<EntryTags> for Vec<EntryTag> {
...
+            EntryTag::Encrypted("c".into(), "d".into()),
+        ]);
+
+        let res = serde_json::to_string(&tags).unwrap();
+
+        assert_eq!("{\"~a\":\"b\",\"c\":\"d\"}", res);
+    }
+
+    #[test]
+    fn shoud_deserialize_entry_tags() {
```
mirgee, line 139: typo. Also, it's better to make it easy to group the test names based on prefixes, e.g. `test_entry_tags_serialize` and `test_entry_tags_deserialize`, because then you can run `cargo test -- test_entry_tags` to run both tests.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -134,3 +133,26 @@ const WALLET_OPTIONS: &str =
 const SEARCH_OPTIONS: &str = r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true, "retrieveRecords": true}"#;
 
 impl BaseWallet for IndySdkWallet {}
+
+#[cfg(test)]
+pub mod test {
```
mirgee, line 138: The convention (which even you *mostly* follow) is to put test code in module called `tests`.

mirgee (changes requested):

aries/aries_vcx_core/src/wallet/indy/indy_record_wallet.rs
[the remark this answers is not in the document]
mirgee: Tracking issue #1112.

aries/aries_vcx_core/src/wallet/indy/mod.rs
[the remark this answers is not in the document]
mirgee: It would be nice to remove its usage from the abovementioned `libvcx_core` as well while we are at it.

mirgee (changes requested):

aries/aries_vcx_core/src/wallet/entry_tag.rs
xprazak2: Do I understand correctly that you expect having only a single variant `EntryTag::Tag(key, val)`?

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/profile.rs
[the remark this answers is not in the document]
xprazak2: Fair enough, but this change is backwards compatible. I do not know what the long-term plan is, but swapping wallet implementations seems like a good opportunity to move away from a crate that is explicitly declared as legacy.

If a decision is made to keep `libvcx_core` at that point, separate changes can be made for this crate to support other wallet implementations as well.

aries/aries_vcx_core/src/wallet/indy/mod.rs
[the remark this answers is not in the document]
xprazak2: `IndyTags` are now removed from `libvcx_core`.

aries/aries_vcx_core/src/wallet/indy/mod.rs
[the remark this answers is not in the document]
xprazak2: `internal` is used by `api_global`, so it needs to be public (at least for now).

aries/aries_vcx_core/src/wallet/entry_tag.rs
xprazak2: Nice.

aries/aries_vcx_core/src/wallet/entry_tag.rs
xprazak2: Fixed by no longer deriving (de)serialization traits for `EntryTag` since tags are never serialized individually.

aries/aries_vcx_core/src/wallet/entry_tag.rs
xprazak2: Changed into a static method for consistency.

aries/aries_vcx_core/src/wallet/entry_tag.rs
xprazak2: Changed.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -1,85 +1,30 @@
...
+use crate::{errors::error::VcxCoreResult, WalletHandle};
 
-pub mod indy_wallet;
+mod indy_did_wallet;
+mod indy_record_wallet;
+pub mod indy_tag;
 pub mod internal;
-pub mod signing;
 pub mod wallet;
 pub mod wallet_non_secrets;
```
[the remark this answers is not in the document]
xprazak2: Seems like it is not used, I removed it.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/wallet.rs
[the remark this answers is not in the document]
xprazak2: I am not sure, I did not delve too deep into this crate. Refactoring can be done separately.

aries/aries_vcx_core/src/wallet/indy/internal/mod.rs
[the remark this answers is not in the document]
xprazak2: It is an existing file/folder structure, changed for consistency.

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: Updated.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -142,3 +107,72 @@ pub struct RestoreWalletConfigs {
...
 }
+
+const WALLET_OPTIONS: &str =
+    r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true}"#;
+
+const SEARCH_OPTIONS: &str = r#"{"retrieveType": true, "retrieveValue": true, "retrieveTags": true, "retrieveRecords": true}"#;
+
+impl BaseWallet for IndySdkWallet {}
+
+pub struct IndyTag((String, String));
```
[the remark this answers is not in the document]
xprazak2: `IndyTag` is no longer public.

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: Updated.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/wallet.rs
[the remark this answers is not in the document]
mirgee: Sure, the point was more so that `internal` could be deleted if it wasn't used here (which seems like quite a bit benefit for little effort)

aries/aries_vcx_core/src/wallet/entry_tag.rs
mirgee: At this point I do think it probably would be better.

aries/aries_vcx_core/src/wallet/indy/mod.rs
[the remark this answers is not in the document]
mirgee: I was arguing for not using `internal` in `api_global`. But probably fine not extending the scope of this PR; leaving the choice up to you.

aries/misc/legacy/libvcx_core/src/api_vcx/api_global/profile.rs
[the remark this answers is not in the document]
mirgee: Moving away from `libvcx_core` is non-trivial, as it relied on by our nodejs codebase. On the other hand, swapping wallet implementation should be a matter of setting a feature flag.

aries/aries_vcx_core/src/wallet/entry_tag.rs
mirgee: I don't think this is a fix at all. These derived impls are way too easy to add. A struct using custom (de)serialize impl for it's member fields instead of relying on their own (de)serialize impl (i.e. inconsistent impls inside / outside of another struct) is clearly problematic.

aries/aries_vcx_core/src/wallet/indy/mod.rs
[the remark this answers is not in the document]
xprazak2: tracking issue: https://github.com/hyperledger/aries-vcx/issues/1115

aries/aries_vcx_core/src/wallet/entry_tag.rs
xprazak2: I generally agree that struct using custom (de)serialize for member fields will create problems.

However in this case, `EntryTags` is not divisible into individual fields. It is a container/wrapper and individual tags do not exist in json separately, only as a collection. `{ "a": "b" }` is not a tag, it is tags with a length of 1. I cannot (de)serialize a single `EntryTag`, same as the existing code was not (de)serializing `(String, String)`. It was always `"{ ... }" -> HashMap<String, String>`.

Patrik-Stas (dismissed):

aries/aries_vcx_core/src/wallet/entry_tag.rs
xprazak2: Changed to a tuple.

aries/aries_vcx_core/src/wallet/entry_tag.rs
xprazak2: No longer relevant since the tag is now a tuple.

aries/aries_vcx_core/src/wallet/entry_tags.rs
```diff
@@ -2,32 +2,26 @@ use std::fmt;
...
+//         match self {
+//             Self::Tag(key, _) => key,
+//         }
+//     }
+// }
 
 #[derive(Debug, Default, Clone, PartialEq)]
 pub struct EntryTags {
-    inner: Vec<EntryTag>,
+    inner: Vec<(String, String)>,
```
mirgee, line 24: Perhaps we can keep
```rust
pub type EntryTag = (String, String);
```
That way, the type is defined one place instead of being scattered around.

aries/aries_vcx_core/src/wallet/entry_tags.rs
```diff
@@ -2,32 +2,26 @@ use std::fmt;
...
+//     pub fn from_pair(pair: (String, String)) -> Self {
+//         Self::Tag(pair.0, pair.1)
+//     }
+
+//     pub fn key(&self) -> &str {
+//         match self {
+//             Self::Tag(key, _) => key,
+//         }
+//     }
+// }
```
mirgee, line 20: Just a reminder about this commented out code

mirgee (approved):

Patrik-Stas (approved):
lets goooooo 🚀

PR #1099, opened 2024-01-09
DID parser using nom

mirgee:
* (Hopefully) simplifies DID parsing implementation
* Covers edge cases not covered by the current DID parser

- [x] Test for specific edge cases
- [x] Clarify the expected output for specific edge cases
- [x] (Potentially injectable) method-specific parsing
- [ ] Performance comparison (optional)


did_core/did_parser_nom/src/did/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use self::parsing::parse_did;
+use crate::{error::ParseError, DidRange, DidUrl};
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct Did {
+    did: String,
+    method: Option<DidRange>,
+    namespace: Option<DidRange>,
+    id: DidRange,
```
Patrik-Stas, line 19: Hmm, I am not super sure about this, but is it still meaningful to use ranges for internal representation? The output of the high level parsers such as `parse_qualified_sovrin_did` is on happy path tuple of the individual fragments `(&'a str, &'a str, Option<&'a str>, &'a str)` - each being slice of the `did` attribute. Why not to just store `&str` slices as references to `did: String`? (I might be overseeing some challenges this might imply.)

In the original implementation, these Ranges were inherently created as our implementation went over the string, but since we have nom now, it seems we have to go extra mile to construct ranges after parsing is done. After `struct Did` is constructed, we really utilize the ranges somewhat special - seems that nom-produced slices would do?

did_core/did_parser_nom/src/did/parsing/did_core.rs
Patrik-Stas, line 1: Perhaps we could embed link in this file pointing directly to the Did grammar section of didcore spec? https://www.w3.org/TR/did-core/#did-syntax

did_core/did_parser_nom/src/did/parsing/did_core.rs
```diff
@@ -0,0 +1,102 @@
...
+// did = "did:" method-name ":" method-specific-id
+pub(super) fn parse_qualified_did(input: &str) -> IResult<&str, DidPart> {
+    tuple((tag("did"), method_name, opt(namespace), method_specific_id))(input)
+}
+
+pub(super) fn to_id_range(id: &str) -> DidRanges {
+    (None, None, Some(0..id.len()))
+}
+
+pub(super) fn to_did_ranges((did_prefix, method, namespace, id): DidPart) -> DidRanges {
```
Patrik-Stas, line 70: Arguable, but imo this belong to `mod.rs` as conversion to the ranges is related to our internal representation. Then this file would contain only implementation reflecting the didcore grammar spec

did_core/did_parser_nom/src/did/parsing/did_web.rs
```diff
@@ -0,0 +1,22 @@
...
+pub(super) fn parse_did_web(input: &str) -> IResult<&str, DidPart> {
+    fn did_web_method(input: &str) -> IResult<&str, &str> {
+        delimited(char(':'), tag("web"), char(':'))(input)
+    }
+
+    tuple((
+        tag("did"),
+        did_web_method,
+        opt(fail::<_, &str, _>),
+        take_till(|c| "?/#".contains(c)),
```
Patrik-Stas, line 20: A comment could be nice, this is non-obvious for reader. You mentioned there's some trickiness with `did:web`, but still I am not sure what this is/why

did_core/did_parser_nom/src/did_url/mod.rs
Patrik-Stas, line 1: assuming you accept de/serialization tests on `Did`, likewise there could be analogous tests here

did_core/did_parser_nom/src/did_url/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let did_url = String::deserialize(deserializer)?;
+        DidUrl::parse(did_url).map_err(serde::de::Error::custom)
+    }
+}
+
+impl TryFrom<&DidUrl> for Did {
```
Patrik-Stas, line 152: If we are implementing a trait for `Did`, shouldn't that rather be placed under the `did` module?

did_core/did_parser_nom/src/did_url/parsing.rs
```diff
@@ -0,0 +1,167 @@
...
+
+// did-url = did path-abempty [ "?" query ] [ "#" fragment ]
+pub fn parse_did_url(did_url: String) -> Result<DidUrl, ParseError> {
+    if did_url.is_empty() {
+        return Err(ParseError::InvalidInput("Empty input"));
+    }
+
+    let (remaining, did_ranges) = parse_did_ranges_with_empty_allowed(&did_url)?;
+
+    let (_, url_part) = all_consuming(parse_url_part)(remaining)?;
```
Patrik-Stas, line 150: I have a hunch this could be prettier if we don't deal with `DidRanges`, just in terms of raw parsers... but I am speaking bit out of my ballpark, not sure what challenges I am missing.

Just a hunch, if `parse_did_ranges` would return `IResult<&str, DidRanges>`, could't we instead have here something like 
```
let (did_parts_tuple, did_url_parts_tuple) = tuple((
  parse_did_ranges,
  parse_url_part
))?;
```
?

did_core/did_parser_nom/tests/did/negative.rs
```diff
@@ -0,0 +1,27 @@
...
+
+test_cases_negative! {
+    test_failure_case1: ""
+    test_failure_case2: "not-a-did"
+    test_failure_case3: "did:example"
+    test_failure_case4: "2ZHFFhtTD6hJqzux"
+    test_failure_case5: "did:indy:s@vrin:7Tqg6BwSSWapxgUDm9KKgg"
+    test_failure_case6: "did:indy:sovrin:alpha:%0zqg6BwS.Wapxg-Dm9K_gg"
+    test_failure_case7: "did:sov:2wJPyULfLLnYTEFYzByf"
+    test_failure_case8: "did:sov:2wJPyULfOLnYTEFYzByfUR"
```
Patrik-Stas, line 23: It's non-obvious why these are not valid dids - I presume some non-base58 characters? I'd suggest to make test cases more descriptive, although they won't line up here visually as pretty

did_core/did_parser_nom/tests/did/negative.rs
```diff
@@ -0,0 +1,27 @@
...
+    test_failure_case9: "2wJPyULfLLnYTEFYzByf"
+    test_failure_case10: "2wJPyULfOLnYTEFYzByfUR"
```
Patrik-Stas, line 25: what's the difference between these two, given they both are negative?

did_core/did_parser_nom/tests/did/positive.rs
```diff
@@ -0,0 +1,93 @@
...
+                assert_eq!(parsed_did.id(), $expected_id, "ID");
+            }
+        )*
+    };
+}
+
+test_cases_positive! {
+    test_case1:
+        "did:example:123456789abcdefghi",
+        "did:example:123456789abcdefghi",
```
Patrik-Stas, line 23: given every test case have the first 2 values identical, it seems like redundancy

Patrik-Stas (changes requested):
left some comments, but very very cool 👍

did_core/did_parser_nom/src/did/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+impl std::fmt::Debug for Did {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Did")
+            .field("did", &self.did)
+            .field("method", &self.method())
+            .field("id", &self.id())
+            .finish()
```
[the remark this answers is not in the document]
mirgee: Good catch!

did_core/did_parser_nom/src/did/parsing/did_core.rs
mirgee: Good point and if you mean `parsing/mod.rs`, I tend to agree. These conversions sit between the parsing (they know about the parsed parts) and our internal representation (they know about the ranges), same as the `parsing` module. Additionally, `did_url/parsing.rs` might be split into parsers and range convertors for symmetry.

did_core/did_parser_nom/src/did/parsing/did_sov.rs
```diff
@@ -0,0 +1,37 @@
...
+fn base58char(input: &str) -> IResult<&str, &str> {
+    recognize(one_of(
+        "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz",
+    ))(input)
+}
+
+// idstring = 21*22(base58char)
+pub(super) fn parse_unqualified_sovrin_did(input: &str) -> IResult<&str, &str> {
+    recognize(many_m_n(21, 22, base58char))(input)
+}
```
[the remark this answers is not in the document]
mirgee: Moved the alphabet static const a layer above

did_core/did_parser_nom/src/did/parsing/did_web.rs
mirgee: This stops parsing the id portion when either DID URL query, path or fragment are encountered.

did_core/did_parser_nom/src/did_url/parsing.rs
mirgee: I am not sure I understand what you mean or are going for, but I strongly disagree with a parser returning ranges.

did_core/did_parser_nom/tests/did/negative.rs
mirgee: I assumed comparing against the valid cases makes it clear.

did_core/did_parser_nom/tests/did/negative.rs
mirgee: What do you mean by difference? :) They are strings, isn't the difference obvious? :)

did_core/did_parser_nom/tests/did/positive.rs
mirgee: I agree with the statement, it seems redundant, but the input and expected output of a method are conceptually distinct (e.g. can change independently).

did_core/did_parser_nom/src/did_url/mod.rs
mirgee: I don't think implementing a trait for a struct necessarily means it should be placed in the same module. Doing what you suggest would mean:

* `did` module would know about `DidUrl`, whereas it currently doesn't (which makes sense, `DidUrl` conceptually "derives" from / elaborates on `Did`),
* `DidUrl` would need to expose multiple private fields it currently doesn't.

did_core/did_parser_nom/src/did/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct Did {
+    did: String,
+    method: Option<DidRange>,
+    namespace: Option<DidRange>,
+    id: DidRange,
+}
+
+// TODO: Add a builder / constructor so that we don't have to create strings and parse them?
```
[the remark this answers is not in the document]
mirgee: You seem add the unstated assumption that the reason to do this for the builder methods to be infallible, whereas it was more to allow users to construct (relative) DID (URLs) from parts (paths, queries, fragments), which was found to be commonly useful during the did doc / resolution impl (at least before the major rewrite).

did_core/did_parser_nom/src/did/mod.rs
mirgee: No, the main reason the ranges were conceived in first place was to avoid exactly this - dealing with self-referential structs.

did_core/did_parser_nom/tests/did/negative.rs
Patrik-Stas: Both input fail, but the inputs are slightly difference. What am I missing?

did_core/did_parser_nom/tests/did/negative.rs
Patrik-Stas: Are you suggesting I should look up into different file to understand what is `test_failure_case7` doing?

did_core/did_parser_nom/src/did/mod.rs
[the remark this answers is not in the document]
Patrik-Stas: Ah I see, makes sense. Yeah that was indeed my assumption

did_core/did_parser_nom/src/did/mod.rs
Patrik-Stas: I see fair, yeah now I see the self-ref struct I am suggesting gets tricky real quick, even if willing to give up on mutability of the struct.

did_core/did_parser_nom/src/did/parsing/did_core.rs
Patrik-Stas: Yes, I meant `parsing/mod.rs`, didn't mean to suggest a new module. Anyway, I'll leave the final placement up to you.

did_core/did_parser_nom/src/did/parsing/did_web.rs
Patrik-Stas: I understand that `did:web` can have variable number of colons, eg `did:web:w3c-ccg.github.io:user:foo:bar:alice`, but I am still missing the `?/#` handling here. If I understand the current implementation right, given input 
```
did:web:w3c-ccg.github.io:user:alice#abcd
```
which is in fact DID URL, will in fact succesfully parse DID `did:web:w3c-ccg.github.io:user:alice`. Is that right?

But if `did:web` parser implementation is lenient enough to accept DID URL as DID, why wouldn't other did methods do the same? Eg. shouldn't `did:sov:abcd123#abcd` be also parsed as DID `did:sov:abcd123`.

did_core/did_parser_nom/src/did_url/parsing.rs
Patrik-Stas: I had feeling like this could be refactored, but wasn't' quite sure about details. If this didn't ring any bells, chances are this has no merit. Not a big deal anyway, the code is well readable as it is.

Patrik-Stas:
@mirgee just a reminded there's a few comments unresolved but I am looking forward to merge this  🚀

did_core/did_parser_nom/src/did/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+impl Default for Did {
+    fn default() -> Self {
+        Self {
+            did: "did:example:123456789abcdefghi".to_string(),
+            method: Some(4..11),
+            namespace: None,
+            id: 12..30,
+        }
+    }
+}
```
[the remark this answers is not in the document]
mirgee: Yeah, I agree, probably not

did_core/did_parser_nom/tests/did/negative.rs
mirgee: To understand what makes a DID invalid you should understand what makes it valid, which at the very least means looking at the positive test cases, preferably the method's specification; there is no way to explain the same in a test name.

At best, the test name can give an illusion of understanding or partial explanation (like `test_sov_unqualified_invalid_len` or `test_sov_unqualified_invalid_char`), at the cost of long, hard-to-read and potentially confusing test names (like `test_indy_non_method_specific_id_char_in_namespace`).

I understand and am sympathetic to the argument that (or test comments) is better than no explanation / documentation at all.

did_core/did_parser_nom/tests/did/negative.rs
mirgee: No problem, let's do it

Patrik-Stas: This discussion got duplicated, I am here quoting from thread https://github.com/hyperledger/aries-vcx/pull/1099#discussion_r1457413330

> To understand what makes a DID invalid you should understand what makes it valid, which at the very least means looking at the positive test cases, preferably the method's specification; there is no way to explain the same in a test name.

> At best, the test name can give an illusion of understanding or partial explanation (like test_sov_unqualified_invalid_len or test_sov_unqualified_invalid_char), at the cost of long, hard-to-read and potentially confusing test names (like test_indy_non_method_specific_id_char_in_namespace).

> I understand and am sympathetic to the argument that (or test comments) is better than no explanation / documentation at all.

I think the names you suggested are better then simply `negative_test_x`. Yes, the example `test_indy_non_method_specific_id_char_in_namespace` is taken to the other extreme of the spectrum. I don't think it has to be that explicit, but at least brief indication what the test is testing can be helpful to reader.

did_core/did_parser_nom/src/did/mod.rs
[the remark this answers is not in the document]
mirgee: There is no issue in removing it in this PR. The problem will manifest when the usage of the old parser is replaced by this one (of course).

did_core/did_parser_nom/src/did/parsing/did_web.rs
mirgee: All the parsers parse DID URLs (to be reusable when parsing a DID URL), but for DID parsing the `all_consuming` combinator is used.

did_core/did_parser_nom/tests/did/negative.rs
mirgee: I gave names to the negative cases, let me know if they are good enough

did_core/did_parser_nom/src/did/mod.rs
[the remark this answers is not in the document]
mirgee: I am not sure that using a specific (de)serializer is the way to test ((de)serializer-agnostic) `Serialize` / `Deserialize` impl. Serde documentation [suggests](https://serde.rs/unit-testing.html) using `serde_test` instead.

did_core/did_parser_nom/src/did/parsing/did_core.rs
mirgee: Moved.

did_core/did_parser_nom/src/did_url/mod.rs
mirgee: Let's leave this discussion to the [comment above](https://github.com/hyperledger/aries-vcx/pull/1099#discussion_r1448754947).

did_core/did_parser_nom/tests/did/negative.rs
Patrik-Stas: looks good! Thanks

did_core/did_parser_nom/src/did/parsing/did_web.rs
Patrik-Stas: I think it would be worthy a brief comment in the code, but no biggie

did_core/did_parser_nom/src/did/mod.rs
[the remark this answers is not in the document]
Patrik-Stas: Ah I see. I didn't realize that. But seems like right thing to do.

Patrik-Stas (changes requested):
Almost there, just minor comments

did_core/did_parser_nom/tests/did/serde.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[test]
+fn test_did_serialization() {
+    assert_tokens(&Did::parse(DID.to_string()).unwrap(), &[Token::Str(DID)]);
+}
+
+#[test]
+fn test_did_deserialization() {
+    assert_tokens(&Did::parse(DID.to_string()).unwrap(), &[Token::Str(DID)]);
+}
```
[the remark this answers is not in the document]
mirgee: Was using different assert and forgot to remove

Patrik-Stas (approved):

Issue #1091, opened 2023-12-18
 DID:SOV DIDDoc does expand shorthand verkey notation for publicKeyBase58


gmulhearn:
part of this series of issues; https://github.com/hyperledger/aries-vcx/issues/

On sovrin ledgers, it is possible to register a NYM's verkey in an abbreviated form with a tilda prefix (e.g. `~CoRER63DVYnWZtK8uAzNbx`). Some indy notes on this here: https://hyperledger-indy.readthedocs.io/projects/sdk/en/latest/docs/getting-started/indy-walkthrough.html#step-3-getting-the-ownership-for-steward-s-verinym.

This abbreviated verkey is also what the ledger returns when you do a get_nym:
```
    let nym_res = profile.ledger_read.get_nym(&profile.institution_did).await.unwrap();
    println!("{}", nym_res);
```
```
{"op":"REPLY","result":{"seqNo":1,"txnTime":null,"data":"{\"dest\":\"V4SGRU86Z58d6TV7PBUe6f\",\"identifier\":null,\"role\":\"0\",\"seqNo\":1,\"txnTime\":null,\"verkey\":\"~CoRER63DVYnWZtK8uAzNbx\"}","dest":"V4SGRU ...
```

since the did_resolver_sov crate is just using the ledger's returned `verkey` as the `publicKeyBase58` for the DID's `didDoc.verificationMethod[0]`, the DIDDoc ends up looking like:
```
...,
"verificationMethod": [
    {
      "id": "did:sov:V4SGRU86Z58d6TV7PBUe6f",
      "controller": "did:sov:V4SGRU86Z58d6TV7PBUe6f",
      "type": "Ed25519VerificationKey2018",
      "publicKeyBase58": "~CoRER63DVYnWZtK8uAzNbx"
    }
  ],
```

This is ofcourse invalid. It should be the responsibility of the did_resolver_sov crate to notice if the get_nym verkey is in it's abbreviated form (tilda prefix), and if it is, then it should transform the NYM+Abbreviated-Verkey into the full Verkey, which can then be set as `publicKeyBase58`.

Issue #1090, opened 2023-12-18
DID:SOV DIDDoc does resolve without a service/endpoint


gmulhearn:
part of this series of issues; https://github.com/hyperledger/aries-vcx/issues/1089

in the did_resolver_sov crate, if the did:sov DID does not have an endpoint attribute, then the entire DIDDoc fails to resolve. 

E.g. the following fails with `NotFound("DID not found")`:
```

#[tokio::test]
async fn write_new_nym_and_get_did_doc() {
    let profile = build_setup_profile().await;
    let (new_nym, verkey) = profile
        .wallet
        .create_and_store_my_did(None, None)
        .await
        .unwrap();

    profile
        .ledger_write
        .publish_nym(
            &profile.wallet,
            &profile.institution_did,
            &new_nym,
            Some(&verkey),
            None,
            None,
        )
        .await
        .unwrap();

    // NEED TO WRITE ENDPOINT FOR IT TO RESOLVE
    // write_test_endpoint(&profile.wallet, &profile.ledger_write, &new_nym).await;

    let resolver = DidSovResolver::new(profile.ledger_read);
    let did = format!("did:sov:{}", new_nym);

    let did_doc = resolver
        .resolve(
            &Did::parse(did.clone()).unwrap(),
            &DidResolutionOptions::default(),
        )
        .await
        .unwrap();

    println!(
        "{}",
        serde_json::to_string_pretty(&did_doc.did_document()).unwrap()
    );
}
```

Expected behaviour according to spec https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html#crud-operation-definitions :

> If an endpoint ATTRIB does not exist for the NYM, the service block is an empty array (e.g. "service": []

PR #1105, opened 2024-01-10
Fix: #1090 #1091 DID:SOV DIDDoc resolves without a service/endpoint and expands shorthand verkey notation for publicKeyBase58

lukewli-anonyome:
This PR introduces a fix for the DIDDoc resolution error outlined in issue https://github.com/hyperledger/aries-vcx/issues/1090 and the shorthand verkey error outlined in issue https://github.com/hyperledger/aries-vcx/issues/1091.

Closes https://github.com/hyperledger/aries-vcx/issues/1090
Closes https://github.com/hyperledger/aries-vcx/issues/1091


gmulhearn-anonyome:
weird that this clippy error is just showing up now :( maybe try fixing it (removing the dep)

did_core/did_methods/did_resolver_sov/src/resolution/utils.rs
```diff
@@ -53,6 +53,15 @@ fn unix_to_datetime(posix_timestamp: i64) -> Option<DateTime<Utc>> {
...
 
+fn expand_abbreviated_verkey(did: &str, verkey: &str) -> String {
+    if verkey.starts_with('~') {
+        let did_public_key_part = &did[8..];
+        format!("{}{}", did_public_key_part, &verkey[1..])
+    } else {
+        verkey.to_string()
+    }
+}
+
```
gmulhearn-anonyome, line 64: this won't exactly work; have a dig around some of the links in the original ticket. Here's a test vector to play with:

* nym/did: "7Sqc3ne5NfUVxMTrHahxz3"
* shortened verkey: "~DczaFTexiEYv5abkEUZeZt"
* expected (expanded) verkey: "4WkksEAXsewRbDYDz66aTdjtVF2LBxbqEMyF2WEjTBKk"

did_core/did_methods/did_resolver_sov/src/resolution/utils.rs
```diff
@@ -53,6 +53,21 @@ fn unix_to_datetime(posix_timestamp: i64) -> Option<DateTime<Utc>> {
...
 
+fn expand_abbreviated_verkey(did: &str, verkey: &str) -> String {
+    if let Some(stripped_key) = verkey.strip_prefix('~') {
+        let decoded_did = bs58::decode(did).into_vec().unwrap();
+        let decoded_stripped_key = bs58::decode(stripped_key).into_vec().unwrap();
+        let decoded_did_string = String::from_utf8(decoded_did).unwrap();
+        let decoded_stripped_key_string = String::from_utf8(decoded_stripped_key).unwrap();
+
+        let decoded_verkey = format!("{}{}", decoded_did_string, decoded_stripped_key_string);
+
```
gmulhearn-anonyome: not the biggest fan of treating raw bytes (Vec<u8> / &[u8]) as `String`s. I think you did this so that you could concatenate the bytes, which is correctly.

The `String::from_utf8` is error-prone here, since we aren't necessarily dealing with utf8 bytes, and this will panic if they aren't valid utf8 bytes.

there are other ways to concatenate bytes together (two vec<u8>'s )

Also in general; avoid `unwrap`s in non-testing code; you might have to make this return a `Result`

did_core/did_methods/did_resolver_sov/src/resolution/utils.rs
```diff
@@ -53,6 +53,29 @@ fn unix_to_datetime(posix_timestamp: i64) -> Option<DateTime<Utc>> {
         .map(|date_time| DateTime::<Utc>::from_naive_utc_and_offset(date_time, Utc))
 }
 
+fn expand_abbreviated_verkey(did: &str, verkey: &str) -> Result<String, DidSovError> {
+    if let Some(stripped_key) = verkey.strip_prefix('~') {
+        let stripped_did = &did[8..];
```
gmulhearn-anonyome, line 58: bit of a nit pick;

i'd refer if we referred to "stripped_did" as the "nym". i think that term is more widely understood.

also, altho `did[8..]` would work to remove the `did:sov` prefix, it isn't very clear that this is what is happening here.

IMO it would be nice if `expand_abbreviated_verkey` just took in `nym` rather than `did`.

then where `expand_abbreviated_verkey` is called, you'll need to pre-calculate the `nym` from the `did`. you might be able to use the `ddo_id` 
`Did` struct for this actually; it should have a `.id()` method, which will return the end-part of the DID (aka the NYM)

Patrik-Stas:
@lukewli-anonyome @gmulhearn-anonyome I am sorry folks for the conflicts that have arose after finally merging https://github.com/hyperledger/aries-vcx/pull/1097 

However, when you have bandwidth feel free to resume this effort

did_core/did_methods/did_resolver_sov/tests/resolution.rs
```diff
@@ -58,3 +58,50 @@ async fn test_error_handling_during_resolution() {
...
+
+    let resolver = DidSovResolver::new(profile.ledger_read);
+    let did = format!("did:sov:{}", profile.institution_did);
+
+    let DidResolutionOutput { did_document, .. } = resolver
+        .resolve(&Did::parse(did.clone()).unwrap(), &())
+        .await
+        .unwrap();
+
+    println!("{}", serde_json::to_string_pretty(&did_document).unwrap());
```
gmulhearn, line 106: nit; could you remove this pls

did_core/did_methods/did_resolver_sov/src/resolution/utils.rs
```diff
@@ -69,50 +91,71 @@ pub(super) async fn ledger_response_to_ddo(
...
+            serde_json::Value::Null
+        }
+    };
+
+    let mut ddo = DidDocument::new(ddo_id.clone());
+
+    let mut services = Vec::new();
+
+    if !service_data.is_null() {
+        let endpoint: EndpointDidSov = serde_json::from_value(service_data["endpoint"].clone())?;
```
gmulhearn, line 107: Note for anyone else reviewing; to my surprise, this should not be prone to panics:

https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html#method.index

thanks serde!

gmulhearn (approved):
nice! thanks Luke

PR #1106, opened 2024-01-11
refactor: keep single way to resolve peer dids

Patrik-Stas:
PR https://github.com/hyperledger/aries-vcx/pull/1097 performed changes to did resolution in did:peer:2 contexts. There's lack of clarity whether these changes should be accepted. This reverts the changes and opens discussion whether the changes in the #1097 should be accepted, or reverted (by merging this PR)

### DIFF
--- a/aries/aries_vcx/src/protocols/did_exchange/state_machine/requester/request_sent/mod.rs
+++ b/aries/aries_vcx/src/protocols/did_exchange/state_machine/requester/request_sent/mod.rs
@@ -3,9 +3,8 @@ use std::sync::Arc;
 use did_parser::Did;
 use did_peer::{
     peer_did::{numalgos::numalgo2::Numalgo2, PeerDid},
-    resolver::{options::PublicKeyEncoding, PeerDidResolutionOptions, PeerDidResolver},
+    resolver::options::PublicKeyEncoding,
 };
-use did_resolver::traits::resolvable::{resolution_output::DidResolutionOutput, DidResolvable};
 use did_resolver_registry::ResolverRegistry;
 use messages::msg_fields::protocols::did_exchange::{
     complete::Complete as CompleteMessage, request::Request, response::Response,
@@ -23,6 +22,7 @@ use crate::{
         states::{completed::Completed, requester::request_sent::RequestSent},
         transition::{transition_error::TransitionError, transition_result::TransitionResult},
     },
+    utils::didcomm_utils::resolve_didpeer2,
 };
 
 impl DidExchangeRequester<RequestSent> {
@@ -34,16 +34,8 @@ impl DidExchangeRequester<RequestSent> {
         let their_did_document = resolver_registry
             .resolve(their_did, &Default::default())
             .await?
-            .did_document
-            .clone();
-        let DidResolutionOutput { did_document, .. } = PeerDidResolver::new()
-            .resolve(
-                our_peer_did.did(),
-                &PeerDidResolutionOptions {
-                    encoding: Some(PublicKeyEncoding::Base58),
-                },
-            )
-            .await?;
+            .did_document;
+        let our_did_document = resolve_didpeer2(our_peer_did, PublicKeyEncoding::Base58).await?;
         let invitation_id = Uuid::new_v4().to_string();
 
         let request = construct_request(invitation_id.clone(), our_peer_did.to_string());
@@ -55,7 +47,7 @@ impl DidExchangeRequester<RequestSent> {
                     invitation_id,
                 },
                 their_did_document,
-                did_document,
+                our_did_document,
             ),
             output: request,
         })
@@ -82,16 +74,9 @@ impl DidExchangeRequester<RequestSent> {
         } else {
             let peer_did = PeerDid::<Numalgo2>::parse(response.content.did)
                 .map_err(to_transition_error(self.clone()))?;
-            let DidResolutionOutput { did_document, .. } = PeerDidResolver::new()
-                .resolve(
-                    peer_did.did(),
-                    &PeerDidResolutionOptions {
-                        encoding: Some(PublicKeyEncoding::Base58),
-                    },
-                )
+            resolve_didpeer2(&peer_did, PublicKeyEncoding::Base58)
                 .await
-                .map_err(to_transition_error(self.clone()))?;
-            did_document
+                .map_err(to_transition_error(self.clone()))?
         };
 
         let complete_message = construct_didexchange_complete(
--- a/aries/aries_vcx/src/protocols/did_exchange/state_machine/responder/response_sent/mod.rs
+++ b/aries/aries_vcx/src/protocols/did_exchange/state_machine/responder/response_sent/mod.rs
@@ -4,9 +4,8 @@ use aries_vcx_core::wallet::base_wallet::BaseWallet;
 use did_doc::schema::did_doc::DidDocument;
 use did_peer::{
     peer_did::{numalgos::numalgo2::Numalgo2, PeerDid},
-    resolver::{options::PublicKeyEncoding, PeerDidResolutionOptions, PeerDidResolver},
+    resolver::options::PublicKeyEncoding,
 };
-use did_resolver::traits::resolvable::{resolution_output::DidResolutionOutput, DidResolvable};
 use did_resolver_registry::ResolverRegistry;
 use messages::msg_fields::protocols::did_exchange::{
     complete::Complete, request::Request, response::Response,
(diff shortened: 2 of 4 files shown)


Patrik-Stas:
Hi @gmulhearn-anonyome @xprazak2 @nain-F49FF806 
can you folks have a look and share your thoughts which way you lean toward?

Let me know if you'd need some additional context or background. Not posting initially much so you can form opinion independently, without biases.

In general it's question:
- of using shared general interface of `DidResolvable` (represented by this revert) 
- VS having additional resolution interface on the bottom-most component and make that public, eg. `DidPeer<Numalog2` (represented by keeping #1097 as is, no revert)

gmulhearn:
to be honest, i don't really have strong opinions after looking at this for 15mins. I understand the two options as follows:
1. `PeerDid<Numalgo2>::resolve` owns the logic for resolving a `DidDocument` and `PeerDidResolver` uses it. But most code holding `PeerDid<_>`s just use the `PeerDid::resolve` method, since it is more concise then constructing a `PeerDidResolver` (the conciseness is evident from the positive git diff on this)
2. `PeerDidResolver` owns the logic for resolving a `DidDocument`, and `PeerDid<Numalgo2>::resolve` is removed

IMO it makes sense if `PeerDidResolver` owns the main logic for resolving a `DidDocument` (option 2), however this option clearly bloats the code up. So maybe another option:
3. `PeerDidResolver` owns the main logic for resolving a `DidDocument`, and the `PeerDid<Numalgo2>::resolve` still exists, but uses the `PeerDidResolver::resolve` for resolving. That way, we still have the convenient `PeerDid::resolve` method for code conciseness, but `PeerDidResolver` still owns the resolution logic.

but also the problem with `3` is that now `PeerDid::resolve` is unnecessarily async (whereas in option 1 it is not).

Patrik-Stas:
I like that. There's no logic duplication, both functions still refer to >>resolving<< (in my original-original version I distinguished "resolving" VS "decoding" which behaved little bit different) and for me, importantly, gives the ergonomic API in contexts where users work with PeerDid instances directly. 

The `resolve` method would then look like this
```
impl PeerDid<Numalgo2> {
    pub async fn resolve(
        &self,
        public_key_encoding: PublicKeyEncoding,
    ) -> Result<DidDocument, DidPeerError> {
        let DidResolutionOutput { did_document, .. } = PeerDidResolver::new()
            .resolve(
                self.did(),
                &PeerDidResolutionOptions {
                    encoding: Some(public_key_encoding),
                },
            )
            .await?;
        Ok(did_document)
    }
}
```
and `PeerDidResolver` goes without modifications.

@mirgee what do you say?

mirgee:
I say if you both like it then proceed, no need to ask me, who am I to stand in the face of majority anyway? :)

gmulhearn:
Nice, yea again; I don't feel strongly about any of the options, they all have pros and cons.

Personally if I was consuming Aries-vcx and kept having to type out 

```
let DidResolutionOutput { did_document, .. } = PeerDidResolver::new()
            .resolve(
                self.did(),
                &PeerDidResolutionOptions {
                    encoding: Some(public_key_encoding),
                },
            )
            .await?;
        Ok(did_document)
```

then I would just make my own helper function/method like PeerDid::resolve and put that code in there

Patrik-Stas:
I omitted my arguments or opinion upon starting this thread, which in nutshell is

- General interfaces for generic use cases, local APIs for local use cases
- Giving `PeerDid` users the only option through general `DidResolvable` interface is far less developer friendly
   - Need to deal with all the extra fluff which is useless for the particular need (no use of metadata, function is async even though it's in fact sync)
- The cost of adding `fn resolve` on `PeerDid<N>` is API surface expanded by additional function, I think well worth the benefits it brings consumers. 

About the option 3:
- It does `DidPeer` dependent on `DidPeerResolver`, but it's encapsulated internal dependency, so I don't see this as much of a problem. 
- What I see as main downside is rather it's still `async` even though it doesn't have to be, so not 100% achieving what I am after

You see, in fact the only thing this "DidPeerResolver" does is adding `AlsoKnown` field, which is even optional. If we were to decide not to do it, `DidPeerResolver` would in fact do completely nothing. The actual gore did peer resolution lies in decoding the DID ID string, not in adding `alsoKnownAs` to the decoded did document. The resolution largerly happens in `DidPeer` itself either way, it's just a question of whether there's exposed additional catered optimized API, or only 1 general interface.


I am opinionated about this not simply because of this particular issue (I could just put helper function to both `aries-vcx` and `aries-vcx-agent` and call it a day), but rather for the precedent of how we think about things like this, and what's the approach to APIs (I am quite sure same way goes for others).

I am favoring the mentioned variants in order of preference:
1. `option 2`
2. `option 3`
3. `option 1`

I believe we all understand the options, the trade offs and philosophies between the different options. Please either write yours or say you abstain from the vote. 

@nain-F49FF806 @xprazak2 if you'd like, feel free to join the battlefield here :-D

mirgee:
I abstain from the vote then.

nain-F49FF806:
Hi, I took a stroll through 
- [Aries RFC 0627: Static Peer DIDs](https://github.com/hyperledger/aries-rfcs/blob/main/features/0627-static-peer-dids/README.md#targeted-generation-methods-numalgo)
-  [Peer DID Method Specification](https://identity.foundation/peer-did-method-spec/#method-4-short-form-and-long-form)
for context, since I haven't actively interacted with `did:peer` before. 

What I think:

#### 1. Method specific resolution is not sufficient, so better to have a general, common resolution interface. 

What I see is that there is a general [expectation of flexibility on reading (/resolving?) side](https://github.com/hyperledger/aries-rfcs/blob/main/features/0627-static-peer-dids/README.md#targeted-generation-methods-numalgo). 

> ..MUST accept peer DIDs that use any of these methods
>  ..MUST give peer DIDs that use at least one of these three methods

So having a general, common interface for *resolution* makes more sense (but may not be so relevant for generation). 

The generation code can be modular, and distributed separately for each method. But any resolution is incomplete (in Aries context) without supporting multiple methods. So why separate the resolution logic? 

This would perhaps indicate to go with something like @mirgee [suggests here](https://github.com/hyperledger/aries-vcx/pull/1106#issuecomment-1888666221). 

#### 2. did:peer:4 may require a dedicated resolver due to storage requirement. 

*Please correct me if I am mistaken on this.* My reading of method 4 makes me think it'll be necessary to have some storage to 'remember' mapping of short -> long form representations of did:peer4. In that case, having the resolver external could be necessary/more ergonomic. 

Otherwise, we may need to add some storage interface argument to PeerDid<NumAlgo4>'s resolve method.

Conclusion: I like the convenience of having a `resolve` method local to the PeerDid<_> struct, but if it's not a consistent interface across methods, or has gotchas  … then I'd prefer a dedicated revolver interface instead. 



Currently, given the above considerations, I would vote for option 1.

Patrik-Stas:
Hi @nain-F49FF806 thanks a lot for taking time to go through this! Without previous context and involvement, this is quite a chunk  to swallow! 

> Please correct me if I am mistaken on this. My reading of method 4 makes me think it'll be necessary to have some storage to 'remember' mapping of short -> long form representations of did:peer4. In that case, having the resolver external could be necessary/more ergonomic.

It is a good reminder - for `did:peer:4*` resolution to be complete, it would need access to some state (and perhaps not just ready, but also writes to enable resolution of future, shortened form of did:peer:4 DIDs), so it wouldn't in fact be much more of a win to have sync `resolve()` defined for `PeerDid<Numalgo4>` like it can be done for `PeerDid<Numalgo2>` (which is always "self-encoding" and synchronous).

You've expressed favor in both `1` and `3` - to prevent any misunderstanding, what's your vote in order of preference?

Patrik-Stas:
I want to wrap this up so despite your official abstain so:
- Despite official abstaining from vote, I count mirgee's vote for `option 1` given strong held opinions expressed in other channels
- Naian seem bit on the fence, but given the nuance response, it seem to gravitate more toward `option 1`
- George not strongly opinionated but suggested `option 3`
- My top preference is for `option 2` as I explained why from my perspective `option 3` is not achieving goal I am after in my `option 2` suggestion

Thank you all for participation! This PR applies `option 1` and is now ready for review & merge.

aries/aries_vcx/src/protocols/did_exchange/state_machine/requester/request_sent/mod.rs
```diff
@@ -33,8 +34,9 @@ impl DidExchangeRequester<RequestSent> {
         let their_did_document = resolver_registry
             .resolve(their_did, &Default::default())
             .await?
-            .did_document;
-        let did_document = our_peer_did.resolve(PublicKeyEncoding::Base58)?;
+            .did_document
+            .clone();
+        let our_did_document = resolve_didpeer2(our_peer_did, PublicKeyEncoding::Base58).await?;
```
mirgee, line 39: Is this clone necessary?

PR #1108, opened 2024-01-18
refactor(aries_vcx_core): switch from derive_builder to typed-builder

xprazak2:
Based on https://github.com/hyperledger/aries-vcx/pull/1093#discussion_r1456331167

### DIFF
--- a/aries/aries_vcx_core/src/wallet/indy/mod.rs
+++ b/aries/aries_vcx_core/src/wallet/indy/mod.rs
@@ -83,7 +83,7 @@ impl Drop for IndyWalletRecordIterator {
     }
 }
 
-#[derive(Clone, Debug, Default, TypedBuilder, Serialize, Deserialize)]
+#[derive(Clone, Debug, TypedBuilder, Serialize, Deserialize)]
 #[builder(field_defaults(default))]
 pub struct WalletConfig {
     pub wallet_name: String,
@@ -106,7 +106,7 @@ pub struct WalletConfig {
     pub rekey_derivation_method: Option<String>,
 }
 
-#[derive(Clone, Debug, Default, TypedBuilder, Serialize, Deserialize)]
+#[derive(Clone, Debug, TypedBuilder, Serialize, Deserialize)]
 #[builder(field_defaults(default))]
 pub struct IssuerConfig {
     pub institution_did: String,
--- a/aries/aries_vcx_core/src/wallet/indy/mod.rs
+++ b/aries/aries_vcx_core/src/wallet/indy/mod.rs
@@ -1,10 +1,10 @@
 use std::thread;
 
 use async_trait::async_trait;
-use derive_builder::Builder;
 use futures::executor::block_on;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use typed_builder::TypedBuilder;
 
 use crate::{
     errors::error::{AriesVcxCoreError, VcxCoreResult},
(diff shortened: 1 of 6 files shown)


Patrik-Stas (approved):

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -83,11 +83,14 @@ impl Drop for IndyWalletRecordIterator {
     }
 }
 
-#[derive(Clone, Debug, Default, Builder, Serialize, Deserialize)]
-#[builder(setter(into, strip_option), default)]
+#[derive(Clone, Debug, Default, TypedBuilder, Serialize, Deserialize)]
+#[builder(field_defaults(setter(into, strip_option), default))]
 pub struct WalletConfig {
+    #[builder(setter(!strip_option))]
```
Patrik-Stas, line 89: Since more fields are non-option than options, it would be cleaner to simply define `strip_option` on the field which use `Option`, rather than defining `strip_option` as default and disabling it on the fields it doesn't apply on.

Patrik-Stas (changes requested):

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: Changed.

Patrik-Stas (dismissed):

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -83,26 +83,31 @@ impl Drop for IndyWalletRecordIterator {
     }
 }
 
-#[derive(Clone, Debug, Default, Builder, Serialize, Deserialize)]
-#[builder(setter(into, strip_option), default)]
```
mirgee, line 87: We should decide some common properties for the builders derived within the codebase. IMO, `strip_options` and `default` significantly simplify ergonomics without loss of clarity and expressiveness, unlike `into` which one one hand has the benefit of not having to write an extra `.into()` (which is not a huge win) and obscures the fact that (potentially complex (but hopefully not - discipline required)) conversion is happening.

Fully standardizing the properties may require changes which I don't want to do in this PR, but we should also not make steps further away from the desired state.

mirgee (changes requested):

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: `into` was removed from default setter opts.

mirgee (approved):

aries/aries_vcx_core/src/wallet/indy/mod.rs
Patrik-Stas: @mirgee correct me if I am wrong, but perhaps you didn't necessarily mean to ask remove `into`, but rather open discussion? 
I personally don't have strong opinion on this, it's pretty cool, but I am also leaning towards not using these `into` magic. So I am rather aligned with the removal (and not using it elsewhere).

About `strip_options`:
- We are using `strip_options` on these builder and commonly elsewhere. I think this one is pretty contextual.

About `default`, actually good catch. @xprazak2 is there a use-case / rationale to define both `derive(Defautlt)` and `#[builder(default)]` for this?

Patrik-Stas (commented):
shared some thoughts in the thread about builder options/config

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: `derive(Default)` does not seem to be used here, I removed it also in other places where possible. There is a very good reason for using both `derive(Default)` and `#[builder(default)]`:


```rust
#[derive(TypedBuilder)]
pub struct Foo {
   #[builder(default)]
    bar: Bar
}

#[derive(TypedBuilder, Default)]
pub struct Bar {
  #[builder(default)]
  item: String
}

```

in this case, `Bar` needs `Default` so that builder knows how to build a default value when building `Foo`.

Patrik-Stas (changes requested):
left a comment

aries/aries_vcx_core/src/wallet/indy/mod.rs
Patrik-Stas: The usage of `default` currently still enables writing code we rather wouldn't allow. Eg:
```
let wallet_config = WalletConfig::builder().build();
```
This compiles, because `#[builder(field_defaults(default))]` applies also on String attributes, and therefore all of the fields become optional from builder point of few - which kinda defeats the purpose of `TypedBuilder`, which should give us guarantee in compile time that only semantically valid structures are created. Wallet config definitely needs to have properly set up values like `name`, `key` etc. 
It would be advantageous to have `None` default for optional arguments which are not typically supplied, like `rekey`, `rekey_derivation_method`.

On a side note, I would suggest not using `strip_option` for `wallet_type`, `storage_config`, `storage_credentials` as these are pretty important, so user better be explicit about wanting to keep them `None`.

xprazak2: I am happy to make the changes, but it makes me wonder what is actually the scope here. This started as a switch from one builder library to another which [should be possible with minimum changes, as they produce similar builder APIs](https://github.com/hyperledger/aries-vcx/pull/1093) and suddenly we are in the middle of discussion about introducing changes to how certain existing structs are built.

Removing `strip_option` seems to go directly against [previous note](https://github.com/hyperledger/aries-vcx/pull/1108#discussion_r1458733046) which sees it as an ergonomics improvement. So if these attributes are important, should we go from `Option<String>` to `String`?

Patrik-Stas: I understand where you are coming from and am also looking to get all the wallet stuff merged ;-) The builders pattern was introduced in the wallet domain in this+previous PR, and that's the subject of the discourse. So I wouldn't say it's out of scope. Though it would be nice if we found a way to facilitate faster process of the discussion.

But what I am really only "going after" now is mainly the default stuff, and assuring we do not enable building non-sensical structures - being main value proposition of `typed-builder`.

My comment on the `strip_option` is indeed a side note and less important. We can leave that out for separate discussion. Imo `strip_option` should be used case-by-case with nuance and I don't even know if you guys agree with that.

Patrik-Stas (approved):

PR #1122, opened 2024-02-06
refactor: use BaseWallet instead of a specific wallet type

xprazak2:
Replaces the direct usage of `IndySdkWallet` with `BaseWallet` trait.


aries/aries_vcx_core/src/wallet/base_wallet/mod.rs
```diff
@@ -1,76 +1,63 @@
...
-    async fn pack_message(
-        &self,
-        sender_vk: Option<Key>,
-        receiver_keys: Vec<Key>,
-        msg: &[u8],
-    ) -> VcxCoreResult<Vec<u8>>;
+    async fn configure_issuer(&self, key_seed: &str) -> VcxCoreResult<IssuerConfig>;
 
-    async fn unpack_message(&self, msg: &[u8]) -> VcxCoreResult<UnpackMessageOutput>;
+    async fn all(&self) -> VcxCoreResult<Box<dyn AllRecords + Send>>;
```
mirgee, line 41: Why custom trait instead of simply implementing an iterator? Why on `BaseWallet` instead of `RecordWallet`?

aries/aries_vcx_core/src/wallet/base_wallet/record.rs
```diff
@@ -28,3 +31,54 @@ impl Record {
...
+
+    pub fn category(&self) -> &Option<String> {
+        &self.category
+    }
+
+    pub fn tags(&self) -> &Option<RecordTags> {
+        &self.tags
+    }
+
+    #[cfg(feature = "vdrtools_wallet")]
```
mirgee, line 61: Should be implemented in the `indy` module.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -27,119 +57,91 @@ impl IndySdkWallet {
...
-    #[serde(skip_serializing_if = "Option::is_none")]
-    rekey: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    storage_credentials: Option<serde_json::Value>,
-    key_derivation_method: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    rekey_derivation_method: Option<String>,
-}
+#[async_trait]
+impl BaseWallet for IndySdkWallet {
```
mirgee, line 66: Why is it called `IndySdkWallet`, while the feature flag is called `vdrtools_wallet`? Doesn't it seem wrong? A newcomer would be rightfully confused.

aries/aries_vcx_core/src/wallet/indy/mod.rs
```diff
@@ -1,16 +1,46 @@
...
+    BaseWallet,
+};
 use crate::{errors::error::VcxCoreResult, WalletHandle};
 
 mod indy_did_wallet;
 mod indy_record_wallet;
-mod indy_tags;
-pub mod internal;
-pub mod wallet;
+pub(crate) mod indy_tags;
```
mirgee, line 16: Does not need to be made public (again).

aries/aries_vcx_core/src/wallet/indy/restore_wallet_configs.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub wallet_name: String,
+    pub wallet_key: String,
+    pub exported_wallet_path: String,
+    pub backup_key: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub wallet_key_derivation: Option<String>,
+}
+
+#[async_trait]
+impl ImportWallet for RestoreWalletConfigs {
```
mirgee, line 19: Why are we calling the same concept different names (restore, import)?

aries/aries_vcx_core/src/wallet/indy/wallet_credentials.rs
```diff
@@ -0,0 +1,13 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+struct WalletCredentials {
```
mirgee, line 4: Where is this used?

aries/aries_vcx_core/src/wallet/indy/wallet_config.rs
mirgee, line 1: As you probably know, there are many things wrong with this and `restore_wallet_configs.rs` files, but let's focus elswhere since we don't plan to support this implementation long-term.

aries/aries_vcx_core/src/wallet/indy/indy_tags.rs
mirgee, line 1: `IndyTags` don't need to be `pub(crate)`

mirgee (changes requested):

aries/aries_vcx_core/src/wallet/base_wallet/mod.rs
xprazak2: Because I need a custom abstraction for how different wallet implementations fetch all records. vdrtools wallet provides [WalletIterator](https://github.com/hyperledger/aries-vcx/blob/main/aries/misc/legacy/libvdrtools/indy-wallet/src/iterator.rs#L14-L38) which does not conform to `std::iter::Iterator` nor to `std::async_iter::AsyncIterator`. 

Moved to `RecordWallet`.

aries/aries_vcx_core/src/wallet/base_wallet/record.rs
xprazak2: Moved.

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: Having unified naming would be nice, any suggestions? 

However these are existing names and a rename could be done separately.

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: Changed.

aries/aries_vcx_core/src/wallet/indy/restore_wallet_configs.rs
xprazak2: The struct renamed to `ImportWalletConfigs`

aries/aries_vcx_core/src/wallet/indy/wallet_credentials.rs
xprazak2: Good catch, removed.

aries/aries_vcx_core/src/wallet/base_wallet/did_wallet.rs
```diff
@@ -0,0 +1,83 @@
...
+        sender_vk: Option<Key>,
+        receiver_keys: Vec<Key>,
+        msg: &[u8],
+    ) -> VcxCoreResult<Vec<u8>>;
+
+    async fn unpack_message(&self, msg: &[u8]) -> VcxCoreResult<UnpackMessageOutput>;
+}
+
+#[async_trait]
+impl DidWallet for Arc<dyn BaseWallet> {
```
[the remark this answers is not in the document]
mirgee: @xprazak2 ?

aries/aries_vcx_core/src/wallet/base_wallet/mod.rs
mirgee: I asked why it's implemented on `BaseWallet` instead of `RecordWallet`, I am not arguing for moving it (or not).

Also, I am sorry, but I don't see how "I need a custom abstraction for how different wallet implementations fetch all records" is explanation for creating a custom trait instead of implementing an iterator.

aries/aries_vcx_core/src/wallet/indy/mod.rs
mirgee: What suggestion do you have in mind? vdrtools vs. indy-sdk? Do *you* have an opinion?

aries/aries_vcx_core/src/wallet/base_wallet/did_wallet.rs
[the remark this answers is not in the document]
xprazak2: So that methods that take wallet as and argument do not need to change the argument type from `impl BaseWallet` to `Arc<dyn BaseWallet>`.

aries/aries_vcx_core/src/wallet/base_wallet/mod.rs
xprazak2: `RecordWallet` operates only on records, so it was implemented in `BaseWallet` because it fetches both records and keys. Even though keys are "key records" since everything is stored in the same table. Indy wallet fetches everything with a single method while askar has separate methods for keys and records. 

Instead of just implementing an iterator, a wallet specific trait was created because 
[getting all records](https://github.com/hyperledger/aries-vcx/blob/main/aries/misc/legacy/libvdrtools/indy-wallet/src/wallet.rs#L472) for indy gives me [WalletIterator](https://github.com/hyperledger/aries-vcx/blob/main/aries/misc/legacy/libvdrtools/indy-wallet/src/iterator.rs#L14-L38) which is async and I do not see a way how to implement a sync trait using async function.

I could implement async iterator but that requires bringing in additional dependency.

aries/aries_vcx_core/src/wallet/indy/mod.rs
xprazak2: I am asking about your suggestion for a new name since you brought it up.

We could rename the struct to `VdrtoolsWallet` or change to `IndyWallet` and change the flag to `indy_wallet` as well.

aries/aries_vcx_core/src/wallet/base_wallet/did_wallet.rs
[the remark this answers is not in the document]
xprazak2: So that methods can retain the `-> impl BaseWallet` return type.

xprazak2:
Things that will need to be addressed in the future:
* close/delete wallet
* import/export wallet

aries/aries_vcx_core/src/wallet/base_wallet/did_wallet.rs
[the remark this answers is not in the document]
xprazak2: Also, ergonomics. `Arc<dyn BaseWallet>` does not implement `BaseWallet` by default, so we would need to call `wallet.as_ref().some_wallet_method()` everywhere.

aries/aries_vcx_core/src/wallet/indy/indy_tags.rs
Patrik-Stas: @xprazak2 if this is still true, let's mininize the visibility.

aries/aries_vcx_core/src/wallet/indy/mod.rs
Patrik-Stas: I agree renaming/unifying can be done, but possibly not a concern of this PR. There was an inconsistency between indysdk/vdrtools naming already before. I'll leave to your judgement @xprazak2 to address this / or not / in this PR

Patrik-Stas (changes requested):
Left a review, I am mostly concerned about usage of trait objects

aries/aries_vcx_core/src/wallet/indy/indy_tags.rs
xprazak2: If I make them private, how can they be used in `indy_record_wallet.rs`?

aries/aries_vcx_core/src/wallet/base_wallet/did_wallet.rs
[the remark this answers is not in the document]
mirgee: I don't think these are good reasons to reimplement all these traits for `Arc<dyn BaseWallet>` even assuming that we decide to go with dynamic dispatch (which I agree we shouldn't).

aries/aries_vcx_core/src/wallet/indy/indy_tags.rs
mirgee: I am not sure, but I think what I meant when posting this comment *3 weeks ago* was just that the `(crate)` modifier was unnecessary as the module is private anyways.

aries/agents/rust/mediator/src/aries_agent/mod.rs
```diff
@@ -34,24 +33,24 @@ pub mod client;
 pub mod utils;
 
 #[derive(Clone)]
-pub struct Agent<T: BaseWallet, P: MediatorPersistence> {
-    wallet: Arc<T>,
+pub struct Agent<P: MediatorPersistence> {
```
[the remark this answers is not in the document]
Patrik-Stas: Libvcx is second class citizen for use - it shouldn't drive design of the core aries components - `aries_vcx` / `aries_vcx`. I think it should be possible to use concrete types there, and resort to trait objects in `libvcx`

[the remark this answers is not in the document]
xprazak2: I agree that library consumer should not drive the design of the core components, however the tricky part is that if `libvcx` uses trait objects (and therefore type erasure), how can it pass a concrete type when calling core components that require a specific type.

aries/aries_vcx_core/src/wallet/indy/indy_tags.rs
xprazak2: Changed.

xprazak2:
The dynamic dispatch is no longer used.

Patrik-Stas (approved):
LGTM ✅ Great job @xprazak2 this is good improvement

PR #1132, opened 2024-02-22
fix: push napi image using a simple command

xprazak2:


.github/workflows/push-napi-image.yml
Patrik-Stas, line 1: I think you are missing this step prior to publishing though

```
      - name: "Docker Login"
        uses: azure/docker-login@v1
        with:
          login-server: ${{ env.URL_DOCKER_REGISTRY }}
          username: $GITHUB_ACTOR
          password: ${{ secrets.GITHUB_TOKEN }}
```

Patrik-Stas (changes requested):

Patrik-Stas (approved):

PR #1139, opened 2024-02-29
feat(mediator): Containerise and package mediator

nain-F49FF806:


aries/agents/rust/mediator/.env
Patrik-Stas, line 1: Given the default dev configuration assumes that the database has certain configuration - eg. certain version of mysql, certain username and password, It would be nice to add small shell script to startup dependencies for local development - I guess for now that would by just the mysql. 
Then also add note in the `readme.md`

Patrik-Stas (changes requested):
Just a small note about dev onboarding/documentation

Patrik-Stas (dismissed):

nain-F49FF806:
@Patrik-Stas I have added further small improvements to the readme. The push seems to have automatically made the review stale. But the changes are small, you can take a glance.

PR #1142, opened 2024-03-02
feat(actions): add input in publish image action and default it to ghcr.io

### LINKED ISSUE #1140: Publish image CI action has undeclared dependency.
`.github/actions/publish-image` uses environment variable `$URL_DOCKER_REGISTRY` to decide which root registry to push to.   
If this is not set globally, or at some step before the action's invocation, it will fail to publish. 

This can be surprising, as the necessity of this environment variable is not mentioned anywhere. 

It would be better to have this as an input to the action, with a default value: 'ghcr.io'.

PanGan21:
- Adds `url-docker-registry` input in publish image github action with default value to `ghcr.io`
- Pass parameter where publish image action is called

Closes #1140

### DIFF
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -92,3 +92,4 @@ runs:
       with:
         docker-img: ${{ inputs.docker-img }}
         publish-version: ${{ inputs.branch-name }}
+        url-docker-registry: ${{ env.URL_DOCKER_REGISTRY }}
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -28,6 +28,10 @@ inputs:
   skip-build-optimization:
     description: "If set to true, will not try to load image from cache"
     default: false
+  url-docker-registry:
+    description: 'URL of the Docker registry to push the image to'
+    required: false
+    default: 'ghcr.io'
 
 runs:
   using: "composite"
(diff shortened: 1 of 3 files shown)


.github/actions/build-image/action.yml
```diff
@@ -92,3 +92,4 @@ runs:
       with:
         docker-img: ${{ inputs.docker-img }}
         publish-version: ${{ inputs.branch-name }}
+        url-docker-registry: ${{ env.URL_DOCKER_REGISTRY }}
```
nain-F49FF806, line 95: This seems like the right way to go. 
But note that now we will again have a surprising fail scenario if this environment variable is not set. 

Let's repeat the earlier process and add *url-docker-registry* as explicit input of this action, and then pass it down the chain.

nain-F49FF806 (changes requested):
Thank you for making the PR! This looks good. You have alleviated the main concern of #1140.
I have only a few minor additions/edits I would like to see before we merge it in. It should take you only a couple of minutes more. Please take a look through the review.

.github/actions/publish-image/action.yml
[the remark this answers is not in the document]
PanGan21: Updated!

.github/actions/build-image/action.yml
PanGan21: Done! Thanks a lot for the feedback! 
I still kept the same pattern, optional with default value. Let me know if that works!

nain-F49FF806 (approved):

nain-F49FF806:
LGTM. 

I'll let this be open until Monday to let other maintainers have a look. 
Thank you for the contribution!

Patrik-Stas (approved):

PR #1144, opened 2024-03-12
feat: add wallet migration to askar

xprazak2:


.github/workflows/main.yml
```diff
@@ -100,7 +100,8 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        wallet: ["vdrtools_wallet", "askar_wallet"]
+        main_wallet: ["main_vdrtools_wallet,vdrtools_wallet", "main_askar_wallet,askar_wallet"]
```
Patrik-Stas, line 103: Let's have a separate job for migration clippy check. This seems like slippery slope if we were to start combining with even more feature flag going forward. 

In fact, let's keep just `["vdrtools_wallet", "askar_wallet"]` here when clippy-ing workspace holistically.

.github/workflows/main.yml
```diff
@@ -399,7 +417,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        wallet: ["vdrtools_wallet", "askar_wallet"]
+        wallet: ["main_vdrtools_wallet,target_askar_wallet", "main_askar_wallet,target_askar_wallet"]
```
Patrik-Stas, line 420: Integration test do not know about migrations, do they? I am sure there was technical reason which required you introduce these feature flags here, but conceptually when running normal integration tests, there should be no notion of migrations.

aries/aries_vcx_core/src/wallet/base_wallet/mod.rs
```diff
@@ -123,6 +133,8 @@ mod tests {
 
         assert_eq!(new_key.key().len(), 32);
 
+        println!("first did data: {:?}", first_data);
```
Patrik-Stas, line 136: forgotten println

aries/wrappers/uniffi-aries-vcx/core/Cargo.toml
```diff
@@ -16,9 +16,11 @@ path = "uniffi-bindgen.rs"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["vdrtools_wallet"]
-vdrtools_wallet = ["aries_vcx/vdrtools_wallet"]
+default = ["main_vdrtools_wallet"]
 askar_wallet = ["aries_vcx/askar_wallet"]
+vdrtools_wallet = ["aries_vcx/vdrtools_wallet"]
+main_vdrtools_wallet = ["vdrtools_wallet"]
```
Patrik-Stas, line 22: Seems like feature alias? I'd rather keep it simple, just keep the original feature flag names.

Patrik-Stas (changes requested):
Some changes requested. Also we are now going to remove `libvcx_core` and `vcx_napi_rs`, so you can disable the CI for those components and that should simplify your life here. 
In fact feel free to open PRs to delete `libvcx_core` and `vcx_napi_rs` (+all nodejs stuff) so you won't have to deal with it at all.

.github/workflows/main.yml
xprazak2: Well, libvcx [knows](https://github.com/hyperledger/aries-vcx/pull/1144/files#diff-67003c7e81812e12857fcefa42000b0b3d953a8db1073404a173d7bd9888682fR275) about migrations and exposes the migration as a part of its public API.

The reason for the new feature flags is that a migration has source and destination wallet and we need to decide which wallet is to be considered 'main' so that the appropriate global variable is included. Anyway, this seems to be irrelevant now, removing the vcx_napi_rs and libvcx_core will definitely make things easier.

Patrik-Stas:
@xprazak2 needs some rebasing now, should be prolly easier now to simplify feature flags?

xprazak2:
Rebased, looks much cleaner now.

Patrik-Stas (approved):

Patrik-Stas:
well done 👍

Issue #1163, opened 2024-03-22
Enhance didcore `VerificationMethod` in-memory representation


Patrik-Stas:
The current implementation of `VerificationMethod` looks as follow:

```
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, TypedBuilder)]
#[serde(rename_all = "camelCase")]
pub struct VerificationMethod {
    id: DidUrl,
    controller: Did,
    #[serde(rename = "type")]
    verification_method_type: VerificationMethodType,
    #[serde(flatten)]
    public_key: PublicKeyField,
}
```

and 

```
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
#[serde(untagged)]
#[serde(deny_unknown_fields)]
pub enum PublicKeyField {
    #[serde(rename_all = "camelCase")]
    Multibase { public_key_multibase: String },
    #[serde(rename_all = "camelCase")]
    Jwk { public_key_jwk: JsonWebKey },
    #[serde(rename_all = "camelCase")]
    Base58 { public_key_base58: String },
    #[serde(rename_all = "camelCase")]
    Base64 { public_key_base64: String },
    #[serde(rename_all = "camelCase")]
    Hex { public_key_hex: String },
    #[serde(rename_all = "camelCase")]
    Pem { public_key_pem: String },
    #[serde(rename_all = "camelCase")]
    Pgp { public_key_pgp: String },
}

```

The current internal representation `VerificationMethod` maps 1:1 with how Verification Method object is described. This makes it de/serialization simple, given that it default serde de/serialization is sufficient. However, it's not data model one would naturally come up if this was designed if de/serialization wasn't concern, for following reasons:
  - it's doesn't follow Law of Demeter - encoded representation is concern of de/serialization, in-memory encoding serves here no purpose
  - it's not efficient - whenever we want to read raw key value from a Verification Method, we need to re-decode they key value
   - deserialization of `VerificationMethod` does not provide guarantee as to whether the key is correctly encoded - it only guarantees the value is a String. Decoding is indirectly verifier upon upon an attempt to read raw key value from `public_key`

That being said, we likely want to keep attached information about "how should the raw key be encoded"  upon serialization. Upon deserialization, we can set the value based on key encoding in the input.

Addressing this will be mainly matter of changes to `VerificationMethod` struct:
- getting rid of `PublicKeyField` struct in favor of storing raw bytes
- introducing flag signalling encoding to be used upon serialization
[...]

PR #1170, opened 2024-03-28
refactor: improve didcore VerificationMethod in-memory representation

xprazak2:
I followed the approach outlined in https://github.com/hyperledger/aries-vcx/issues/1163#issue-2202258364
 * removed `PublicKeyField` in favour of bytes
 * introduced a new flag that determines the serialization format
 * implemented custom serialization

I do not think that this is really an improvement since it has its own downsides:
* we now have 2 separate attributes to determine the format of a single value
* `VerificationMethod` is now responsible for implementing how to serialize other type - key should know how to serialize itself
* while bytes can represent various key types, having just raw bytes us not that useful for working with the key. Usually, those bytes are used to construct a type that has methods for key operations

So I think we might go from:

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
#[serde(untagged)]
#[serde(deny_unknown_fields)]
pub enum PublicKeyField {
    #[serde(rename_all = "camelCase")]
    Multibase { public_key_multibase: String },
    #[serde(rename_all = "camelCase")]
    Jwk { public_key_jwk: JsonWebKey },
    #[serde(rename_all = "camelCase")]
    Base58 { public_key_base58: String },
    #[serde(rename_all = "camelCase")]
    Base64 { public_key_base64: String },
    #[serde(rename_all = "camelCase")]
    Hex { public_key_hex: String },
    #[serde(rename_all = "camelCase")]
    Pem { public_key_pem: String },
    #[serde(rename_all = "camelCase")]
    Pgp { public_key_pgp: String },
}
```

to something like:

```rust
#[derive(Clone, Debug, PartialEq)]
pub enum PublicKeyField {
    Multibase(public_key::Key),
    Jwk (JsonWebKey),
    Base58(public_key::Key),
    Base64(public_key::Key),
    Hex(public_key::Key), 
    Pem(pem::Pem),
    Pgp(pgp::Pgp),
}

impl Serialize for PublicKeyField {...}
```

The key type is responsible for holding the key bytes and `PublicKeyField` knows how to de/serialize itself. Your thoughts?

### DIFF
--- a/did_core/did_doc/src/schema/legacy.rs
+++ b/did_core/did_doc/src/schema/legacy.rs
@@ -10,6 +10,8 @@ use crate::schema::{
     verification_method::{VerificationMethod, VerificationMethodType},
 };
 
+use super::verification_method::PublicKeyFormat;
+
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq, display_as_json::Display)]
 #[serde(deny_unknown_fields)]
 pub struct LegacyDidDoc {
@@ -162,8 +164,9 @@ fn construct_new_did_document(
                 id,
                 did.clone(),
                 VerificationMethodType::Ed25519VerificationKey2018,
+                PublicKeyFormat::Multibase(multibase::Base::Base58Btc),
             )
-            .add_public_key_multibase(fingerprint.clone())
+            .add_public_key(fingerprint.clone().as_bytes().to_vec())
             .build(),
         );
     }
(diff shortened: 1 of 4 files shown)


did_core/did_doc/src/schema/legacy.rs
```diff
@@ -162,8 +164,9 @@ fn construct_new_did_document(
                 id,
                 did.clone(),
                 VerificationMethodType::Ed25519VerificationKey2018,
+                PublicKeyFormat::Multibase(multibase::Base::Base58Btc),
```
xprazak2, line 167: Leaking the implementation details to the upper layer. Can we even reliably determine the key format at this point?

Patrik-Stas:
@xprazak2 thanks for the insights! You have good points.

> The key type is responsible for holding the key bytes and PublicKeyField knows how to de/serialize itself. Your thoughts?

I think this is good idea, lets do it!

JamesKEbert:
I think this should likely be closed at this point. @gmulhearn any thoughts?

PR #1172, opened 2024-04-04
refactor: move ledger to a separate crate

xprazak2:
Extracts ledger from aries_vcx_core and moves it to a separate crate.


aries/aries_vcx/Cargo.toml
```diff
@@ -42,6 +42,8 @@ messages = { path = "../messages" }
 diddoc_legacy = { path = "../misc/legacy/diddoc_legacy" }
 aries_vcx_core = { path = "../aries_vcx_core" }
 aries_vcx_wallet = { path = "../aries_vcx_wallet" }
+aries_vcx_ledger = { path = "../aries_vcx_ledger" }
+indy-vdr = { git = "https://github.com/hyperledger/indy-vdr.git", rev = "c143268", default-features = false, features = ["log"] }
```
Patrik-Stas, line 46: Why is `indy-vdr` introduced as direct `aries-vcx` dep?

aries/aries_vcx/src/errors/mapping_indyvdr.rs
```diff
@@ -0,0 +1,49 @@
+use indy_vdr::{
```
Patrik-Stas, line 1: It seems that this is reason why `indy_vdr` was introduced as dependency - but `aries_vcx` shouldn't know anything about the underlying ledger implementation.

Patrik-Stas: Doesn't the mapping for `VcxLedgerError` in the file below suffice?

aries/aries_vcx_core/src/anoncreds/credx_anoncreds/mod.rs
```diff
@@ -130,7 +130,7 @@ impl RecordWallet for WalletAdapter {
     async fn search_record(
         &self,
         category: RecordCategory,
-        search_filter: Option<SearchFilter>,
+        search_filter: Option<String>,
```
Patrik-Stas, line 133: At glance seem to be unrelated with the nature of the PR. Why was this changed?

aries/aries_vcx_ledger/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+name = "aries_vcx_ledger"
+version.workspace = true
+authors.workspace = true
+description.workspace = true
+license.workspace = true
+edition.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+vdr_proxy_ledger = ["dep:indy-vdr-proxy-client"]
```
Patrik-Stas, line 11: It would be ideal if we use `vdr_proxy_ledger`, we don't actually compile and bring in `indy-vdr` itself (as it brings/requires all the zmq stuff with it).

I suppose the reason why this can't be done now (having 2 separate feature `vdr_proxy_ledger` and `vdr_zmq_ledger`) is because even when you use proxy implementations, I guess we are still relying on some types defined indy `indy-vdr` (which carries the zmq stuff).

If you can confirm my suspition, there's potential refactoring task:
Refactor indy-vdr library such that it's broken down into
 - types crates
 - logic
 
 So that we can use the types, given that's necessary regardless of what transport we use, but not necessary require zmq library installed on the system if the user only intends to use `proxy` implementation.

aries/aries_vcx_wallet/src/wallet/base_wallet/search_filter.rs
```diff
@@ -1,20 +1,20 @@
-use std::fmt;
+// use std::fmt;
```
Patrik-Stas, line 1: was this intended in this PR?

aries/aries_vcx_wallet/src/wallet/indy/mod.rs
```diff
@@ -104,16 +104,9 @@ impl IndySdkWallet {
     async fn search(
         &self,
         category: RecordCategory,
-        search_filter: Option<SearchFilter>,
+        search_filter: Option<String>,
```
Patrik-Stas, line 107: same

Patrik-Stas (commented):
Left few comments, given these changes `aries_vcx_core` could be instead renamed to `aries_vcx_anoncreds` to keep consistent naming with these other extracted crates.

aries/aries_vcx/src/errors/mapping_indyvdr.rs
xprazak2: This file is the reason why `indy_vdr` is now a dependency - it was moved from `aries_vcx_core`. There used to be `VdrError` -> `AriesVcxCoreErrr` -> `AriesVcxError` conversion path but because `aries_vcx_core` is going away, we need to change to a direct `VdrError` -> `AriesVcxError` mapping. Alternatively, we could create an error variant in `VcxLedgerError` for each `VdrError`.

xprazak2: The mapping from `VcxLedgerError` currently does not suffice, because `VcxLedgerError` does not create its own variant for each `VdrError` but rather [wrapps it](https://github.com/hyperledger/aries-vcx/pull/1172/files#diff-e8ad81d5df90cb2b04e8295db885f1a06795ce78ee034fd6200824384bbb53a7R19)

xprazak2:
> Left few comments, given these changes `aries_vcx_core` could be instead renamed to `aries_vcx_anoncreds` to keep consistent naming with these other extracted crates.

That will be my next step.

aries/aries_vcx_wallet/src/wallet/indy/mod.rs
xprazak2: This is intentional, I can open it separately if you prefer.

aries/aries_vcx/src/errors/mapping_indyvdr.rs
Patrik-Stas: Lets do the changes necessary to shield off vcx from `indy-vdr`, be it creating/adjusting `VcxLedgerError`. There's not much code relying on specific ledger error kinds (afaik the only place where we directly deal with them is when we are creating credential definition, and we have some specific behaviour based on whether the same cred-definition already exists/not exist on ledger.

aries/aries_vcx_core/src/anoncreds/credx_anoncreds/mod.rs
Patrik-Stas: Ah I see, the 2 wallet implementations use different types for search and there's some feature flagging to put them both under `SearchFilter` type. 
Yeah we can address that in the future. Resolving

aries/aries_vcx/src/errors/mapping_indyvdr.rs
Patrik-Stas: We can merge after this is sorted

Patrik-Stas (approved):

PR #1178, opened 2024-04-12
feat: load wallet config from environment variables

nain-F49FF806:
This should give full control over mediator. 

Note: We could later support a dedicated config file format such as *yaml*/*toml*. Or command line arguments to pass config parameters. 

But simply with env variables, we still have sufficient control. 
One can pass the environment variables on the command line or declare them in a `.env` file. 
The `.env` file may be in the same directory as where mediator is executed, or a parent directory. [[dotenvy](https://docs.rs/dotenvy/latest/dotenvy/fn.dotenv.html)]


xprazak2 (approved):

nain-F49FF806:
@xprazak2 Thanks for the review!

PR #1184, opened 2024-05-14
fix: allow agent wallet re-creation

Zzocker:


gmulhearn (approved):
LGTM. It's a little bit invisible/non-obvious that the behaviour of `build_indy_wallet` is to "_idempotently_ create and open" the wallet.  But this error handling fits to that behaviour (ignore re-initializing the link secret).

Also as a note/warning: the aries-vcx-agent is not meant for production use. i believe it's primary use case currently is for testing against other aries implementations in the AATH. Hopefully this is somewhat obvious by the amount of dangerous `unwrap`s that are in these functions.

PR #1185, opened 2024-05-16
chore: updating dependabot to support gradle, docker, gha and cargo packages

rajpalc7:


JamesKEbert (approved):
LGTM. We can also refine further in the future if settings defined here are too frequent, etc.

PR #1213, opened 2024-06-05
[Update] Update to anoncreds-rs 0.2

gmulhearn-anonyome:
In order to migrate off of mirgee's fork and onto 0.2.0, we lose the following patches made on 0.2.0: https://github.com/hyperledger/anoncreds-rs/compare/v0.2.0...mirgee:anoncreds-rs:refactor/public-api-changes.

However we can get around those public APIs being missing by using into our `Convert` layer: we can switch between the anoncreds-rs definition of RevocationStatusList, and our own anoncreds_types definition of RevocationStatusList. These definitions are exactly the same, and JSON de/serialize the exact same way, however our version has the public APIs we need. So we can switch between theirs and ours when we need to. 

Note that it is not perfectly efficient, as we are converting to JSON Value as an intermediate type (as we do for some other anoncreds types too), but this could potentially be cleaned up if we fully move to using the real types from anoncreds-rs crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -315,13 +315,13 @@ dependencies = [
 
 [[package]]
 name = "anoncreds"
-version = "0.2.0-dev.7"
-source = "git+https://github.com/mirgee/anoncreds-rs.git?rev=a85d739#a85d739c10a90834dbd912508cbceba77b114b64"
+version = "0.2.0"
+source = "git+https://github.com/hyperledger/anoncreds-rs.git?tag=v0.2.0#394a1ce469ef4d6cacf404d882266d798e3e171c"
 dependencies = [
- "anoncreds-clsignatures 0.3.1",
+ "anoncreds-clsignatures 0.3.2",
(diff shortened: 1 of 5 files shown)


gmulhearn-anonyome:
@JamesKEbert thoughts on this? it gets us off the forked version of anoncreds-rs, but continues to lean into the potential longer-term issues with `anoncreds_types` that we discussed in #1212

JamesKEbert (approved):
LGTM, and aligns with our previous discussion. Thanks for making this PR!

JamesKEbert:
DCO check is broken as noted here--https://discord.com/channels/905194001349627914/1156961689971998782/1247613586638110894
I can take a look at the workaround tomorrow morning

gmulhearn:
@JamesKEbert ow yea yes please. didn't notice it was stuck 😅. i'll leave this PR unmerged for now

JamesKEbert (approved):
LGTM

PR #1214, opened 2024-06-05
fix: dco checks

JamesKEbert:
Fixes DCO checks due to dcoapp issue - https://github.com/dcoapp/app/issues/211
Uses recommend approach from Hyperledger - https://discord.com/channels/905194001349627914/1156961689971998782/1247613586638110894


gmulhearn:
thanks @JamesKEbert. it looks like the issue was fixed upstream right after you made this fix 😅. is it still necessary?

JamesKEbert:
I somehow managed to make this PR 7 minutes before the service was restored :man_facepalming:

PR #1217, opened 2024-06-06
[FIX] Reduce unnecessary required deps

gmulhearn-anonyome:
I noticed that after pulling in the newest version of aries-vcx into my project, that i started picking up vdrtools_wallet and vdrtools dependencies again.

This was mainly due to:
* aries_vcx_anoncreds had default features of credx & vdrtools_wallet
* aries_vcx depends on aries_vcx_anoncreds without disabling default features

there were also a few other unnecessary dependencies

solution:
* removed default features of aries_vcx_anoncreds. other core crates like aries_vcx_wallet do not have default features; the default is that everything is disabled. I prefer it this way, as default features often get forgotten and lead to things being unnecessarily packed in
* removed "vdrtools_wallet" & "askar_wallet" features of aries_vcx_anoncreds: it is not necessary for the crate to have these features. the crate only cares about "BaseWallet", not any specific BaseWallet implementations.
* found an error that would happen if aries_vcx was compiled AND aries_vcx_wallet/vdrtools_wallet was NOT enabled: mapping_wallet.rs would not see the VcxWalletError::IndyApiError branch (since VcxWalletError::IndyApiError is only a branch if aries_vcx_wallet/vdrtools_wallet is enabled)
* removed indy-vdr dependency from aries_vcx_wallet. it was unnecessary: only was used for base64. i replaced this with the normal base64 crate
* moved did_resolver_sov to be a dev dependency of aries_vcx
* moved aries_vcx_wallet to be a dev dep of did_resolver_sov

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ dependencies = [
  "actix-service",
  "actix-utils",
  "ahash",
- "base64 0.21.7",
+ "base64",
  "bitflags 2.4.0",
  "brotli",
  "bytes",
@@ -319,7 +319,7 @@ version = "0.2.0"
 source = "git+https://github.com/hyperledger/anoncreds-rs.git?tag=v0.2.0#394a1ce469ef4d6cacf404d882266d798e3e171c"
 dependencies = [
  "anoncreds-clsignatures 0.3.2",
- "base64 0.21.7",
+ "base64",
  "bitvec",
  "bs58 0.5.0",
  "chrono",
@@ -490,7 +490,7 @@ dependencies = [
  "async-channel",
  "async-trait",
  "backtrace",
- "base64 0.22.1",
+ "base64",
  "bs58 0.5.0",
  "chrono",
  "derive_builder",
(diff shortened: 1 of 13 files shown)


JamesKEbert (approved):
LGTM! Good reduction of unnecessary dependencies


> removed default features of aries_vcx_anoncreds. other core crates like aries_vcx_wallet do not have default features; the default is that everything is disabled. I prefer it this way, as default features often get forgotten and lead to things being unnecessarily packed in

How do you feel about us recommending the Askar feature / newer Shared Components features in like the README documentation? Reason why I ask is that I can see that having defaults can lead to unnecessary packages, but I do think it's a good idea to encourage new users to use the newer packages. Also happy to chat about this on the community call if you think it's an involved discussion or related to our other discussion

gmulhearn-anonyome:
@JamesKEbert yea i do agree. let's discuss in the call. for consistency i removed the default here, but potentially we could add defaults to all and be more careful about our internal usage.

PR #1221, opened 2024-06-10
[Update] Update AATH backchannel to build image 

gmulhearn-anonyome:
The main change here is to add a new CI/CD job to build and publish our backchannel image. then this image can be consumed in the AATH repo, rather than us having to synchronise code between here and AATH repo. The CI/CD should do the following:
* re-release/replace an image tagged with `main` whenever we push to main branch. the `main` tagged image will therefore be the "latest and greatest"
* publish an image tagged with some tag version whenever we create a tagged release. e.g. `0.64`. these are our "stable release" tag images

_Note that i haven't done github workflows before, so please carefully check that what i've done is sane... i've mostly tried to match some of the existing patterns in the workflows main yaml._

# Other Changes
* remove unnecessary Cargo.lock in the aath-backchannel (unused)
* fix the "version" returned from the AATH controller. this version string is what is used on the AATH reports. we will need to make sure to keep it in sync with the aries-vcx crate
* update some aath deps
* removed some old artifacts from the github CI (hashing dirs that don't exist, etc)
* update the seed creation approach. the previous approach only worked in some environments, as some times you would end up with an ENDORSER issuer_did vs a USER issuer_did ROLE. a USER issuer_did cannot write schemas, causing some tests to fail. changes ensure that we end up with an ENDORSER issuer_did.
* minor changes to the image. The previous image just straight up didn't work on newer aries-vcx code.

# Fixes for ACAPY compatibility
I've also made some changes to get a bit more interop with ACApy. particularly around the DIDExchange suite of interop tests.
* fix did peer key format
* fix did peer service key reference
* fix DID Peer 3 & 4 handling
* fix in-tolerant URL_SAFE decoding

new approach is tested upstream in the AATH: https://github.com/hyperledger/aries-agent-test-harness/pull/833


JamesKEbert:
[I poked Shel on Discord](https://discord.com/channels/905194001349627914/941708308459438121/1250118728256061531)--if we wanted we could add a second image that only gets build on release. It's a little bit more work (and could be a separate item if we want), but could be a good idea potentially.

gmulhearn:
@JamesKEbert  hey, this one is ready for review. I believe the updated CI/CD should let us support the different version testing style that you suggested. Pls see my MR overview.

tbh, we won't know _for sure_ that this is all working until we merge it in and start getting tagged versions of the image. If it fails for whatever reason, i will iterate on it quickly of course.

gmulhearn:
Forgive me for making indirectly related changes, but when testing acapy, i found some issues with our did:peer implementations. My latest commit should address them: https://github.com/hyperledger/aries-vcx/pull/1221/commits/a3c19840b7f1ee831337bdfefe77ca449f1bd894

Will test in acapy now.

gmulhearn:
ok with my latest changes, we get very close to successful RFC0023 tests with acapy. we fail on the final step when we try to decode their DIDDoc within the DIDExchange Request they sent us... we fail here because they are using legacy DIDDocs when we are expecting a modern diddoc structure.. i suspect this is related to acapy using unqualified DIDs in these requests. i'll see if there's a way to run the tests with acapy using qualified did:peers rather than unqualified

JamesKEbert:
The AATH CI pieces look good to me, even if we accidentally did something wrong with the CI I am A-okay with us doing followup fixes there :+1:

JamesKEbert:
As for the DID Exchange issues w/ACA-Py, the DIDDoc appears to be eventually resolved in /aries_vcx/src/protocols/did_exchange/state_machine/responder/response_sent/mod.rs `receive_request()`, which calls `let their_ddo = resolve_ddo_from_request(&resolver_registry, &request).await?;`, which attempts to map the attachment to a DIDDoc, but if that fails it will attempt to resolve it via the resolver registry (which the AATH agent has did:sov and did:peer registered, but since in the request the DID is unqualified the parser/resolver would fail I'd assume). 

The DIDDoc format corresponds to [RFC 0067](https://hyperledger.github.io/aries-rfcs/latest/features/0067-didcomm-diddoc-conventions/#example-domain-and-did-document), but that RFC hasn't been officially accepted/adopted, so I am unsure as to whether or not this is acceptable/expected/ideal behavior. From your POV, is this a sufficient DIDDoc inside of Aries VCX?

gmulhearn:
@JamesKEbert thanks for taking a look. yea i think we are doing what the DIDExchange spec defines. Essentially:
1. if the request contains a did_doc attachment, then decode that as the DIDDocument. If that decode fails then fail early (`.transpose()?`)
2. else if there was no did_doc, then try use the resolver to resolve the DID.

The RFC from my reading says that implementors should either:
1. have an unqualified DID and a did_doc attachment set in their request, or
2. have a qualified (resolvable) DID and do not attach a did_doc attachment

So in the case of doing pure did:peer:4 exchanges with ACApy, we should be good, as we will hit case 2 and be able to properly resolve the DID; i've also looked at acapy code to confirm this. However when acapy is using an unqualified DID and goes with case 1, then they will attach the weird old/legacy/aries version of a DIDDocument, which does not match the real spec: https://www.w3.org/TR/did-core/#did-document-properties

I had a look at how acapy handles this situation (i.e. acapy handling receiving and processing a DIDExchange request in case 2). The following happens:
1. The request is funnelled into acapy's [_extract_and_record_did_doc_info](https://github.com/hyperledger/aries-cloudagent-python/blob/5ad52c15d2f4f62db1678b22a7470776d78b36f5/aries_cloudagent/protocols/didexchange/v1_0/manager.py#L685) function.
2. This function extracts the DIDDoc as a raw JSON/dict (no conversion yet) and stores it in the wallet against the unqualified DID (all in raw JSON/dict)
3. Eventually when acapy needs to retrieve the doc, i believe it is retrieved via the [LegacyPeerDIDResolver](https://github.com/hyperledger/aries-cloudagent-python/blob/5ad52c15d2f4f62db1678b22a7470776d78b36f5/aries_cloudagent/resolver/default/legacy_peer.py#L353) implementation, which fetches the doc from the wallet.
4. After fetching the raw dict/JSON doc from the wallet, it will pass it thru a [LegacyDocCorrections](https://github.com/hyperledger/aries-cloudagent-python/blob/5ad52c15d2f4f62db1678b22a7470776d78b36f5/aries_cloudagent/resolver/default/legacy_peer.py#L27) implementation to convert the weird/legacy/aries DIDDoc into a real DIDDoc.

long story short; acapy will proactively handle both by converting the legacy format into the correct format.

So i suppose aries-vcx should consider whether it should handle this legacy doc and try convert it? i think it probably should... I'll have a go at implementing this now. if it's too much, then i'll create another ticket

gmulhearn:
Supporting DIDExchange v1.1 will also improve our interop with acapy: https://github.com/hyperledger/aries-vcx/issues/1228

JamesKEbert (approved):
LGTM! Good work on this!

Issue #1226, opened 2024-06-13
[FIX] DIDExchange handlers should do more signature checking


gmulhearn:
I'm finding that currently, the `DIDExchange` requester and responder impls are not checking some cryptographic parts of the DIDExchange flow. Particularly:

1. ~when the inviter/responder _receives a request_, (`DidExchangeResponder::<ResponseSent>::receive_request`), then it should check that the wire-level sender of the request is from a key that matches a key of the DIDDoc inside the request. The spec says to do this:~
> ~After receiving the exchange request, the responder evaluates the provided DID and DID Doc according to the DID Method Spec.~
> ~The responder should check the information presented with the keys used in the wire-level message transmission to ensure they match.~

~I don't believe that ACAPy does this~

**EDIT: descoping this, as the community does not seem to find this case necessary: https://github.com/hyperledger/aries-rfcs/issues/717#issuecomment-1069768955**

2. when the invitee/requester _receives a response_ (`DidExchangeRequester::receive_response`), then it should check that DIDDoc or the DIDRotate attachment is signed JWS, and is signed with the original inviter key. The spec is clear on this:
> When the requester receives the response message, they will decrypt the authenticated envelope which confirms the source's authenticity. After decryption validation, the signature on the did_doc~attach or did_rotate~attach MUST be validated, if present. The key used in the signature MUST match the key used in the invitation. After attachment signature validation, they will update their wallet with the new information, and use that information in sending the complete message.

ACApy seems to do this

gmulhearn:
Related to signed did_rotate attachments: https://github.com/hyperledger/aries-vcx/issues/1228

gmulhearn-anonyome:
Part 2 is fixed in #1232 . I'm still questioning the importance of part 1, as other agents (acapy, credo) don't seem to do this AFAIK

JamesKEbert:
I don't know that (1) is required as the signatures are not serving any real purpose as I understand it, also (as mentioned [here](https://github.com/hyperledger/aries-rfcs/issues/717#issuecomment-1069768955)), for instance.

gmulhearn-anonyome:
yea agreed. ok i'm going to descope it from this issue

gmulhearn:
Closed by #1232

PR #1230, opened 2024-06-18
[Feature] Support DIDExchange 1.1 #1228

### LINKED ISSUE #1228: [Feature] Support DIDExchange 1.1
The current implementation of DIDExchange is for DIDExchange v1.0, we should update this to support v1.1. The new feature of this minor version is the did_rotate attachment, which removes the need for qualified DID users to need to attach a full did_doc in the final response, instead they can use the did_rotate attachment: https://github.com/hyperledger/aries-rfcs/blob/main/features/0023-did-exchange/README.md#version-change-log.

Another reason to support v1.1, is such that other agents know to use qualified DIDs when interacting with us, improving our AATH score. For instance, if ACApy receives an OOB invitation from us with handshake protocol of didexchange/1.0, then ACApy will default to using unqualified DIDs: https://aca-py.org/latest/features/QualifiedDIDs/#did-exchange.

# Open questions
* should we bother still supporting v1.0?

gmulhearn:
This PR closes #1228 by implementing DIDExchange v1.1 support. It also does alot of DIDExchange/OOB/AATH fixes of bugs i found along the way whilst testing this implementation in the AATH setup against itself and against acapy.

# DIDExchange v1.1 Implementation
## Messages
The previous message structure for DIDExchange were scoped to v1.0. This implementation creates v1.1 message structures to sit alongside v1.0. The approach taken largely follows the guide laid out in the [README](https://github.com/hyperledger/aries-vcx/blob/main/aries/messages/README.md). This is also the approach i did previously whilst creating issue-credential-v2 & present-proof-v2 message structures.

Where i took a bit of creative liberty though, was with creating a _common_ module called `v1_x`. I felt this was necessary, as all messages in DIDExchange v1.1 & v1.0 are the exact same, except for the RESPONSE message (which simply adds the did_rotate field). So the v1_x module contains the common definitions for the `complete, problemreport & request` message structures, and some utility code for the dealing with the varying `response` messages. 

Again, taking some creative liberty, i added an invisible `version` marker field on these common messages:
```
    #[serde(skip, default = "DidExchangeTypeV1::new_v1_1")]
    pub(crate) version: DidExchangeTypeV1,
```
This field is invisible to the consumer and to the JSON serialized format, however it's a run-time marker that allows us to convert from and to the `AriesMessage` type without losing the context of which DIDExchange version the message is. The unit tests ensure this is as expected. I modified the `AriesMessage` and `DidExchange` de/serializers to deal with setting and using this field properly.

I also played with another way of doing this, which used a compile-time marker for V1.0 vs V1.1 messages, it looked like this:
```rust
// in v1_x module
pub struct RequestContent<MinorVer> {
    ...,
    _marker: PhantomData<MinorVer>
}

// in v1_1 module
pub type RequestV1_1Content = RequestContent<DidExchangeTypeV1_1>;

// in v1_0 module
pub type RequestV1_0Content = RequestContent<DidExchangeTypeV1_0>;
```
this way, at compile time, there is 2 distinct types for v1.1 vs v1.0 messages. However, this got a bit ugly when actually using the implementation, as we need to create `AnyXYZ` wrappers for the two variants of each message, and pass those around everywhere, and then all consumers would need to switch/match on the different variants. I preferred the run-time way, but open to the other way.

# Implementation (State machine)
The implementation did not have to change dramatically. essentially:
* take `AnyResponse` instead of `Response`
* requester impl will always try to request with a 1.1 request (and deal with it if the responder sends back a 1.0 response)
* responder will decide what to do based on whether the request is v1.0 or v1.1. If v1.1, then we attach a signed did_rotate attachment, else default is to attach a signed did_doc.
* The requester does not currently verify the signature of the response did-rotate. this is a detail delegated to this ticket: https://github.com/hyperledger/aries-vcx/issues/1226
* add passthru of `label` - bcus why not
* some code cleanup

# Fixes for AATH
* some of the backchannel APIs were not including enough data for the AATH controller to be happy, i've added some of those fields in.
* ACApy compatibility: made it so that responses and complete messages will keep re-using the pthid. I don't necessarily agree with this being the correct usage of pthid, however it's what other agents do and what DIDExchange expects, and it doesn't hurt our implementation. some discussion: https://github.com/hyperledger/aries-rfcs/issues/817
* changed our DIDExchange did:peer:4 creation helper it put a Verification Method (VM) reference in the keyAgreement of the DIDDoc, rather than directly embedding the VM. this is more "normal" from my understanding
* Fixed up a bunch of `didcomm_utils` which were a bit slack
  * `resolve_base58_key_agreement` not being clear on it's purpose of fetching ed25519 base58 keys
  * `resolve_base58_key_agreement` not checking that the keys it fetched were ed25519
  * exact same case for `get_ed25519_base58_routing_keys` ^
  * implemented `get_ed25519_base58_recipient_keys`, which resolves keys referenced in a given diddoc service
* make it so that the EncryptionEnvelope first tries `get_ed25519_base58_recipient_keys` (yields more accurate result than just "get the first key agreement key" as was previously done and is now used as a backup)
* TODO comments where i think there should be future work
* add support for mimetype of `text/string`. i'm not sure if it's valid, but acapy uses it, and so does the DIDExchange spec.
* also made it so that the attachment type has a `MaybeKnown` for the `MimeType` when deserializing. Without this, it would fail to deserialize the whole attachment/message if the mimetype is unrecognised (i ran into this with acapy's `text/string`.)
* made it so that `dereference_key` DIDDoc utility will check ALL places that an embedded VerificationMethod could be. Rather than just the "normal" case where it is embedded in `verificationMethod` field.
* added support for understanding the verificationMethodType of `MultiKey`, which seems to be the preferred type in the did:peer spec. Also expanded the `.public_key` method to know how to resolve this VM type into a `Key`.
* a bunch of fixes to the did:peer2 implementation:
  * `append_encoded_key_segments` was adding some encoded VMs twice
  * append keys into the verificationMethod field, and add a reference in the appropriate type (authentication, assertion, etc). rather than embedded (less "normal")
  * add an acapy sample and spec defined sample into the test suite
  * allow extra JSON fields to be included in the encoded service (e.g. for `recipientKeys` (acapy uses this correctly))
  * the VM key IDs in resolved DIDDocs were being given an ID based on the "prefixless fingerprint" (`to_did_url_reference`), i'm not sure where this idea came from, but the spec defined way is to just index the keys `#key-N` (`nth_key_did_url_reference`).


gmulhearn:
Some notes on AATH testing:
Aries VCX (ACME) <-> ACApy (BOB):
RFC0023:
* T003: success: behave -D Acme=http://0.0.0.0:9020 -D Bob=http://0.0.0.0:9030 -t @T003-RFC0023
* others: fail

RFC0793:
* T001-RFC0793.1 fails (we don't support unqualified)
* T001-RFC0793.2 fails (DIDDOc key agreement issue)
* T001-RFC0793.3 success (did:peer4)
* T002-RFC0793 fails (we accept only qualified DIDs)
* T003-RFC0793.1 fails (key agreement issue)
* T003-RFC0793.2 fails (we accept send did:peer4)
   * Note that if we force the scenario to use pure did:peer4, then it succeeds.

aries/aries_vcx/src/protocols/did_exchange/state_machine/requester/request_sent/mod.rs
```diff
@@ -25,10 +29,11 @@ impl DidExchangeRequester<RequestSent> {
     pub async fn construct_request(
         resolver_registry: Arc<ResolverRegistry>,
         invitation_id: Option<String>,
+        // TODO - label pass thru
```
gmulhearn, line 32: TODO

aries/aries_vcx/src/protocols/did_exchange/state_machine/requester/request_sent/mod.rs
```diff
@@ -77,17 +88,20 @@ impl DidExchangeRequester<RequestSent> {
                 state: self,
             });
         }
+        // TODO - process differently depending on version
```
gmulhearn, line 91: TODO

gmulhearn: actually, that's a job for https://github.com/hyperledger/aries-vcx/issues/1226

JamesKEbert:
Awesome stuff! Been slowly working through the PR review 👍

> requester impl will always try to request with a 1.1 request (and deal with it if the responder sends back a 1.0 response)

We might be able to make the default be based on what is specified in a given OOB invite handshake protocols array. If they list 1.0, then that's what we use. If they specify 1.1 then we use 1.1. If this is using did exchange not via oob (not sure if we are/care to support) then defaulting to 1.1 makes sense. 

Although I suppose it should work fine to use just 1.1 by default and then downgrade to 1.0 if that's what gets returned in the response. And that might be better as that means we try and use 1.1 always first.

gmulhearn-anonyome:
@JamesKEbert good point, i had a go at this in my latest commit: https://github.com/hyperledger/aries-vcx/pull/1230/commits/5f5133d57b2a1a8207002a6afd0edc42efcf48d8

let me know what you think, i can always revert.

>  If this is using did exchange not via oob (not sure if we are/care to support)
This is the case when we send a request based off an implicit invitation (e.g. a public resolvable DID with a service)

aries/aries_vcx/src/utils/encryption_envelope.rs
```diff
@@ -80,6 +89,8 @@ impl EncryptionEnvelope {
         wallet: &impl BaseWallet,
         data: &[u8],
         sender_vk: Option<&str>,
+        // TODO - why not have encryption envelope take typed [Key]s, and enforce they are
+        // KeyType::Ed25519
```
JamesKEbert, line 93: Agreed

aries/aries_vcx/src/protocols/did_exchange/state_machine/responder/response_sent/mod.rs
```diff
@@ -28,35 +32,57 @@ impl DidExchangeResponder<ResponseSent> {
...
                 //       the key the request used as recipient_vk to anoncrypt the request
                 //       So argument "invitation_key" should be required
-                ddo_attachment_unsigned
-            }
-            Some(invitation_key) => {
-                // TODO: this must happen only if we rotate DID; We currently do that always
-                //       can skip signing if we don't rotate did document (unique p2p invitations
-                //       with peer DIDs)
-                jws_sign_attach(ddo_attachment_unsigned, invitation_key, wallet).await?
+                unsigned_attachment
```
JamesKEbert, line 63: Given the following from the DID Exchange 1.1 RFC:
> The `did_rotate~attach` attribute is optional, but SHOULD be included if the did attribute is resolvable and the `did_doc~attach` is not included. The value is the Base64url encoded DID, and signed with the key used in the invitation.

How is an implicit invitation (just having a resolvable DID that a `requester` uses to sends a DIDExchange Request to the `responder`) compatible with the DID Exchange 1.1? Do we essentially just ignore the SHOULD value here since there is no invitation? :thinking:

JamesKEbert (commented):
Awesome stuff!
I definitely agree with the choice to have a common module for all the messages, as it's really just an extra field and does not deserve an entire new full gambit of protocol folder/messages/etc.

I have some concerns about the VCX message structures though, in particular relating to versions and invisible markers on Aries messages--I'd much rather we store the full version with all of the messages. Cause I can also agree that having multiple versions and needing to create `AnyXYZ` wrappers is gross. Chatting about it on the community call might be a good idea

aries/aries_vcx/src/protocols/did_exchange/state_machine/responder/response_sent/mod.rs
gmulhearn-anonyome: IMO, it'd be good if we always require an `invitation_key` to be passed in. Even in the case of the implicit invitation, i believe that ideally the responder will sign the did_rotate with the service/recipient key associated with their public DID that received the request. 

^ This is what ACApy does when it receives a request as a result of an implicit invitation; the response did_doc/did_rotate is still signed

gmulhearn-anonyome: > IMO, it'd be good if we always require an invitation_key to be passed in

i didn't change that here, as it was already like that, but maybe i should just change `invitation_key` to be required (not Option)

JamesKEbert: Yeah that'd make sense to me
> i didn't change that here, as it was already like that, but maybe i should just change invitation_key to be required (not Option)

Didn't mean to bring up unrelated stuff necessarily, but I noticed it next to the changes in this PR. I think it'd be appropriate to require it.

gmulhearn-anonyome:
As discussed on call, i'm changing the approach to NOT have the invisible version marker inside the message. instead we create AnyXYZ wrappers, which have the minor version metadata externally in the enum variants

aries/aries_vcx/src/protocols/did_exchange/state_machine/responder/response_sent/mod.rs
gmulhearn-anonyome: ok nvm i've moved this to a new issue. it requires a bit more work than this PR should handle IMO: https://github.com/hyperledger/aries-vcx/issues/1245

JamesKEbert: Makes sense to me :+1:

JamesKEbert (approved):
Good to go from my POV :+1:

PR #1232, opened 2024-06-24
[FIX] DIDExchange handlers should do more signature checking #1226

### LINKED ISSUE #1226: [FIX] DIDExchange handlers should do more signature checking

gmulhearn-anonyome:
branched off #1230 
diff: https://github.com/anonyome/aries-vcx/compare/gm/1228-did-exch-1_1...anonyome:aries-vcx:gm/1226-didexchange-signature-checks

fixes part 2 of #1226.

adds JWS signature verification when receiving a response. The new approach mostly matches ACApy's implementation found here: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/didexchange/v1_0/manager.py#L942.

It is a relatively aggressive approach where it will fail if neither DIDDoc attachments nor DIDRotate attachment have signatures.


gmulhearn-anonyome:
TODO - verify still working in Acapy aath

gmulhearn-anonyome:
@JamesKEbert this is ready for review after #1230 is merged

gmulhearn-anonyome:
alright this should be good to go

aries/aries_vcx/src/protocols/did_exchange/state_machine/helpers.rs
```diff
@@ -180,51 +180,111 @@ pub(crate) fn assemble_did_rotate_attachment(did: &Did) -> Attachment {
...
+                "Attachment JWS signature was not correctly base64Url encoded.",
+            )
+        })?;
+
+    let res = wallet
+        .verify(expected_signer, &sign_input, &signature)
+        .await?;
+
+    Ok(res)
 }
```
JamesKEbert, line 288: Would it be appropriate to have a test case for this? Or is that adding a test for test's sake?

aries/aries_vcx/src/protocols/did_exchange/mod.rs
```diff
@@ -103,11 +106,25 @@ pub async fn resolve_enc_key_from_invitation(
                 "resolve_enc_key_from_invitation >> Resolved did document {}",
                 output.did_document
             );
-            let key = resolve_first_key_agreement(&output.did_document)?;
```
JamesKEbert, line 106: Do you think `resolve_first_key_agreement()` is still a valid/useful function with this change? Just curious, no strong opinion from my side

aries/aries_vcx/src/protocols/did_exchange/state_machine/requester/request_sent/mod.rs
```diff
@@ -134,3 +130,96 @@ impl DidExchangeRequester<RequestSent> {
...
+            ..
+        } = resolver_registry.resolve(did, &Default::default()).await?;
+        return Ok(did_doc);
+    }
+
+    // default to error
+    Err(AriesVcxError::from_msg(
+        AriesVcxErrorKind::InvalidInput,
+        "DIDExchange response could not be verified. No DIDDoc nor DIDRotate was attached.",
+    ))
```
JamesKEbert, line 224: Do we have a strategy in Aries VCX as how to abort a protocol and send a problem report? As I would think this would be an appropriate time to create a problem report.

JamesKEbert: I recognize that's somewhat outside the scope of this PR, but it came to mind while reviewing

aries/aries_vcx/src/protocols/did_exchange/state_machine/helpers.rs
JamesKEbert: I suppose this might be covered under the `did_exchange_test_with_invalid_rotation_signature` test case. :thinking:

JamesKEbert (dismissed):
Looks good to me--good additions! I left some minor thoughts, but that may not need to hold up merging

aries/aries_vcx/src/protocols/did_exchange/mod.rs
gmulhearn: hmm yea it's a tough one. Ideally, DIDDocs that we resolve will have:
* a service with a recipient key

and then we just just use that recipient key as the peer's "encryption key".

However it is possible that some DIDs may not be perfect, and they will have a service, but no recipient keys, in which case, i think it's a decent backup solution to use the first key agreement key (`resolve_first_key_agreement`).

The above is the `resolve_enc_key_from_did_doc` after these changes. I don't think it's perfect, but i think this might be a situation where we need to handle random interop edge cases as they reveal themselves over time... atleast we know it interops with ourselves and with ACApy

aries/aries_vcx/src/protocols/did_exchange/state_machine/requester/request_sent/mod.rs
gmulhearn: Hmmm, so this protocol state machine is a little different to some of the other state machines in aries-vcx (this one is the "most modern"), but the handler has a `.fail` method and `.receive_problem_report` method which creates and processes problem report messages: https://github.com/hyperledger/aries-vcx/blob/main/aries/aries_vcx/src/protocols/did_exchange/state_machine/mod.rs#L37

I think the thinking was that, you get returned a TransitionError, you should process the error and decide if it's a fatal error or not; and subsequently decide if you want to create a problem report with `.fail`, or just try again

aries/aries_vcx/src/protocols/did_exchange/state_machine/helpers.rs
gmulhearn: Yeah the did_exchange_tests should cover a lot of this code in the happy and negative path. But i just added some better direct tests, testing self-fulfilling-ly and also testing against an acapy test vector i recorded

aries/aries_vcx/src/protocols/did_exchange/state_machine/requester/request_sent/mod.rs
gmulhearn: it isn't being tested or demonstrated anywhere though.. so i think in general it's been a bit neglected

JamesKEbert: Got it. I'd say that perhaps it's something we should look at at some point, but would be outside the scope of this PR.

JamesKEbert (approved):
LGTM

PR #1249, opened 2024-07-03
(refactor) small refactor to the new DIDExchange AnyXYZ wrappers

gmulhearn:
I started these changes mostly wanting to remove `Deserialize` trait from the `AnyXYZ` wrappers. As IMO trying to deserialize JSON back into AnyXYZ is dangerous; if both V1_0 & V1_1 variants are the same, then the serialized format is the same, so when deserializing `serde` cannot tell which variant it is, so it just picks the first one (misleading).

Also whilst i was here... i realised the Any wrappers were duplicating themselves alot, so i made a generic enum for it

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4888,9 +4888,9 @@ dependencies = [
 
 [[package]]
 name = "serde"
-version = "1.0.195"
+version = "1.0.203"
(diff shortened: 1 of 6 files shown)


JamesKEbert (approved):
Yeah this is great--I also applaud the code de-duplication you did here

PR #1262, opened 2024-07-09
Removed unused code and suppressed warnings/errors

alberto-instnt:
- Added `#[allow(dead_code)]`.
- Removed `TryGetIndex` trait code since its not being used.


gmulhearn:
Nice! thanks. as we've discussed on discord, usually deadcode should be investigated, however since this is "legacy" code to be deleted in the future, this is fine

gmulhearn (approved):

PR #1276, opened 2024-07-23
chore: adjust dependabot config

JamesKEbert:
Adjust our dependabot PR config to remove the legacy crates and mobile directories (until they are solidified or not in a demo state), added missing cargo crates, and adjusted the frequency to 'monthly' so that it's not too frequent, but still provides value in keeping dependencies up to date (hopefully).


gmulhearn (approved):
v nice

PR #1281, opened 2024-08-08
feat: add additional formats for parsing and outputting Out Of Band Invitations

JamesKEbert:
Adds the ability to parse from and output to additional formats for the `OutOfBandReceiver`:
- JSON
- Base64URL
- URL

Additionally, adds parsing from the same formats to the `OutOfBandSender`, as new invites should always be created as `Invitation` or `AriesMessage` or with a builder. Consequently I've also removed the unused `from_string()` method from `OutOfBandSender`.

Adds tests for these methods (some repetition between the files, so I can see if I can consolidate if there's too much repeated during review). 

Does not include behavior for handling shortened URLs. I will create a followup GH issue for this item.

### DIFF
--- a/aries/aries_vcx/src/handlers/out_of_band/receiver.rs
+++ b/aries/aries_vcx/src/handlers/out_of_band/receiver.rs
@@ -41,22 +41,28 @@ impl OutOfBandReceiver {
         }
     }
 
-    pub fn from_json_string(oob_json: &str) -> VcxResult<Self> {
+    pub fn create_from_json_encoded_oob(oob_json: &str) -> VcxResult<Self> {
         Ok(Self {
-            oob: from_json_string(oob_json)?,
+            oob: extract_encoded_invitation_from_json_string(oob_json)?,
         })
     }
 
-    pub fn from_base64_url(base64_url_encoded_oob: &str) -> VcxResult<Self> {
+    fn create_from_base64_url_encoded_oob(base64_url_encoded_oob: &str) -> VcxResult<Self> {
         Ok(Self {
-            oob: from_json_string(&from_base64_url(base64_url_encoded_oob)?)?,
+            oob: extract_encoded_invitation_from_json_string(
+                &extract_encoded_invitation_from_base64_url(base64_url_encoded_oob)?,
+            )?,
         })
     }
 
-    pub fn from_url(oob_url_string: &str) -> VcxResult<Self> {
+    pub fn create_from_url_encoded_oob(oob_url_string: &str) -> VcxResult<Self> {
         // TODO - URL Shortening
         Ok(Self {
-            oob: from_json_string(&from_base64_url(&from_url(oob_url_string)?)?)?,
+            oob: extract_encoded_invitation_from_json_string(
+                &extract_encoded_invitation_from_base64_url(&extract_encoded_invitation_from_url(
+                    oob_url_string,
+                )?)?,
+            )?,
         })
     }
 
@@ -80,37 +86,37 @@ impl OutOfBandReceiver {
         }
     }
 
-    pub fn to_aries_message(&self) -> AriesMessage {
+    pub fn invitation_to_aries_message(&self) -> AriesMessage {
         self.oob.clone().into()
     }
 
-    pub fn to_json_string(&self) -> String {
-        self.to_aries_message().to_string()
+    pub fn invitation_to_json_string(&self) -> String {
+        self.invitation_to_aries_message().to_string()
     }
 
-    pub fn to_base64_url(&self) -> String {
-        BASE64_URL_SAFE.encode(self.to_json_string())
+    fn invitation_to_base64_url(&self) -> String {
+        BASE64_URL_SAFE.encode(self.invitation_to_json_string())
     }
 
-    pub fn to_url(&self, domain_path: &str) -> VcxResult<Url> {
+    pub fn invitation_to_url(&self, domain_path: &str) -> VcxResult<Url> {
         let mut oob_url = Url::parse(domain_path)?;
-        let oob_query = "oob=".to_owned() + &self.to_base64_url();
+        let oob_query = "oob=".to_owned() + &self.invitation_to_base64_url();
         oob_url.set_query(Some(&oob_query));
         Ok(oob_url)
     }
 }
 
-fn from_json_string(oob_json: &str) -> VcxResult<Invitation> {
+fn extract_encoded_invitation_from_json_string(oob_json: &str) -> VcxResult<Invitation> {
     Ok(serde_json::from_str(oob_json)?)
 }
 
-fn from_base64_url(base64_url_encoded_oob: &str) -> VcxResult<String> {
+fn extract_encoded_invitation_from_base64_url(base64_url_encoded_oob: &str) -> VcxResult<String> {
     Ok(String::from_utf8(
         URL_SAFE_LENIENT.decode(base64_url_encoded_oob)?,
     )?)
 }
 
-fn from_url(oob_url_string: &str) -> VcxResult<String> {
+fn extract_encoded_invitation_from_url(oob_url_string: &str) -> VcxResult<String> {
     let oob_url = Url::parse(oob_url_string)?;
     let (_oob_query, base64_url_encoded_oob) = oob_url
         .query_pairs()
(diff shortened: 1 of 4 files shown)


aries/aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -62,11 +84,45 @@ impl OutOfBandReceiver {
...
     }
+
+    pub fn to_base64_url(&self) -> String {
+        BASE64_URL_SAFE.encode(self.to_json_string())
+    }
+
+    pub fn to_url(&self, domain_path: &str) -> VcxResult<Url> {
+        let mut oob_url = Url::parse(domain_path)?;
+        let oob_query = "oob=".to_owned() + &self.to_base64_url();
+        oob_url.set_query(Some(&oob_query));
```
gmulhearn, line 98: nit; could you utilize the Url methods to add the query pair, rather than trying to assemble the pair yourself? e.g. `url.query_pairs_mut().append_pair("oob", self.base64)`

aries/aries_vcx/src/handlers/out_of_band/sender.rs
```diff
@@ -128,10 +134,19 @@ impl OutOfBandSender {
...
-    pub fn from_string(oob_data: &str) -> VcxResult<Self> {
-        Ok(Self {
-            oob: serde_json::from_str(oob_data)?,
-        })
+    pub fn to_json_string(&self) -> String {
+        self.to_aries_message().to_string()
+    }
+
+    pub fn to_base64_url(&self) -> String {
+        BASE64_URL_SAFE.encode(self.to_json_string())
```
gmulhearn, line 142: question; does base64 padding matter? it seems your encoding it _with_ padding, could result in a `=` at the end. If you don't want padding you can use `BASE64_URL_SAFE_NO_PAD`

aries/aries_vcx/src/handlers/out_of_band/sender.rs
```diff
@@ -128,10 +134,19 @@ impl OutOfBandSender {
...
+    }
+
+    pub fn to_url(&self, domain_path: &str) -> VcxResult<Url> {
+        let mut oob_url = Url::parse(domain_path)?;
+        let oob_query = "oob=".to_owned() + &self.to_base64_url();
+        oob_url.set_query(Some(&oob_query));
```
gmulhearn, line 148: same nit here; could use query_pairs_mut().append_pair

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -62,11 +84,45 @@ impl OutOfBandReceiver {
...
+    }
+}
+
+fn from_json_string(oob_json: &str) -> VcxResult<Invitation> {
+    Ok(serde_json::from_str(oob_json)?)
+}
+
+fn from_base64_url(base64_url_encoded_oob: &str) -> VcxResult<String> {
+    Ok(String::from_utf8(
+        URL_SAFE_LENIENT.decode(base64_url_encoded_oob)?,
```
gmulhearn, line 109: nice using the lenient engine. i was getting ready to nitpick it if you hadn't ;)

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -62,11 +84,45 @@ impl OutOfBandReceiver {
...
+    )?)
+}
+
+fn from_url(oob_url_string: &str) -> VcxResult<String> {
```
gmulhearn, line 113: nit; and general nit; IMO some of these function names are not obvious on what they are returning, i didn't know until i read into them. for instance, could you consider calling this fn something like `extract_encoded_invitation_from_url`.

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
JamesKEbert: Ah dang, I switched the parsing to use the query pairs, but didn't for the output to. Nice catch, that'll be a lot cleaner

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -38,6 +41,25 @@ impl OutOfBandReceiver {
         }
     }
 
+    pub fn from_json_string(oob_json: &str) -> VcxResult<Self> {
+        Ok(Self {
+            oob: from_json_string(oob_json)?,
+        })
+    }
+
+    pub fn from_base64_url(base64_url_encoded_oob: &str) -> VcxResult<Self> {
```
gmulhearn, line 50: same nit about naming; wasn't immediately sure what the input of this function was expected to be. i.e. something like `from_base64_url_encoded_oob` might be more specific.

also general question; will this API get much usage from ppl? i imagine most ppl are interested in using a "from invitation URL" or "from invitation message" API.

aries/aries_vcx/src/handlers/out_of_band/sender.rs
JamesKEbert: For parsing, we have to be able to handle with or without padding, but for encoding, it's not specified in RFC 0434 OOB (while it is in 0017 attachments) as far as I can see. The reason I switched to encoding _with_ padding was because the example base64url encoded invitation has a `=` at the end, indicating it was padded. I could take a look at the other frameworks to see what their implementations do here. Perhaps a clarification could be warranted to the RFC?

aries/aries_vcx/src/handlers/out_of_band/sender.rs
```diff
@@ -128,10 +134,19 @@ impl OutOfBandSender {
         self.oob.clone().into()
     }
 
-    pub fn from_string(oob_data: &str) -> VcxResult<Self> {
-        Ok(Self {
-            oob: serde_json::from_str(oob_data)?,
-        })
+    pub fn to_json_string(&self) -> String {
```
gmulhearn, line 137: nit here; maybe could be renamed. As a consumer, i might initially think this is serializing the `OutOfBandSender` struct into JSON (such that i can restore the `OutOfBandSender` later). i.e. `OutOfBandSender::to_xyz`. but it's actually serializing and returning the inner aries message

i realise the same nit can probably be made about the `to_aries_message` API on this struct... Altho the return type of `AriesMessage` makes it a lil more obvious.

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
JamesKEbert: Ahh I agree--I would have preferred to keep them as methods on the Receiver, but since they return the initial instance they can't call each other (which is needed here), so I made them separate functions, but that resulted in some lost context. I'll adjust the function names to be more descriptive. :+1:

aries/aries_vcx/src/handlers/out_of_band/sender.rs
gmulhearn: similar comment to the receiver; do we think ppl want an API which returns this? or would they be happy with just "give me the invitation message" + "give my the invitation encoded as an OOB invitation"

gmulhearn: no worries from me then; the spec gives example with `=` as you said

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
gmulhearn: ahh yea this context makes more sense

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
JamesKEbert: > also general question; will this API get much usage from ppl? i imagine most ppl are interested in using a "from invitation URL" or "from invitation message" API.

I don't necessarily expect so, the most common being from URL / Invitation JSON. The first example I can think of now is if someone does like a custom deep linking implementation (so it being closed source/specific to their org/app), which could be useful to parse the base64url from. But in that same example they could just implement the base64url parsing, so it's kinda mute.

Would your thought here be to remove it from the public API, making it private for usage only by the other methods, for simplicity sake?

aries/aries_vcx/src/handlers/out_of_band/sender.rs
JamesKEbert: Ahh yeah that's a fair confusion. I'll adjust the name here too. Perhaps `invitation_to_json_string()`, and maybe also `invitation_to_aries_message()`?

gmulhearn: nice sounds good

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
gmulhearn: ye i think that'd be my preference

aries/aries_vcx/src/handlers/out_of_band/sender.rs
JamesKEbert: Can confirm that ACA-Py encodes with padding :+1:

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
JamesKEbert: Unfortunately, it appears `query_pairs_mut()` uses the `application/x-www-form-urlencoded` syntax, which breaks when using padded base64url, as queries are strung together using `&` to separate parameters and `=` between the key and the actual value, so it throws `%3D` on the end instead of the base64url `=`. 

So I don't think we can actual do that, even though I like it better. Recommendations for how to do what I am doing now more cleanly @gmulhearn?

aries/aries_vcx/src/handlers/out_of_band/sender.rs
JamesKEbert: This did remind me to add a test to show we can parse with/without padding for base64url :+1:

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
gmulhearn: IMO that sounds like how it _should_ be encoded 😅, and the Aries RFC is incorrect for giving an example with `=` padding at the end of the URL, as that seems like an invalid usage/placement of `=` in a URL string. I would think that it'd work both ways for encoding/decoding in rust though; i.e. when you do `.query_pairs().find(|(name, _value)| name == "oob")`, it would decode `%3D` back into `=` for the extracted query value string.

But i suppose other aries implementation might take the RFC example at face value and then error out when trying to parse `%3D` (this is also another reason i wish the RFC had set a requirement on _unpadded_ base64url). If you're curious, you could look at impls (credo, acapy) to see how they handle extraction of the `oob` query param to see if they would deal with the URL encoded padding `%3D`. 

End rant; if we wanna keep the `=` at the end for interop with other aries agents, then i think how you had it is fine

JamesKEbert: I would posit that given the following from the Base64URL encoding referenced in the Aries RFC (https://datatracker.ietf.org/doc/html/rfc4648#section-5):
> The pad character "=" is typically percent-encoded when used in an
   URI [[9](https://datatracker.ietf.org/doc/html/rfc4648#ref-9)], but if the data length is known implicitly, this can be
   avoided by skipping the padding; see [section 3.2](https://datatracker.ietf.org/doc/html/rfc4648#section-3.2).

That this indicates that an agent _SHOULD_ be able to handle either no padding (which seems maybe inappropriate here given the above text and that our data length is not known implicitly due to variable invitation length) or by percent encoding, and that if they don't it is their failing, not VCX's.

Given that logic, I am going to update this PR to support encoding to with this padding and percent encoding it. I'll also probably open a PR to the Aries RFCs to adjust the example invitation (we'll see what happens there lol)

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
```diff
@@ -38,6 +41,33 @@ impl OutOfBandReceiver {
         }
     }
 
+    pub fn create_from_json_encoded_oob(oob_json: &str) -> VcxResult<Self> {
+        Ok(Self {
+            oob: extract_encoded_invitation_from_json_string(oob_json)?,
+        })
+    }
+
+    // This is here to help facilitate testing
```
gmulhearn, line 50: what's the point in keeping this method if it's only usage is in a test?

gmulhearn (approved):
LGTM, just that one remaining comment

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
JamesKEbert: I thought it was important to be able to validate whether we can handle padded and unpadded, but on reconsideration I should just make the test do that via the URL, not via an API we don't need. 
Thanks for helping me think through that lol, I'll update by the morning 👍

aries/aries_vcx/src/handlers/out_of_band/receiver.rs
JamesKEbert: As per Stephen Curran's guidance in https://github.com/hyperledger/aries-rfcs/pull/849, I think we're good to do unpadded as we discussed on the Community Call yesterday. :+1:

gmulhearn (approved):

PR #1283, opened 2024-08-15
(fix) fix aath publishing

gmulhearn:
Something changed last week where the default permissions for GITHUB_TOKEN was demoted. This meant our package publishing stopped working. This is why our backchannel image hasn't been updating.

After scavenging thru hyperledger discord, adding explicit permissions seems to fix the issue. 

related discord thread: https://discord.com/channels/905194001349627914/905205991073804308/1272872860494467093

related docs: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -146,9 +146,12 @@ jobs:
 
   # todo: move to indy-vdr repo
   build-docker-vdrproxy:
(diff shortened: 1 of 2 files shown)


JamesKEbert (approved):
LGTM

PR #1286, opened 2024-09-03
iOS project and script bindings

alberto-instnt:
This PR includes the following:
- Separation  of iOS and Android projects.
- Added sh script for iOS bindings.
- iOS project.
- Vendored zmq and openssl.
- Dynamic dependencies for vendored zmq.
- Readme adjustments to new structure


JamesKEbert:
Some awesome work, but I think that this needs another iteration before merging. I've uploaded the changes to a branch under alberto-ios-android-folders to preserve the changes for reference.

Issue #1293, opened 2024-09-26
Update indy-vdr version (time crate related issue)


gmulhearn:
The version of indy-vdr we are using has a dependency on `time` crate which is pinned to a version that has issues with rust 1.80+. We should update indy-vdr to a newer version where that bad time crate version is not pinned (0.3.36+). 

See here: https://github.com/hyperledger/indy-vdr/issues/324

Unfortunately the current release of indy-vdr (0.4.3) does not have this change within it yet, so we may have to continue using git hashes (pre-releases) depending on how urgently we want this.

swcurran:
FYI — the delay in getting the Indy VDR update (and Askar) is the pinning of Rust for the react-native/Android wrapper requirements — getting in the updates needed so that there is a combination of dependencies that will compile. We hope to have that resolved Real Soon Now — bumping the NDK and Rust versions and anything else that is needed. If anyone has experience with Android+ReactNative wrappers for Rust, please jump in to help.

PR #1296, opened 2024-10-02
[chore] rust dep updates (including indy-vdr)

gmulhearn:
partially relates to #1293 , but does not fix it yet (since the fix isn't in 0.4.3).

did some cleaning by moving common git-based crate dependencies to our workspace, so that the used version is globally controlled.

also did a general `cargo update` to push up our lockfile


JamesKEbert (approved):
LGTM

PR #1297, opened 2024-10-02
Support JWK public key VM processing into Key

### LINKED ISSUE #1289: Support key resolution of JWK-based DID VerificationMethods (publicKeyJwk)
Currently we do not support the ability to translate a VerificationMethod with a JWK-based key (publicKeyJwk field) into public key bytes (and transitively into a `public_key::Key`). This is a bottleneck in allowing us to use JWK DIDDoc verificationmethods for crypto (e.g. checking a signature from a JWK VM) 

area to be improved:
https://github.com/hyperledger/aries-vcx/blob/main/did_core/did_doc/src/schema/verification_method/public_key.rs#L46

gmulhearn:
Closes #1289 

Decided to use askar-crypto for processing JWK in simplest way possible, did not want to try process the keys myself, and couldn't find another smaller crate that was suitable.

For now I've made this feature optional via `jwk` feature-flag which a consumer can optionally enable. Didn't want to force a requirement on askar-crypto for a relatively small crate (public_key).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4331,6 +4331,7 @@ dependencies = [
 name = "public_key"
 version = "0.1.0"
 dependencies = [
+ "askar-crypto",
  "base64 0.22.1",
  "bs58",
  "multibase",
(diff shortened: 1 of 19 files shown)


JamesKEbert (dismissed):
LGTM :+1:

JamesKEbert:
Apparently there's some minor conflicts present on this one

PR #1308, opened 2024-12-04
(feat) Implement AnoncredsLedgerRead for cheqd #1306

### LINKED ISSUE #1306: [feat] Implement AnoncredsLedgerRead for cheqd
https://docs.cheqd.io/product/advanced/anoncreds

Implement the `AnoncredsLedgerRead` trait using the cheqd method in aries_vcx_ledger, as a feature. Should utilize & expand the resolver established in #1305 . System testing may be hard without a paired registrar - perhaps some sample testnet data can be found

gmulhearn:
Closes #1306 

## Additions
* Adds a Cheqd implementation of the `AnoncredsLedgerRead` trait to the `aries_vcx_ledger` crate (hidden behind a feature flag) `CheqdAnoncredsLedgerRead`. This implementation uses cheqd resource resolution, and is largely inspired by credo-ts https://github.com/openwallet-foundation/credo-ts/blob/main/packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
* Adds a `resolve_resource` method to the cheqd DID resolver, allowing DLRs to be resolved directly or via a name+type+time? DLR query. The implementation is largely inspired by credo-ts https://github.com/openwallet-foundation/credo-ts/blob/main/packages/cheqd/src/dids/CheqdDidResolver.ts#L69 . Note that the implementation for resolving DLR by name+type+time? query is not very efficient - it involves resolves the full set of DLR metadatas for a DID (collection) and then filtering in memory. This is a current limitation of the cheqd gRPC as pagination is not functioning as expected - this will need to be fixed once the gRPC is updated, however for now we ignore pagination (like credo)
* Adds a `MultiLedgerAnoncredsRead` implementation of the `AnoncredsLedgerRead` trait. This is similar to the DID resolver registry, it allows you to add multiple `AnoncredsLedgerRead` implementations and automatically switch between them depending on the requested schema/cred-def/rev-reg ID.
* add new attachment types for anoncreds attachments in issue-cred-v2 & present-proof-v2

## Refactors
* add a method to the anoncredsledgerread trait, `get_rev_status_list`, which directly queries for the revocation status list - this is now the preferred method, instead of `get_rev_reg_delta_json` and then leaving the delta -> status list conversion elsewhere.  
* All holder logic has been refactored to use `get_rev_status_list` instead of `get_rev_reg_delta_json`, however issuer & verifier is still a bit tied to it (issue moved to #1309 )
* (holder) parts of the `Anoncreds` trait refactored to take in a statuslist instead of delta, this allows some junk & conversions to be cut from the implementation: https://github.com/hyperledger/aries-vcx/pull/1308/files#diff-0c817a2ecd6c1ff01c6fc49b9c82add42df79bdbebc03a864fdf6d2566fc5aa1R936
* Adjusted DidUrl parsing to support/understand percent encoded characters, and to return the queries in their decoded form (matches behaviour of crates like `url`).

## Testing
* manual testing against credo-ts is successful (including revocation) (after this PR https://github.com/openwallet-foundation/credo-ts/pull/2126)
* manual testing against acapy is successful (including revocation) ( on this branch https://github.com/cheqd/aries-cloudagent-python/tree/acapy-cheqd )
* Integration tests for cheqd resources & anoncreds which query known resources/vectors
* unit testing for edge cases & multiledgeranoncreds


gmulhearn:
TODO - put cheqd vcx ledger sys tests behind feature flag

aries/aries_vcx_ledger/tests/test_cheqd.rs
```diff
@@ -0,0 +1,218 @@
+use std::sync::Arc;
```
gmulhearn, line 1: TODO - put cheqd vcx ledger sys tests behind feature flag

aries/aries_vcx_ledger/src/ledger/multi_ledger.rs
```diff
@@ -0,0 +1,670 @@
...
+use did_parser_nom::Did;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+use super::base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerSupport};
+use crate::errors::error::{VcxLedgerError, VcxLedgerResult};
+
+/// Struct to aggregate multiple [AnoncredsLedgerRead] implementations into a single
+/// [AnoncredsLedgerRead]. The child [AnoncredsLedgerRead] implementations are
+/// utiliized depending on whether or not they support resolution of the given object ID
```
gmulhearn, line 24: utiliized - typo

aries/aries_vcx_ledger/src/ledger/multi_ledger.rs
```diff
@@ -0,0 +1,670 @@
...
+    use super::*;
+
+    mock! {
+        #[derive(Debug)]
+        pub Reader {}
+        #[async_trait]
+        impl AnoncredsLedgerRead for Reader {
+            type RevocationRegistryDefinitionAdditionalMetadata = Value;
+
+            fn get_schema<'life0,'life1,'life2,'async_trait>(&'life0 self,schema_id: &'life1 SchemaId,submitter_did:Option< &'life2 Did> ,) ->  ::core::pin::Pin<Box<dyn ::core::future::Future<Output = VcxLedgerResult<Schema> > + ::core::marker::Send+'async_trait> >where 'life0:'async_trait,'life1:'async_trait,'life2:'async_trait,Self:'async_trait;
```
gmulhearn, line 291: sorry for ugliness in this test code! - sometimes `mock!` macro has trouble when combined with the `async_trait` macro, so the easiest fix is to insert the _fully expanded_ `async_trait` contents. 

i.e. if it was written like this:
```rust
#[async_trait]
impl AnoncredsLedgerRead for Reader
```
then it would expand into what you're seeing below.

aries/aries_vcx_ledger/src/ledger/base_ledger.rs
```diff
@@ -85,16 +93,35 @@ pub trait AnoncredsLedgerRead: Debug + Send + Sync {
...
+    /// Also returns any trait-specific additional metadata for the rev reg.
     async fn get_rev_reg_def_json(
         &self,
         rev_reg_id: &RevocationRegistryDefinitionId,
-    ) -> VcxLedgerResult<RevocationRegistryDefinition>;
+    ) -> VcxLedgerResult<(
+        RevocationRegistryDefinition,
+        Self::RevocationRegistryDefinitionAdditionalMetadata,
+    )>;
     async fn get_rev_reg_delta_json(
```
gmulhearn, line 106: add deprecated flag

aries/aries_vcx_ledger/src/ledger/multi_ledger.rs
```diff
@@ -0,0 +1,670 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+use super::base_ledger::{AnoncredsLedgerRead, AnoncredsLedgerSupport};
+use crate::errors::error::{VcxLedgerError, VcxLedgerResult};
+
+/// Struct to aggregate multiple [AnoncredsLedgerRead] implementations into a single
+/// [AnoncredsLedgerRead]. The child [AnoncredsLedgerRead] implementations are
+/// utiliized depending on whether or not they support resolution of the given object ID
+/// (e.g. based on the DID Method).
```
gmulhearn, line 25: make note about support for multi unqualified DID resolvers

aries/aries_vcx_ledger/src/ledger/base_ledger.rs
gmulhearn: @JamesKEbert i added this flag in my recent commit, you'll see i had to add `#[allow(deprecated)]` in all places where it's still being used - this makes it a bit clearer where it's still being used

aries/aries_vcx_ledger/src/ledger/arc.rs
gmulhearn, line 1: consumers could write this wrapper themselves (i was about to do it as a consumer @anonyome ), but figured it could go here too. If you think it's overkill for the vcx repo i'm happy to remove.

gmulhearn:
@JamesKEbert pls forgive me for some rust sins i committed in latest commit: https://github.com/hyperledger/aries-vcx/pull/1308/commits/06f9afd40f86523a10754d675555c6c44d1a5c44

particularly in `aries/aries_vcx/src/protocols/issuance/holder/state_machine.rs`, there is a several lack of typing, so everything is JSON strings and passed around in tuples. This needs a good re-evaluation, as it's way behind the rest of the VCX codebase in terms of type safety

aries/aries_vcx_ledger/src/ledger/arc.rs
JamesKEbert: If I understand correctly, the benefit of this wrapper is so that the Arc<AnoncredsLedgerRead> == AnoncredsLedgerRead, which is not possible via `.as_ref()` or otherwise (since the methods are all uniform in just needing a reference)? I'm wasn't able to immediately glean _why_ the AnoncredsLedgerRead would be in an Arc to begin with, but perhaps I just need to keep looking at the PR code lol.

JamesKEbert:
>  @JamesKEbert pls forgive me for some rust sins i committed in latest commit: https://github.com/hyperledger/aries-vcx/commit/06f9afd40f86523a10754d675555c6c44d1a5c44
> 
> particularly in aries/aries_vcx/src/protocols/issuance/holder/state_machine.rs, there is a several lack of typing, so everything is JSON strings and passed around in tuples. This needs a good re-evaluation, as it's way behind the rest of the VCX codebase in terms of type safety

Oh yeah that is gross, but I approve of your approach in this commit given it's already not good and IMO should be addressed in a separate PR entirely anyways.

JamesKEbert (approved):
Really great stuff--LOTS in here! Great to see further implementation on the Cheqd integration. Love the addition of the multi ledger read and the cleanup to use status lists.

Thanks for your hard work here @gmulhearn!

aries/aries_vcx_ledger/src/ledger/arc.rs
gmulhearn: Ye, it's a thin wrapper to make `Arc<T> : LedgerTrait` if `T: LedgerTrait`. `Arc<T>` isn't really being used in this PR, however i was running into cases as a consumer of it where i wanted this functionality. My particular use case was essentially:
* construct an `IndyVdrLedgerRead` instance
* wrap it in `Arc` so that i can
* pass 1 reference (clone the arc reference) of the instance into my `MultiLedger.register_reader(..)`, and
* pass 1 other reference into my VCX `DidSovResolver`

Perhaps i could've just constructed the `IndyVdrLedgerRead` instance twice, however i wasn't sure if that was safe/optimal (e.g. would this mean i'd have 2 zmq instances/managers running?!). i wasn't confident in these things, so sharing the reference with Arc felt safest (at the cost of a wee bit of performance - but VCX isn't forcing the consumer to make that decision)

aries/aries_vcx_ledger/src/ledger/base_ledger.rs
```diff
@@ -73,8 +73,16 @@ pub trait IndyLedgerWrite: Debug + Send + Sync {
...
 
+/// Trait for reading anoncreds-related objects from some ledger/s.
 #[async_trait]
 pub trait AnoncredsLedgerRead: Debug + Send + Sync {
+    /// Opaque additional metadata associated with retrieving a revocation registry definition.
+    /// Returned as part of `get_rev_reg_def_json`, and optionally passed into
+    /// `get_rev_status_list`. Depending on the ledger anoncreds-method, this metadata may be
+    /// used in the subsequent revocation status list fetch as an optimization (e.g. to save an
+    /// additional ledger call).
+    type RevocationRegistryDefinitionAdditionalMetadata: Send + Sync;
```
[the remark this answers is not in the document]
gmulhearn: ye ended up needing this across indy & cheqd implementations.

Essentially this metadata contains just a littleee bit of extra context that can save the `get_rev_status_list` implementation from having to re-fetch the revocationregsitrydefinition  from the ledger. Both cheqd & indyvdr needed something different to be able to save making the extra call, them needing something different is why i made it into an associated trait type.

Yea it did complicate the multiledger typing a little, as the multiledger has to adapt the different types into a single type - JSON serde was the most straight forward way to do so. But i don't think it's necessarily a problem, as there's an emphasis that this type is opaque; i.e. the consumer is not meant to try use the data being return, it's just meant to come out (of `get_rev_reg_delta_json`) and then get passed back in (into `get_rev_status_list`). Sorta like a pagination token; it's just some string blob that shouldn't really mean anything to the consumer, they're just meant to pass it back in and the service behind the scenes will understand.

aries/aries_vcx_ledger/src/ledger/multi_ledger.rs
gmulhearn: good call; done

aries/aries_vcx_ledger/src/ledger/arc.rs
JamesKEbert: Yeah that's a definitely fair approach IMO. In terms of it being in the repo I think that's fine as I could see it being useful for another consumer, and it's not overly large either. Thanks for the explanation :+1:

aries/aries_vcx_ledger/src/ledger/base_ledger.rs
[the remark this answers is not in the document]
JamesKEbert: That's about what I thought, thank you for running through it. Seems sufficient to me :+1:

JamesKEbert (approved):
LGTM

PR #1310, opened 2024-12-11
Simplify DIDComm unpack(), use `Key` instead of `String`

JamesKEbert:
# Refactors
- Simplifies the DIDComm unpack methods, as there previously used to be separate methods for authcrypt and anoncrypt, however under the hood they both would do authcrypt and anoncrypt unpacking, so the only difference really was that authcrypt would compare whether the sender verkey matched a supplied expected sender verkey (and provide an incorrect impression that only one strategy was being utilized). So I combined them and provided an optional arg for the expected sender verkey check. 
- Additionally, this PR adjusts the usage of Keys to be passed everywhere by `Key` instead of as `String` to provide better typing and access to the methods on the Key object. I also added a method to do a match against a given KeyType that will return an error if it is not valid to reduce the overhead of validation needed in a variety of places. Coincidentally this change removes a couple of spots where we go `Key` -> `String` -> `Key` (which is gross).
- Unpacked message keys are now assumed to be of KeyType ED25519.


aries/aries_vcx/src/utils/didcomm_utils.rs
```diff
@@ -50,23 +50,13 @@ pub fn resolve_ed25519_base58_key_agreement(did_document: &DidDocument) -> VcxRe
...
-                format!("Cannot resolve key agreement as an Ed25519 key: {vm:?}"),
-            ))
-        }
-    }
-
-    Ok(vm.public_key()?.base58())
+    Ok(key.validate_key_type(KeyType::Ed25519)?.to_owned())
 }
 
 pub fn get_ed25519_base58_routing_keys(
```
gmulhearn, line 56: this function name isn't accurate anymore, atleast the "base58" part

aries/aries_vcx/src/utils/didcomm_utils.rs
```diff
@@ -36,7 +36,7 @@ pub(crate) fn resolve_service_key_to_typed_key(
 
 /// Resolves the first ed25519 base58 public key (a.k.a. verkey) within the DIDDocuments key
 /// agreement keys. Useful for resolving keys that can be used for packing DIDCommV1 messages.
-pub fn resolve_ed25519_base58_key_agreement(did_document: &DidDocument) -> VcxResult<String> {
+pub fn resolve_ed25519_base58_key_agreement(did_document: &DidDocument) -> VcxResult<Key> {
```
gmulhearn, line 39: same comment about base58 funct name

aries/aries_vcx/src/utils/didcomm_utils.rs
```diff
@@ -78,7 +68,12 @@ pub fn get_ed25519_base58_routing_keys(
         let pub_key = resolve_service_key_to_typed_key(key, their_did_doc)?;
 
         if pub_key.key_type() == &KeyType::Ed25519 {
-            naked_routing_keys.push(pub_key.base58());
+            naked_routing_keys.push(pub_key);
```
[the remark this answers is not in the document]
JamesKEbert: Ahh fair, I think I was going to but got side tracked :man_facepalming:

aries/aries_vcx/src/utils/didcomm_utils.rs
JamesKEbert: Ooh good catch

aries/aries_vcx/src/utils/encryption_envelope.rs
```diff
@@ -88,17 +97,24 @@ impl EncryptionEnvelope {
...
         wallet: &impl BaseWallet,
         data: &[u8],
-        sender_vk: Option<&str>,
-        // TODO - why not have encryption envelope take typed [Key]s, and enforce they are
-        // KeyType::Ed25519
-        recipient_key: String,
-        routing_keys: Vec<String>,
+        sender_vk: Option<Key>,
+        recipient_key: Key,
+        routing_keys: Vec<Key>,
```
gmulhearn, line 102: nit; can all or any of these be passed by reference instead of taking ownership? e.g. `Option<&Key>`, `&Key`, `&[Key]`

gmulhearn: general comment from around the codebase, IMO it's best to not make consumers give up ownership unless we need it

JamesKEbert: I improved this somewhat--ended up being challenging to make many improvements here as the underlying code getting called for packing and unpacking is the did_wallet trait and its implementations, and I didn't really want to go down that rabbit hole tonight (especially as it may end up going all the way down to Askar itself). Also tried to use [u8] where possible, but same story as above.

PR #1313, opened 2024-12-13
(fix) Return IndyVdr based CredentialDefinition schemaId as full schemaId (not sequence number)

gmulhearn:
indy ledgers will return creddef.schemaId as seqNo sometimes (see related https://github.com/openwallet-foundation/acapy/issues/220 older acapy discussion), this is not a valid anoncreds schemaId, and is not consumer friendly.

This fix adds an extra ledger call (if necessary) to convert from schema seqNo to schemaId by resolving the raw txn and extracting the data. This matches acapy & credo's strategy:

credo-ts:
* https://github.com/openwallet-foundation/credo-ts/blob/main/packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts#L274
* https://github.com/openwallet-foundation/credo-ts/blob/main/packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts#L981

acapy:
* i believe acapy does still have the behaviour i'm trying to resolve here, but acapy does support resolving schema by it's seqNo: https://github.com/openwallet-foundation/acapy/blob/main/acapy_agent/ledger/indy_vdr.py#L496

### DIFF
--- a/aries/aries_vcx_ledger/src/ledger/indy_vdr_ledger.rs
+++ b/aries/aries_vcx_ledger/src/ledger/indy_vdr_ledger.rs
@@ -529,7 +529,7 @@ where
             .response_parser
             .parse_get_cred_def_response(&response, None)?;
 
-        // extract a map seqNo -> schemaId if required
+        // extract and map seqNo -> schemaId if required
         let IndyVdrCredentialDefinition::CredentialDefinitionV1(mut cred_def) = cred_def;
         if let Ok(seq_no) = cred_def.schema_id.0.parse::<i32>() {
             cred_def.schema_id = self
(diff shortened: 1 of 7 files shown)


aries/aries_vcx_ledger/src/ledger/indy_vdr_ledger.rs
```diff
@@ -487,7 +528,18 @@ where
         let cred_def = self
             .response_parser
             .parse_get_cred_def_response(&response, None)?;
-        Ok(cred_def.convert(())?)
+
+        // extract a map seqNo -> schemaId if required
```
JamesKEbert, line 532:
```suggestion
        // extract and map seqNo -> schemaId if required
```
maybe?

JamesKEbert (approved):

aries/aries_vcx_ledger/src/ledger/indy_vdr_ledger.rs
gmulhearn-anonyome: oops, thanks!

JamesKEbert:
Would a GH issue be appropriate on the ACA-Py side about this (given that we're adjusting it and Credo already has this behavior)?

JamesKEbert (approved):

gmulhearn:
@JamesKEbert thnx for review. And yea acapy issue probably a good idea to re-discuss with them. Although I should probably re-test that it still occurs, my testing environment where it was happening a few months ago might be dated

PR #1316, opened 2024-12-18
chore: add openssl vendoring feature flags

JamesKEbert:
Adds a feature flag for using a vendored version of openssl. This needs to drill down to any packages consuming anoncreds-rs or anoncreds-clsignatures-rs, so it impacts a lot of our crates. I _think_ I got them all.


JamesKEbert:
I'll retake a look at this after #1320, as I suspect that it'll need some conflict resolution. Plus I need to fix the failing test, not sure what that's about.

PR #1317, opened 2024-12-25
(chore) Remove unused dependencies

nain-F49FF806:
These were found using [cargo-shear]

[cargo-shear]: https://crates.io/crates/cargo-shear

If there is interest then we could add a shear step to the CI to automate this.

Note: If automating through CI then some false positives (like `strum`) may need to be ignored using [metadata hints](https://github.com/Boshen/cargo-shear?tab=readme-ov-file#ignore-false-positives)


gmulhearn (approved):
LGTM! mediator test is failing, however your other now merged PR should fix that. so i'll merge this

PR #1318, opened 2024-12-25
(refactor) Centralise dependencies definitions into workspace version catalog

nain-F49FF806:
I recall facing a lot of dependency conflicts when adding or updating any dependency. 

This was in part because all our packages define their own dependencies individually, 
and since there's cross dependencies among our crates, there would often be version conflicts. 
For example when adding a new dependency or updating an existing one. 

This PR centralises as many dependencies as possible into the workspace definition. 
So there's one central place to define versions. 
One can always manually use a different version in a package if desired. 

The resulting benefits should be:
* easy upgrade of deps in future 
* may help reduce binary size (as we carry fewer versions of each dep)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -538,7 +538,6 @@ dependencies = [
  "aries_vcx_anoncreds",
  "aries_vcx_ledger",
  "aries_vcx_wallet",
- "async-channel",
  "async-trait",
  "backtrace",
  "base64 0.22.1",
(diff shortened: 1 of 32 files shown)


gmulhearn:
doh! i started doing this as well after you: https://github.com/hyperledger/aries-vcx/pull/1320 . i'll have a look at this PR this week

gmulhearn:
hey, i got your other 2 PRs in, but this one has lots of conflicts and don't want to force you to deal with them 😅. i decided just to rebase my own https://github.com/hyperledger/aries-vcx/pull/1320, which does similar to yours + some extra dep & rust updates

PR #1320, opened 2024-12-31
(chore) rust 1.84 update, dependency updates and workspace deps usage

gmulhearn:
Updates a bunch of major dep versions (notably askar), updates minor versions in the lockfile with `cargo update`, and move alot of common dependencies to workspace deps.

Note that #1318 includes some of the same changes for workspace, however this PR adds rust updates and dep updates in addition. As such this PR would close #1318.  My bad for not coordinating these PRs better

PR #1324, opened 2025-02-26
Adjust MSRV, fix VDR Proxy Build

JamesKEbert:
Missed version updates from #1320 that are breaking some CI checks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ authors = [
 description = "Library to work with Aries protocols & collection of supporting components"
 license = "Apache-2.0"
 edition = "2021"
-rust-version = "1.79"
+rust-version = "1.84.1"
 
 [profile.release]
 debug = 0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ authors = [
 description = "Library to work with Aries protocols & collection of supporting components"
 license = "Apache-2.0"
 edition = "2021"
-rust-version = "1.79"
+rust-version = "1.81"
 
 [profile.release]
 debug = 0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ authors = [
 description = "Library to work with Aries protocols & collection of supporting components"
 license = "Apache-2.0"
 edition = "2021"
-rust-version = "1.84.1"
+rust-version = "1.79"
 
 [profile.release]
 debug = 0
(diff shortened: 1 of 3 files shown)


JamesKEbert:
Note that the failing `build-docker-vdrproxy` check is expected

gmulhearn:
Ahhh I think the vdr proxy build might work if you bump the checkout version to match our *new* Indy-vdr version/commit that we use around the codebase.

I.e. changing this:

RUN git checkout c143268

JamesKEbert:
> Ahhh I think the vdr proxy build might work if you bump the checkout version to match our new Indy-vdr version/commit that we use around the codebase.

Ah that'd make sense--trying now :+1: 

> Also, if possible, I'd avoid raising the MSRV for our crates - which I believe the rust-version in cargo.toml enforces

Do we know that our MSRV is 0.79 given the last PR? Meaning, will VCX continue to work on 0.79, even though we're targeting 0.84.1 now? Do we need to in some way validate this via testing, or is that overkill at this stage?

gmulhearn:
> Do we know that our MSRV is 0.79 given the last PR?

@JamesKEbert hmm yeah it's a good question, we probably need a reliable way to check what our MSRV actually is.. And then ideally our CI will test compilation of the project at the MSRV, and at the target version as well (which should be pretty close to the latest rust version)

I just ran this on vcx: https://github.com/foresterre/cargo-msrv?tab=readme-ov-file

`1.81` came out as our MSRV, i believe bcus our transitive dependencies have MSRVs of 1.81+ (which i think we were seeing issues for). So ye i'd recommend 1.81 become our MSRV - apologies for the muck around.

But in terms of future proofing, i think the answer is to run `cargo check --all-features` (or `just check-workspace` perhaps?) on VCX in the CI, once for with our MSRV toolchain (1.81) and once with our target (1.84.1). If you agree, maybe we need a ticket for this?

Cargo.toml
```diff
@@ -42,7 +42,7 @@ authors = [
 description = "Library to work with Aries protocols & collection of supporting components"
 license = "Apache-2.0"
 edition = "2021"
-rust-version = "1.84.1"
+rust-version = "1.79"
```
gmulhearn, line 45: sorry, i think it might technically be 1.81 now (see my other comment).

JamesKEbert:
>1.81 came out as our MSRV, i believe bcus our transitive dependencies have MSRVs of 1.81+ (which i think we were seeing issues for). So ye i'd recommend 1.81 become our MSRV - apologies for the muck around.

Thanks for figuring it out! I'll get it updated on this branch :+1: 

> But in terms of future proofing, i think the answer is to run cargo check --all-features (or just check-workspace perhaps?) on VCX in the CI, once for with our MSRV toolchain (1.81) and once with our target (1.84.1). If you agree, maybe we need a ticket for this?

Agreed!

JamesKEbert:
Also adjusted the README to reflect the MSRV, please critique as you see fit

gmulhearn (approved):
awesome, thanks for the iterations

PR #1331, opened 2025-03-07
(fix) CI fix post OWF migration

gmulhearn:


JamesKEbert (approved):
Ah yeah that could definitely do it 😆

PR #1347, opened 2025-12-04
Remove UniFFI Wrapper & Update Deps

### LINKED ISSUE #1346: (chore) Remove/decommission Android UniFFI Demo
The current android uniffi demo has gone stale, with no active contributions to turn it into a full-fledged SDK/App, it has turned into bloat in the codebase.

With that, i think it's time to remove or decommission it in some way. The UniFFI layer can be redeveloped using the latest UniFFI versions when desired. However right now, it is not in a stable state.

cc @dc-anon

gmulhearn:
Closes #1346 (Remove UniFFI wrapper)

Update deps, bump rust to 1.88


dc-anon (approved):
Looks good to me

